PR #6, opened 2018-05-16
A few fixes to the MNIST model:

mhong:
1. Normalized a few gradient values by the bach size.
2. Changed the loss function from MSE to cross entropy, which works with the
hard-coded gradients.

The change of loss values across training iterations before this CL:

Begin training for 20 iterations.
0.9
0.1
0.1
0.1
0.1
0.1
0.1
0.1
0.1
0.177527
0.25868
0.581863
0.18193
0.1
0.1
0.1
0.1
0.1
0.177527
0.1
0.1

After this CL:

Begin training for 20 iterations.
nan
113.156
88.0401
62.9524
38.1512
16.3276
5.25584
3.61207
3.37472
3.30145
3.2731
3.26038
3.2549
3.25211
3.25062
3.2499
3.24947
3.2495
3.24934
3.24926
3.24926

There might be further tuning opportunities with learning rate, etc, but is beyond the scope of this CL.

Working on this CL exposed a few compiler bugs. Left FIXME comments accordingly,
and will file bugs.

### DIFF
--- a/MNIST/MNIST.swift
+++ b/MNIST/MNIST.swift
@@ -18,7 +18,7 @@ import TensorFlow
 /// Returns the images tensor and labels tensor.
 public func readMnist(
     imagesFile: String, labelsFile: String
-) -> (Tensor<Float>, Tensor<Int32>, Tensor<Float>) {
+) -> (Tensor<Float>, Tensor<Int32>) {
     print("Reading data.")
     let imageData =
         try! Data(contentsOf: URL(fileURLWithPath: imagesFile)).dropFirst(16)
@@ -32,9 +32,7 @@ public func readMnist(
     print("Constructing data tensors.")
     let imagesTensor = Tensor(shape: [rowCount, columnCount], scalars: images)
     let labelsTensor = Tensor(labels)
-    return (imagesTensor.toDevice(),
-            labelsTensor.toDevice(),
-            Tensor<Float>(Float(rowCount)).toDevice())
+    return (imagesTensor.toDevice(), labelsTensor.toDevice())
 }
 
 func main() {
@@ -51,8 +49,8 @@ func main() {
         scriptDirectory.appendingPathComponent("train-images-idx3-ubyte").path
     let labelsFile =
         scriptDirectory.appendingPathComponent("train-labels-idx1-ubyte").path
-    let (images, numericLabels, batchSize) = readMnist(imagesFile: imagesFile,
-                                                       labelsFile: labelsFile)
+    let (images, numericLabels) = readMnist(imagesFile: imagesFile,
+                                            labelsFile: labelsFile)
     let labels = Tensor<Float>(oneHotAtIndices: numericLabels, depth: 10)
     // FIXME: Defining batchSize tensor as follows instead of returning it from
     // readMnist() crashes the compiler: https://bugs.swift.org/browse/SR-7706
--- a/MNIST/MNIST.swift
+++ b/MNIST/MNIST.swift
@@ -15,6 +15,7 @@
 import Foundation
 import TensorFlow
 
+/// Returns the images tensor and labels tensor.
 public func readMnist(
     imagesFile: String, labelsFile: String
 ) -> (Tensor<Float>, Tensor<Int32>) {
@@ -51,6 +52,10 @@ func main() {
     let (images, numericLabels) = readMnist(imagesFile: imagesFile,
                                             labelsFile: labelsFile)
     let labels = Tensor<Float>(oneHotAtIndices: numericLabels, depth: 10)
+    // FIXME: Defining batchSize tensor as follows instead of returning it from
+    // readMnist() crashes the compiler: https://bugs.swift.org/browse/SR-7706
+    // let batchSize = Tensor<Float>(Float(images.shape[0]))
+    let batchSize = Tensor<Float>(images.shapeTensor[0])
 
     // Hyper-parameters.
     let iterationCount: Int32 = 20
@@ -75,7 +80,7 @@ func main() {
         let predictions = sigmoid(z2)
 
         // Backward pass.
-        let dz2 = predictions - labels
+        let dz2 = (predictions - labels) / batchSize
         let dw2 = h1.transposed(withPermutations: 1, 0) ⊗ dz2
         let db2 = dz2.sum(squeezingAxes: 0)
         let dz1 = dz2.dot(w2.transposed(withPermutations: 1, 0)) * h1 * (1 - h1)
@@ -88,8 +93,24 @@ func main() {
         w2 -= dw2 * learningRate
         b2 -= db2 * learningRate
 
-        // Update loss.
-        loss = dz2.squared().mean(squeezingAxes: 1, 0).scalarized()
+        // Update the sigmoid-based cross-entropy loss, where we treat the 10
+        // class labels as independent. This is unnecessary for the MNIST case,
+        // where we want to predict a single label. In that case we should
+        // consider switching to a softmax-based cross-entropy loss.
+        //
+        // Let m be the batch size, y be the target labels, and A be the
+        // predictions.  The formula expressed in TF expression is:
+        // 1/m * tf.reduce_sum(- y * tf.log(A) - (1-y) * tf.log(1-A))
+        let part1 = -(labels) * log(predictions)
+        let part2 = -(1 - labels) * log(1 - predictions)
+        loss = ((part1 + part2).sum(squeezingAxes: 1, 0) / batchSize)
+          .scalarized()
+        // To print out the loss value per iteration, uncomment the following
+        // code.
+        // FIXME: Fix runtime hanging when we print loss directly instead of
+        // printing via lossTensor: https://bugs.swift.org/browse/SR-7705
+        // let lossTensor = Tensor<Float>(loss)
+        // print(lossTensor)
 
         // Update iteration count.
         i += 1
@@ -97,6 +118,11 @@ func main() {
 
     // Print loss.
     print("Loss: \(loss)")
+    // Uncomment the code below if we also print out loss per loop iteration
+    // above. This will not be necessary after fixing
+    // https://bugs.swift.org/browse/SR-7705.
+    // let lossTensor = Tensor<Float>(loss)
+    // print(lossTensor)
 }
 
 main()
--- a/MNIST/MNIST.swift
+++ b/MNIST/MNIST.swift
@@ -52,8 +52,9 @@ func main() {
     let (images, numericLabels) = readMnist(imagesFile: imagesFile,
                                             labelsFile: labelsFile)
     let labels = Tensor<Float>(oneHotAtIndices: numericLabels, depth: 10)
-    // FIXME: Defining batchSize tensor as follows instead of returning it from
-    // readMnist() crashes the compiler: https://bugs.swift.org/browse/SR-7706
+    // FIXME: Defining batchSize as a scalar, or as a tensor as follows instead
+    // of returning it from readMnist() crashes the compiler:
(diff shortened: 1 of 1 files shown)


MNIST/MNIST.swift
```diff
@@ -76,20 +82,32 @@ func main() {
...
+        let part1 = -(labels) * log(predictions)
+        let part2 = -(1 - labels) * log(1 - predictions)
+        loss = ((part1 + part2).sum(squeezingAxes: 1, 0) / batchSize)
+          .scalarized()
+        // To print out the loss value per iteration, uncomment the following
+        // code.
+        // FIXME: Fix runtime hanging when we print loss directly instead of
+        // printing via lossTensor.
+        // let lossTensor = Tensor<Float>(loss)
+        // print(lossTensor)
```
dan-zheng: Uncommenting these two lines to `lossTensor` causes runtime hanging for me.
Is this not the case for you?

dan-zheng: When I uncomment the lines, the line `Begin training for 20 iterations.` never gets printed.
The last line that's printed is `Constructing data tensors.`

MNIST/MNIST.swift
```diff
@@ -76,20 +82,32 @@ func main() {
 
         // Backward pass.
         let dz2 = predictions - labels
-        let dw2 = h1.transposed(withPermutations: 1, 0) ⊗ dz2
-        let db2 = dz2.sum(squeezingAxes: 0)
+        let dw2 = h1.transposed(withPermutations: 1, 0) ⊗ dz2  / batchSize
```
dan-zheng: Extra space before `/ batchSize` on three of these lines.

MNIST/MNIST.swift
```diff
@@ -76,20 +82,32 @@ func main() {
 
         // Backward pass.
         let dz2 = predictions - labels
```
rjpower: I think you can just compute the mean here `(predictions - labels) / batchSize` and avoid the batchsize adjustments for the gradient updates.

MNIST/MNIST.swift
```diff
@@ -76,20 +82,32 @@ func main() {
...
         b1 -= db1 * learningRate
         w2 -= dw2 * learningRate
         b2 -= db2 * learningRate
 
-        // Update loss.
-        loss = dz2.squared().mean(squeezingAxes: 1, 0).scalarized()
+        // Update the cross-entropy loss.
+        // Let m be the batch size, y be the target labels and A be the
+        // predictions.  The formula expressed in TF expression is:
+        // 1/m * tf.reduce_sum(- y * tf.log(A) - (1-y) * tf.log(1-A))
```
dan-zheng: Where did you get this formula?
From the MNIST tutorial, I found this:
```python
cross_entropy = tf.reduce_mean(-tf.reduce_sum(y_ * tf.log(y), reduction_indices=[1]))
```

Applying this yields slightly better results:
```swift
loss = (-(labels * log(predictions)).sum(squeezingAxes: 1)).mean(squeezingAxes: 0).scalarized()
```
After 20 iterations, loss is 2.30131.

MNIST/MNIST.swift
```diff
@@ -15,9 +15,10 @@
 import Foundation
 import TensorFlow
 
+/// Returns the images tensor, labels tensor, and the batch size.
 public func readMnist(
     imagesFile: String, labelsFile: String
-) -> (Tensor<Float>, Tensor<Int32>) {
+) -> (Tensor<Float>, Tensor<Int32>, Tensor<Float>) {
```
rxwei: You don’t have to return the batch size. Batch size is just the first element of the shape of the images tensor.

MNIST/MNIST.swift
```diff
@@ -48,9 +51,12 @@ func main() {
...
     let labelsFile =
         scriptDirectory.appendingPathComponent("train-labels-idx1-ubyte").path
-    let (images, numericLabels) = readMnist(imagesFile: imagesFile,
-                                            labelsFile: labelsFile)
+    let (images, numericLabels, batchSize) = readMnist(imagesFile: imagesFile,
+                                                       labelsFile: labelsFile)
     let labels = Tensor<Float>(oneHotAtIndices: numericLabels, depth: 10)
+    // FIXME: Defining batchSize tensor as follows instead of returning it from
+    // readMnist() would cause S4TF compiler to crash in OSS.
+    // let batchSize = Tensor<Float>(Float(images.shape[0]))
```
rxwei: This is likely because of scalar conversion that the partitioner is trying to do. Changing this to images.shapeTensor[0] will work.

MNIST/MNIST.swift
```diff
@@ -76,20 +82,32 @@ func main() {
...
+        // 1/m * tf.reduce_sum(- y * tf.log(A) - (1-y) * tf.log(1-A))
+        let part1 = -(labels) * log(predictions)
+        let part2 = -(1 - labels) * log(1 - predictions)
+        loss = ((part1 + part2).sum(squeezingAxes: 1, 0) / batchSize)
+          .scalarized()
+        // To print out the loss value per iteration, uncomment the following
+        // code.
+        // FIXME: Fix runtime hanging when we print loss directly instead of
+        // printing via lossTensor.
+        // let lossTensor = Tensor<Float>(loss)
```
rxwei: Is there a reason you had to convert it to a tensor?

MNIST/MNIST.swift
mhong: Done

MNIST/MNIST.swift
mhong: I ran into some other issue without converting it. This will be investigated more in https://bugs.swift.org/browse/SR-7705.

MNIST/MNIST.swift
```diff
@@ -88,15 +93,36 @@ func main() {
...
-        loss = dz2.squared().mean(squeezingAxes: 1, 0).scalarized()
+        // Update the sigmoid-based cross-entropy loss, where we treat the 10
+        // class labels as independent. This is unnecessary for the MNIST case,
+        // where we want to predict a single label. In that case we should
+        // consider switching to a softmax-based cross-entropy loss.
+        //
+        // Let m be the batch size, y be the target labels, and A be the
+        // predictions.  The formula expressed in TF expression is:
+        // 1/m * tf.reduce_sum(- y * tf.log(A) - (1-y) * tf.log(1-A))
+        let part1 = -(labels) * log(predictions)
```
rxwei: Remove parentheses around `labels`?

MNIST/MNIST.swift
```diff
@@ -88,15 +93,36 @@ func main() {
...
+        let part2 = -(1 - labels) * log(1 - predictions)
+        loss = ((part1 + part2).sum(squeezingAxes: 1, 0) / batchSize)
+          .scalarized()
```
rxwei: Code outside of the Swift code base doesn't have to fit in 80 columns. So `.scalarized()` can be moved up :)

The common practice in the Swift world allows pretty long lines. But perhaps we can limit it to 100 cols like google's standard.

MNIST/MNIST.swift
```diff
@@ -88,15 +93,36 @@ func main() {
...
+        // class labels as independent. This is unnecessary for the MNIST case,
+        // where we want to predict a single label. In that case we should
+        // consider switching to a softmax-based cross-entropy loss.
+        //
+        // Let m be the batch size, y be the target labels, and A be the
+        // predictions.  The formula expressed in TF expression is:
+        // 1/m * tf.reduce_sum(- y * tf.log(A) - (1-y) * tf.log(1-A))
+        let part1 = -(labels) * log(predictions)
+        let part2 = -(1 - labels) * log(1 - predictions)
+        loss = ((part1 + part2).sum(squeezingAxes: 1, 0) / batchSize)
```
rxwei: Since this is squeezing all axes + dividing by batch size + scalarizing, you can do all of it using a single `sum()`.

```swift
loss = (part1 + part2).sum()
```

MNIST/MNIST.swift
mhong: Done.

MNIST/MNIST.swift
mhong: Done.

MNIST/MNIST.swift
mhong: Done.

MNIST/MNIST.swift
rxwei: Yep, that's what I meant but I forgot to add it to my code example. :)

MNIST/MNIST.swift
```diff
@@ -101,10 +101,9 @@ func main() {
         // Let m be the batch size, y be the target labels, and A be the
         // predictions.  The formula expressed in TF expression is:
         // 1/m * tf.reduce_sum(- y * tf.log(A) - (1-y) * tf.log(1-A))
-        let part1 = -(labels) * log(predictions)
+        let part1 = -labels * log(predictions)
         let part2 = -(1 - labels) * log(1 - predictions)
-        loss = ((part1 + part2).sum(squeezingAxes: 1, 0) / batchSize)
-          .scalarized()
+        loss = ((part1 + part2).sum() / batchSize).scalarized()
```
rxwei: Don't use `.scalarized()` here. This will make the contextual type `Tensor` and cause `sum()` to resolve to `sum(alongAxes:)` without actually summing anything. See https://forums.swift.org/t/should-labeled-variadic-parameters-accept-0-arguments for details.

Just use the scalar version of `sum`.
```swift
loss = (part1 + part2).sum() / batchSize
```

mhong: I changed the code to:
loss = (part1 + part2).sum() / batchSize.scalarized()

Making batchSize a scalar instead of tensor will make the code even simpler, but it's blocked on 
https://bugs.swift.org/browse/SR-7706. I added a FIXME as such.

rxwei: Thanks!

MNIST/MNIST.swift
dan-zheng: Thanks for all the information!
I missed the detail about sigmoid-based vs softmax-based cross entropy, that makes sense now.

mhong:
Merging this PR. Thanks all for your review!

johndpope:
hi @mhong  - I'm looking into the swift docs and finding the protocols / DifferentiableModule / Optimizer / Learnable and wondering if this sample code couldn't integrate these ? I opened this issue 
 https://github.com/tensorflow/swift-models/issues/11

dan-zheng:
From the [mailing list announcement](https://groups.google.com/a/tensorflow.org/forum/#!topic/swift/8fuBSJ0Vocs) about removing high level APIs:

> Before our open-source release, we coded up a very basic prototype of high-level APIs, placed in [HighLevel.swift](https://github.com/apple/swift/blob/bfc7dbb6a6319473e11d7c209be35896e26a67ec/stdlib/public/TensorFlow/HighLevel.swift). The intention was to explore the programming model and discover the challenges brought by the Graph Program Extraction approach to machine learning code, e.g. lack of support for aggregate parameter updates, etc. However, APIs exposed by this file, e.g. `FullyConnectedLayer` and `Optimizer`, were nothing but confusion on open source testers, as these APIs were not supposed to be used by anyone. As a result, we completely removed these APIs from the Swift TensorFlow library.
>
> A frequently asked question is whether high-level APIs will be part of the standard library in the compiler codebase. The answer to that is no, because high-level APIs will likely not require any compiler support. Instead, future high-level APIs in Swift will be a separate Swift Package under github.com/tensorflow.
>
> ...
>
> When it’s the right time to discuss high-level APIs, the Swift for TensorFlow team will work closely with the open source community, the broader TensorFlow team, and alpha users to come up with the best design.

PR #14, opened 2018-06-25
Rewrite Catch without high level APIs

### LINKED ISSUE #7: Eliminate the use of "high level APIs"
As we are removing `HighLevel.swift`, aka `FullyConnectedLayer`, etc, models should be rewritten using primitives.

dan-zheng:
Rewrite model using bare `Tensor` values, since the experimental high-level API has been removed.
Tested with [pre-built packages from 6/22](https://github.com/tensorflow/swift/blob/master/Installation.md#pre-built-packages).

Resolves #7.

### DIFF
--- a/Catch/catch.swift
+++ b/Catch/catch.swift
@@ -56,16 +56,20 @@ public protocol Agent {
 public struct CatchAgent : Agent {
     public typealias Action = CatchAction
 
-    @_versioned var layer1: FullyConnectedLayer<Float>
-    @_versioned var layer2: FullyConnectedLayer<Float>
-    @_versioned let learningRate: Float
+    @usableFromInline var weight1: Tensor<Float>
+    @usableFromInline var bias1: Tensor<Float>
+    @usableFromInline var weight2: Tensor<Float>
+    @usableFromInline var bias2: Tensor<Float>
+    @usableFromInline let learningRate: Float
 }
 
 public extension CatchAgent {
-    @_inlineable @inline(__always)
+    @inlinable @inline(__always)
     init(learningRate: Float) {
-        layer1 = FullyConnectedLayer(inputCount: 3, outputCount: 50)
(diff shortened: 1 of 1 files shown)


Catch/catch.swift
```diff
@@ -85,47 +89,61 @@ public extension CatchAgent {
...
         let maxIndex = output2.argmax()
 
         // Back-propagation.
         let dOutput2 = output2 * (1 - output2)
-        let (_, dParameters2) = layer2.gradient(for: output1,
-                                                backpropagating: dOutput2)
-        let dOutput1 = output1 * (1 - output1)
-        let (_, dParameters1) = layer1.gradient(for: input,
-                                                backpropagating: dOutput1)
+        let (dMatmul2, dBias2) = #adjoint(Tensor.+)(
```
dan-zheng: I wasn't sure whether or not to bind `#adjoint(Tensor.+)` to a local variable.
I decided not to for the extra clarity offered by the argument labels.

rxwei: SGTM.

Catch/catch.swift
```diff
@@ -56,16 +56,20 @@ public protocol Agent {
...
+    @usableFromInline var weight1: Tensor<Float>
+    @usableFromInline var bias1: Tensor<Float>
+    @usableFromInline var weight2: Tensor<Float>
+    @usableFromInline var bias2: Tensor<Float>
+    @usableFromInline let learningRate: Float
 }
 
 public extension CatchAgent {
-    @_inlineable @inline(__always)
+    @inlinable @inline(__always)
```
rxwei: Quick thought: can you try dropping `@inline(__always)` and see what happens?

dan-zheng: Dropping `@inline(__always)` works great, the exact same SIL is generated.
Done in https://github.com/tensorflow/swift-models/pull/14/commits/02ab4cdb32ef41117d904360310d7ce6827ee0c9.

PR #18, opened 2018-07-24
Update MNIST model to adopt ParameterAggregate synthesis.

rxwei:
Update MNIST model to adopt `ParameterAggregate` synthesis.

Also:

* Since most bugs have been fixed, remove workarounds on scalar promotion, partitioning crashers and send/receive crashers.

* `Tensor.transposed` does not need to specify permutations.

* A `main()` function conveys a wrong message about the programming model. Swift supports and encourages top-level code for training scripts. `main()` has been changed to `train(_:iterationCount:)`, a function that actually has a meaning.

* Code in swift-models does not need to fit within 80 columns.

* `readMnist` -> `readMNIST`

### DIFF
--- a/MNIST/MNIST.swift
+++ b/MNIST/MNIST.swift
@@ -16,7 +16,7 @@ import Foundation
 import TensorFlow
 
 /// Reads MNIST images and labels from specified file paths.
-public func readMnist(imagesFile: String, labelsFile: String) -> (images: Tensor<Float>, labels: Tensor<Int32>) {
+func readMnist(imagesFile: String, labelsFile: String) -> (images: Tensor<Float>, labels: Tensor<Int32>) {
     print("Reading data.")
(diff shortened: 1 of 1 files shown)


MNIST/MNIST.swift
```diff
@@ -35,96 +31,71 @@ public func readMnist(
...
-    var w2 = Tensor<Float>(randomUniform: [30, 10])
-    var b1 = Tensor<Float>(zeros: [1, 30])
-    var b2 = Tensor<Float>(zeros: [1, 10])
-
     // Training loop.
     print("Begin training for \(iterationCount) iterations.")
 
-    var i: Int32 = 0
-    repeat {
+    for i in 0...iterationCount {
```
dan-zheng: `i` is unused, replace with `_`

PR #19, opened 2018-07-24
Adopt derived 'Parameterized' conformance in the Autoencoder model.

rxwei:
Plus some fixes:

* Remove the single-case `AutoencoderError` enum. Data loading for this simple model won't fail in multiple ways, so optionals should be preferred.

* Move everything from `main()` to top level.

### DIFF
--- a/Autoencoder/Autoencoder.swift
+++ b/Autoencoder/Autoencoder.swift
@@ -235,7 +235,7 @@ extension Autoencoder {
 }
 
 guard let dataset = readDataset() else {
-  print("Error")
+  print("Error: could not read dataset.")
   exit(0)
 }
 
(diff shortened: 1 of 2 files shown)


Autoencoder/Autoencoder.swift
```diff
@@ -240,38 +234,45 @@ extension Autoencoder {
...
-        autoencoder.train(on: dataset)
-        embedding = autoencoder.embedding(from: dataset, shouldSaveInput: false, elementCount: 300, step: 5)
-        plot(image: embedding.tensor, labels: embedding.labels, step: 5)
-        
-        print("Now, you can open /tmp/mnist-test/ folder and review the results.")
-    } catch {
-        print(error)
-    }
+guard let dataset = readDataset() else {
+  print("Error")
```
dan-zheng: Make error message more descriptive? `print("Error: could not read dataset")`

PR #25, opened 2018-08-29
Merge master to stable

rxwei:


rxwei:
@dan-zheng I think you need to say "I confirm that I am ok with my commits being contributed to this project."

dan-zheng:
I confirm that I am okay with my commits being contributed to this project.

PR #34, opened 2019-02-02
Fixes epoch loop in MNIST example, adds epoch count to output

valeriyvan:


MNIST/MNIST.swift
```diff
@@ -115,7 +115,7 @@ func train(_ parameters: inout MNISTParameters, epochCount: Int32) {
             totalGuesses += Int(minibatchSize)
         }
         print("""
-              Accuracy: \(correctGuesses)/\(totalGuesses) \
+              Epoch[\(epoch)] Accuracy: \(correctGuesses)/\(totalGuesses) \
```
dan-zheng: How about this format?
```suggestion
              [Epoch \(epoch)] Accuracy: \(correctGuesses)/\(totalGuesses) \
```

valeriyvan: done

rxwei:
We haven't updated the toolchains for a while. Will release a bundle of newly-built things altogether in coming weeks.

PR #39, opened 2019-02-23
A simple RL model solving the CartPole problem from OpenAI gym.

mhong:
On a Linux GPU env, it took ~25 mini batches (batch size 16) to solve (i.e.,
trained to converged) with Adam optimizer. Hyper params such as learning rate have not been well tuned.

Some print statements (including commented out ones) are intentionally left
around for debugging purposes.

### DIFF
--- a/Gym/CartPole.swift
+++ b/Gym/CartPole.swift
@@ -18,14 +18,17 @@ import TensorFlow
 let np = Python.import("numpy")
 let gym = Python.import("gym")
 
+_RuntimeConfig.printsDebugLog = false
+_RuntimeConfigP.printsDebugLog = false
+
 /// Model parameters and hyper parameters.
 let hiddenSize = 128
 let batchSize = 16
 /// Controls the amount of good/long episodes to retain for training.
 let percentile = 70
 
-/// Force unwrapping with ! does not provide source location when unwrapping
-/// nil, so we instead make a util function for debuggability.
+// Force unwrapping with ! does not provide source location when unwrapping
+// nil, so we instead make a util function for debuggability.
 fileprivate extension Optional {
   func unwrapped(file: StaticString = #file, line: UInt = #line) -> Wrapped {
     guard let unwrapped = self else {
@@ -69,7 +72,7 @@ struct Episode {
 /// Filter out bad/short episodes before we feed them as neural net training
 /// data.
 func filteringBatch(episodes: [Episode],
-                    actionCount: Int32
+                    actionCount: Int
 ) -> (input: Tensor<Float>,
       target: Tensor<Float>,
       episodeCount: Int,
@@ -90,7 +93,7 @@ func filteringBatch(episodes: [Episode],
 
     let observationTensor = Tensor<Float>(episode.steps.map { $0.observation })
     let actionTensor = Tensor<Int32>(episode.steps.map { $0.action })
-    let oneHotLabels = Tensor<Float>(oneHotAtIndices: actionTensor, depth: actionCount)
+    let oneHotLabels = Tensor<Float>(oneHotAtIndices: actionTensor, depth: Int32(actionCount))
 
     // print("observations tensor has shape \(observationTensor.shapeTensor)")
     // print("actions tensor has shape \(actionTensor.shapeTensor)")
@@ -116,7 +119,7 @@ func filteringBatch(episodes: [Episode],
 func nextBatch(env: PythonObject,
                net: Net,
                batchSize: Int,
-               actionCount: Int32
+               actionCount: Int
 ) -> [Episode] {
   var observationNumpy = env.reset()
 
@@ -130,12 +133,11 @@ func nextBatch(env: PythonObject,
 
     while true {
       let observationPython = Tensor<Double>(numpy: observationNumpy).unwrapped()
-      let actionProbabilities = softmax(
-        net.applied(to: Tensor<Float>(observationPython).reshaped(to: [1, 4]),
-                    in: context))
+      let actionProbabilities =
+          softmax(net.applied(to: Tensor(observationPython).reshaped(to: [1, 4]), in: context))
       let actionProbabilitiesPython = actionProbabilities[0].makeNumpyArray()
       let len = Python.len(actionProbabilitiesPython)
-      assert(actionCount == Int32(Python.len(actionProbabilitiesPython)))
+      assert(actionCount == Int(Python.len(actionProbabilitiesPython)))
       
       let actionPython = np.random.choice(len, p: actionProbabilitiesPython)
       let (nextObservation, reward, isDone, _) = env.step(actionPython).tuple4
@@ -147,7 +149,7 @@ func nextBatch(env: PythonObject,
 
       episodeReward += Float(reward).unwrapped()
 
-      if Bool(isDone).unwrapped() {
+      if isDone == true {
         // print("Finishing an episode with \(observations.count) steps and total reward \(episodeReward)")
         episodes.append(Episode(steps: steps, reward: episodeReward))
         observationNumpy = env.reset()
@@ -162,11 +164,11 @@ func nextBatch(env: PythonObject,
 }
 
 let env = gym.make("CartPole-v0")
-let observationSize = Int32(env.observation_space.shape[0]).unwrapped()
-let actionCount = Int32(env.action_space.n).unwrapped()
+let observationSize = Int(env.observation_space.shape[0]).unwrapped()
+let actionCount = Int(env.action_space.n).unwrapped()
 // print(actionCount)
 
-var net = Net(observationSize: Int(observationSize), hiddenSize: hiddenSize, actionCount: Int(actionCount))
+var net = Net(observationSize: Int(observationSize), hiddenSize: hiddenSize, actionCount: actionCount)
 /// SGD optimizer reaches convergence with ~125 mini batches, while Adam uses ~25.
 // let optimizer = SGD<Net, Float>(learningRate: 0.1, momentum: 0.9)
 let optimizer = Adam<Net, Float>(learningRate: 0.01)
--- a/Gym/CartPole.swift
+++ b/Gym/CartPole.swift
@@ -18,9 +18,6 @@ import TensorFlow
 let np = Python.import("numpy")
 let gym = Python.import("gym")
 
-_RuntimeConfig.printsDebugLog = false
-_RuntimeConfigP.printsDebugLog = false
-
 /// Model parameters and hyper parameters.
 let hiddenSize = 128
 let batchSize = 16
--- a/Gym/CartPole.swift
+++ b/Gym/CartPole.swift
@@ -1,0 +1,204 @@
+// Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Python
+import TensorFlow
+
+let np = Python.import("numpy")
+let gym = Python.import("gym")
+
+_RuntimeConfig.printsDebugLog = false
+_RuntimeConfigP.printsDebugLog = false
+
+let HIDDEN_SIZE = 128
+let BATCH_SIZE = 16
+let PERCENTILE = 70
+
+func scalarFromPython<Scalar: PythonConvertible>(value: PythonObject) -> Scalar {
+  let swiftValue = Scalar.init(value)
+  assert(swiftValue != nil)
+  return swiftValue!
+}
+
+func tensorFromPython<Scalar: NumpyScalarCompatible>(numpy: PythonObject) -> Tensor<Scalar> {
+  let tensor = Tensor<Scalar>(numpy: numpy)
+  assert(tensor != nil)
+  return tensor!
+}
+
+// A simple two layer dense net.
+public struct Net: Layer {
+  var l1, l2: Dense<Float>
+
+  public init(obsSize: Int, hiddenSize: Int, actionCount: Int) {
+    self.l1 = Dense<Float>(
+      inputSize: obsSize, outputSize: hiddenSize, activation: { $0 })
+
+    self.l2 = Dense<Float>(
+      inputSize: hiddenSize, outputSize: actionCount, activation: { $0 })
+  }
+
+  @differentiable(wrt: (self, input))
+  public func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
+    return l2.applied(to: relu(l1.applied(to: input, in: context)), in: context)
+  }
+}
+
+// An episode is a list of steps, where each step records the observation from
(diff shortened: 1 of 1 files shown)


Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Python
+import TensorFlow
+
+let np = Python.import("numpy")
+let gym = Python.import("gym")
+
+let HIDDEN_SIZE = 128
```
rxwei: Please use camel case for all variables.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+let BATCH_SIZE = 16
+let PERCENTILE = 70
+
+func scalarFromPython<Scalar: PythonConvertible>(value: PythonObject) -> Scalar {
```
rxwei: This is not an idiomatic use. The caller should call the scalar type's initializer on the python object directly, and use force unwrapping.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+  let swiftValue = Scalar.init(value)
+  assert(swiftValue != nil)
+  return swiftValue!
+}
+
+func tensorFromPython<Scalar: NumpyScalarCompatible>(numpy: PythonObject) -> Tensor<Scalar> {
```
rxwei: Similarly, the caller should use `Tensor<Scalar>(numpy:)` directly.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+func loss(model: Net, input: Tensor<Float>, target: Tensor<Float>)
+  -> Tensor<Float> {
+  let context = Context(learningPhase: .training)
+  let logits = model.applied(to: input, in: context)
+  return softmaxCrossEntropy(logits: logits, labels: target)
+}
+
+func train() {
+  let env = gym.make("CartPole-v0")
+  let obsSize: Int32 = scalarFromPython(value: env.observation_space.shape[0])
```
rxwei: Rename to `observationSize`.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+  let tensor = Tensor<Scalar>(numpy: numpy)
+  assert(tensor != nil)
+  return tensor!
+}
+
+// A simple two layer dense net.
+public struct Net: Layer {
+  var l1, l2: Dense<Float>
+
+  public init(obsSize: Int, hiddenSize: Int, actionCount: Int) {
```
rxwei: Avoid short names. Use "observationSize".

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func tensorFromPython<Scalar: NumpyScalarCompatible>(numpy: PythonObject) -> Tensor<Scalar> {
+  let tensor = Tensor<Scalar>(numpy: numpy)
+  assert(tensor != nil)
+  return tensor!
+}
+
+// A simple two layer dense net.
+public struct Net: Layer {
```
rxwei: This is a single file script, so we should drop all `public`s from this file.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+// We pass in `net` by reference to avoid making a copy, even though this function is
+// running inferencing over net, and not mutating it.
+func nextBatch(env: PythonObject, net: inout Net, batchSize: Int, actionCount: Int32
+) -> [Episode] {
+  var obsNumpy = env.reset()
+
+  let context = Context(learningPhase: .inference)
+  var episodes: [Episode] = []
+  
+  // We build up a batch of observations and actions
```
rxwei: End comments with a period.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+    // print("target tensor has shape \(target.shapeTensor)")
+
+    totalReward += episode.reward
+    retainedEpisodeCount += 1
+  }
+  
+  return (input, target, retainedEpisodeCount, totalReward / Float(retainedEpisodeCount))
+}
+
+// We pass in `net` by reference to avoid making a copy, even though this function is
```
rxwei: This is unnecessary and you should pass the net value directly instead. It's a constant, and there is completely no copy.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+// running inferencing over net, and not mutating it.
+func nextBatch(env: PythonObject, net: inout Net, batchSize: Int, actionCount: Int32
```
rxwei: Use idiomatic argument labeling and line wrapping.

```swift
func nextBatch(
    inEnvironment env: PythonObject,
    using net: Net,
    batchSize: Int,
    actionCount: Int32
) -> [Episode] {
    ...
}
```

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+      } else {
+        obsNumpy = nextObs
+      }
+    }
+  }
+
+  return episodes
+}
+
+func loss(model: Net, input: Tensor<Float>, target: Tensor<Float>)
```
rxwei: This function should be inlined into `gradient(at:in:)` trailing closure. These function wrappers are trivial and indirections are unnecessary.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+    self.l1 = Dense<Float>(
+      inputSize: obsSize, outputSize: hiddenSize, activation: { $0 })
+
+    self.l2 = Dense<Float>(
+      inputSize: hiddenSize, outputSize: actionCount, activation: { $0 })
+  }
+
+  @differentiable(wrt: (self, input))
+  public func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
+    return l2.applied(to: relu(l1.applied(to: input, in: context)), in: context)
```
rxwei: Make `l1`'s activation be `relu` instead of calling `relu` here.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+  struct Step {
+    let obs: Tensor<Float>
+    let action: Int32
+  }
+
+  let steps: [Step]
+  let reward: Float
+}
+
+func filterBatch(episodes: [Episode], actionCount: Int32
```
rxwei: Wrap lines correctly like `nextBatch`.
Reference: https://google.github.io/swift/

rxwei: This function does not have side effects and should be named like a noun. Something like `filteringBatch` is good.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+  }
+}
+
+// An episode is a list of steps, where each step records the observation from
+// env and the action taken. They will serve respectively as the input and
+// target (label) of the neural net training.
+// 
+struct Episode {
+  struct Step {
+    let obs: Tensor<Float>
```
rxwei: Avoid using short names. Use `observations`.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+      inputSize: hiddenSize, outputSize: actionCount, activation: { $0 })
+  }
+
+  @differentiable(wrt: (self, input))
+  public func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
+    return l2.applied(to: relu(l1.applied(to: input, in: context)), in: context)
+  }
+}
+
+// An episode is a list of steps, where each step records the observation from
```
rxwei: Please start doc comments with 3 slashes.

Gym/CartPole.swift
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func loss(model: Net, input: Tensor<Float>, target: Tensor<Float>)
+  -> Tensor<Float> {
+  let context = Context(learningPhase: .training)
+  let logits = model.applied(to: input, in: context)
+  return softmaxCrossEntropy(logits: logits, labels: target)
+}
+
+func train() {
```
rxwei: Why wrap things in a `train` function and call it immediately after? This is not very idiomatic.

Gym/CartPole.swift
mhong: Done

Gym/CartPole.swift
mhong: The concern with force unwrapping is that it can be hard to debug. assert failure will point to the source location. Any suggestions?

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: moved the code to top level.

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: Discussed and converged on creating an `Optional.unwrapped()` extension. Thanks for the ideas and code!

Gym/CartPole.swift
```diff
@@ -24,20 +24,15 @@ let batchSize = 16
...
-  return swiftValue!
-}
-
-func tensorFromPython<Scalar: NumpyScalarCompatible>(numpy: PythonObject) -> Tensor<Scalar> {
-  let tensor = Tensor<Scalar>(numpy: numpy)
-  // Before force-unwrapping, we use `assert()` to make unwrapping nil more
-  // debuggable (assert points to the source location.)
-  assert(tensor != nil)
-  return tensor!
+/// Force unwrapping with ! does not provide source location when unwrapping
```
rxwei: Use 2 slashes for non-doc-comments.

Gym/CartPole.swift
```diff
@@ -134,34 +129,31 @@ func nextBatch(env: PythonObject,
...
-      let obsFloat = Tensor<Float>(obsDouble).reshaped(to: [1, 4])
-      let actionProbabilities = softmax(net.applied(to: obsFloat, in: context))
-
-      let actProbs = actionProbabilities[0].makeNumpyArray()
-      let len = Python.len(actProbs)
-      assert(actionCount == scalarFromPython(value: len) as Int32)
+      let observationPython = Tensor<Double>(numpy: observationNumpy).unwrapped()
+      let actionProbabilities = softmax(
+        net.applied(to: Tensor<Float>(observationPython).reshaped(to: [1, 4]),
+                    in: context))
```
rxwei: * Each line break indents by 4 spaces from the previous line.
* Unnecessary generic parameter.
* You can fit all of this in two lines because we are applying a 100 column limit. I recommend loosening line wrapping elsewhere in this file as well.
```swift
        let actionProbabilities =
            softmax(net.applied(to: Tensor(observationPython).reshaped(to: [1, 4]), in: context))
```

Gym/CartPole.swift
```diff
@@ -0,0 +1,197 @@
...
+      let (nextObservation, reward, isDone, _) = env.step(actionPython).tuple4
+      // print(nextObservation)
+      // print(reward)
+
+      steps.append(Episode.Step(observation: Tensor<Float>(observationPython),
+                                action: Int32(actionPython).unwrapped()))
+
+      episodeReward += Float(reward).unwrapped()
+
+      if Bool(isDone).unwrapped() {
```
rxwei:
```suggestion
      if isDone == true {
```

Gym/CartPole.swift
```diff
@@ -0,0 +1,197 @@
...
+
+  return episodes
+}
+
+let env = gym.make("CartPole-v0")
+let observationSize = Int32(env.observation_space.shape[0]).unwrapped()
+let actionCount = Int32(env.action_space.n).unwrapped()
+// print(actionCount)
+
+var net = Net(observationSize: Int(observationSize), hiddenSize: hiddenSize, actionCount: Int(actionCount))
```
rxwei: Why not make this `Int` in the first place? Same for other Int32->Int conversions.

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: done

Gym/CartPole.swift
mhong: Done, and converted some other `Int32` to `Int` as I can.

marcrasi:
Here's a copy of this that runs in Colab: https://colab.research.google.com/drive/1PzJGhCiHrgVWQJZDOlVRP_LpGiskII9-

I made zero modifications to the code, and it runs without problems. I did split it up into a few logical cells, though it also works to just paste the whole file into one cell.

mhong:
Great, work, @marcrasi!

@rxwei, is this PR good to submit? Thank you both.

Gym/CartPole.swift
```diff
@@ -0,0 +1,196 @@
...
+    // print("target tensor has shape \(target.shapeTensor)")
+
+    totalReward += episode.reward
+    retainedEpisodeCount += 1
+  }
+  
+  return (input, target, retainedEpisodeCount, totalReward / Float(retainedEpisodeCount))
+}
+
+func nextBatch(env: PythonObject,
```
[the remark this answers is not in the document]
mhong: done

Gym/CartPole.swift
```diff
@@ -0,0 +1,196 @@
...
+  return episodes
+}
+
+let env = gym.make("CartPole-v0")
+let observationSize = Int(env.observation_space.shape[0]).unwrapped()
+let actionCount = Int(env.action_space.n).unwrapped()
+// print(actionCount)
+
+var net = Net(observationSize: Int(observationSize), hiddenSize: hiddenSize, actionCount: actionCount)
+/// SGD optimizer reaches convergence with ~125 mini batches, while Adam uses ~25.
```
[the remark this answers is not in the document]
mhong: done

Gym/CartPole.swift
```diff
@@ -0,0 +1,196 @@
...
+
+  let steps: [Step]
+  let reward: Float
+}
+
+/// Filter out bad/short episodes before we feed them as neural net training
+/// data.
+func filteringBatch(episodes: [Episode],
+                    actionCount: Int
+) -> (input: Tensor<Float>,
```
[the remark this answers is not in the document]
mhong: Done

rxwei:
Actually this was not switched to 4 space indentation. @mhong can you do that when you get a chance?

mhong:
@rxwei: Sent out https://github.com/tensorflow/swift-models/pull/54 for 4 space indentation. Thanks.

PR #40, opened 2019-02-26
Update CIFAR models for API changes

jekbradbury:
Edit: switched to `sequenced(in:through:)`
Edit again: switched back

### DIFF
--- a/CIFAR/main.swift
+++ b/CIFAR/main.swift
@@ -1,43 +1,46 @@
 import TensorFlow
 
-let batchSize: Int32 = 4
+let batchSize: Int32 = 32
 
 let (trainingDataset, testDataset) = loadCIFAR10()
-let trainingBatches = trainingDataset.batched(Int64(batchSize))
 let testBatches = testDataset.batched(Int64(batchSize))
 
-let learningPhaseIndicator = LearningPhaseIndicator()
-var model = CIFARModel() // learningPhaseIndicator: LearningPhaseIndicator)
+var model = KerasModel() // or PyTorchModel()
 
 // optimizer used in the PyTorch code
-let optimizer = SGD<CIFARModel, Float>(learningRate: 0.001, momentum: 0.9)
+// let optimizer = SGD(for: model, learningRate: 0.001, momentum: 0.9, scalarType: Float.self)
 // optimizer used in the Keras code
-// let optimizer = RMSProp<CIFARModel, Float>(learningRate: 0.0001, decay: 1e-6)
+let optimizer = RMSProp(for: model, learningRate: 0.0001, decay: 1e-6, scalarType: Float.self)
 
 for epoch in 1...10 {
     print("Epoch \(epoch), training...")
-    learningPhaseIndicator.training = true
     var trainingLossSum: Float = 0
     var trainingBatchCount = 0
-    for batch in trainingBatches {
+    let trainingShuffled = trainingDataset.shuffled(sampleCount: 50000, randomSeed: Int64(epoch))
+    for batch in trainingShuffled.batched(Int64(batchSize)) {
+        let (labels, images) = (batch.first, batch.second)
         let gradients = gradient(at: model) {
-            (model: CIFARModel) -> Tensor<Float> in
-            let thisLoss = loss(
-                model: model, images: batch.second, labels: batch.first)
-            trainingLossSum += thisLoss.scalarized()
+            (model) -> Tensor<Float> in
+            let logits = model.applied(to: images, in: Context(learningPhase: .training))
+            let oneHotLabels = Tensor<Float>(
+                oneHotAtIndices: labels, depth: logits.shape[1])
+            let loss = softmaxCrossEntropy(logits: logits, labels: oneHotLabels)
+            trainingLossSum += loss.scalarized()
             trainingBatchCount += 1
-            return thisLoss
+            return loss
         }
         optimizer.update(&model.allDifferentiableVariables, along: gradients)
     }
     print("  average loss: \(trainingLossSum / Float(trainingBatchCount))")
     print("Epoch \(epoch), evaluating on test set...")
-    learningPhaseIndicator.training = false
     var testLossSum: Float = 0
     var testBatchCount = 0
     for batch in testBatches {
-        testLossSum += loss(
-        model: model, images: batch.second, labels: batch.first).scalarized()
+        let (labels, images) = (batch.first, batch.second)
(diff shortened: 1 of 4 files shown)


CIFAR/main.swift
```diff
@@ -1,43 +1,46 @@
...
+var model = KerasModel() // or PyTorchModel()
 
 // optimizer used in the PyTorch code
-let optimizer = SGD<CIFARModel, Float>(learningRate: 0.001, momentum: 0.9)
+// let optimizer = SGD(for: model, learningRate: 0.001, momentum: 0.9, scalarType: Float.self)
 // optimizer used in the Keras code
-// let optimizer = RMSProp<CIFARModel, Float>(learningRate: 0.0001, decay: 1e-6)
+let optimizer = RMSProp(for: model, learningRate: 0.0001, decay: 1e-6, scalarType: Float.self)
 
 for epoch in 1...10 {
```
pschuh: 0..<10 ?

CIFAR/main.swift
```diff
@@ -1,43 +1,46 @@
...
-    for batch in trainingBatches {
+    let trainingShuffled = trainingDataset.shuffled(sampleCount: 50000, randomSeed: Int64(epoch))
+    for batch in trainingShuffled.batched(Int64(batchSize)) {
+        let (labels, images) = (batch.first, batch.second)
         let gradients = gradient(at: model) {
-            (model: CIFARModel) -> Tensor<Float> in
-            let thisLoss = loss(
-                model: model, images: batch.second, labels: batch.first)
-            trainingLossSum += thisLoss.scalarized()
+            (model) -> Tensor<Float> in
```
rxwei: Put this at the end of the previous line. Drop parentheses.

CIFAR/main.swift
```diff
@@ -1,43 +1,46 @@
...
+            let logits = model.applied(to: images, in: Context(learningPhase: .training))
+            let oneHotLabels = Tensor<Float>(
+                oneHotAtIndices: labels, depth: logits.shape[1])
```
rxwei: No line break necessary.

CIFAR/main.swift
```diff
@@ -1,43 +1,46 @@
...
     print("  average loss: \(trainingLossSum / Float(trainingBatchCount))")
     print("Epoch \(epoch), evaluating on test set...")
-    learningPhaseIndicator.training = false
     var testLossSum: Float = 0
     var testBatchCount = 0
     for batch in testBatches {
-        testLossSum += loss(
-        model: model, images: batch.second, labels: batch.first).scalarized()
+        let (labels, images) = (batch.first, batch.second)
+        let logits = model.applied(to: images, in: Context(learningPhase: .inference))
```
rxwei: inferring(from:)

CIFAR/main.swift
```diff
@@ -1,43 +1,46 @@
...
+        let oneHotLabels = Tensor<Float>(
+            oneHotAtIndices: labels, depth: logits.shape[1])
```
rxwei: No line break necessary

CIFAR/models.swift
```diff
@@ -46,49 +43,33 @@ public struct KerasModel : Layer {
...
-        tmp = relu(conv2b.applied(to: relu(conv2a.applied(to: tmp))))
-        tmp = dropout2.applied(to: pool2.applied(to: tmp))
-        let batchSize = tmp.shape[0]
-        tmp = tmp.reshaped(
-            toShape: Tensor<Int32>([batchSize, Int32(64 * 7 * 7)]))
-        tmp = dropout3.applied(to: dense1.applied(to: tmp))
-        tmp = dense2.applied(to: tmp)
-        return tmp
+    public func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
+        let tmp1 = conv1a.applied(to: input, in: context)
```
rxwei: Use side effects? Memory leak has been fixed.

CIFAR/models.swift
```diff
@@ -43,52 +36,29 @@ public struct KerasModel : Layer {
...
-        dropout2 = Dropout<Float>(
-            probability: 0.25, learningPhaseIndicator: learningPhaseIndicator)
-        dense1 = Dense<Float>(
-            inputSize: 64 * 7 * 7, outputSize: 512, activation: relu)
-        dropout3 = Dropout<Float>(
-            probability: 0.5, learningPhaseIndicator: learningPhaseIndicator)
-        dense2 = Dense<Float>(
-            inputSize: 512, outputSize: 10, activation: { $0 })
+    public init() {
+        conv1a = Conv2D(filterShape: (3, 3, 3, 32), padding: .same, activation: relu)
```
rxwei: define inline at definition site?

CIFAR/models.swift
```diff
@@ -43,52 +36,29 @@ public struct KerasModel : Layer {
...
-        tmp = tmp.reshaped(
-            toShape: Tensor<Int32>([batchSize, Int32(64 * 7 * 7)]))
-        tmp = dropout3.applied(to: dense1.applied(to: tmp))
-        tmp = dense2.applied(to: tmp)
-        return tmp
+    @differentiable(wrt: (self, input))
+    public func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
+        return input.sequenced(
+            in: context,
+            through: conv2a, conv1b, pool1, dropout1, conv2a, conv2b, pool2, dropout2, flatten,
```
rxwei: Does it work? We only have these operators up to 5-ary. These models need to work w/ the release toolchain so i'm afraid add it to `DeepLearning` won't resolve your issue. You can define it in this file, however.

CIFAR/models.swift
jekbradbury: Not sure what you mean?

rxwei:
```swift
var conv1a = Conv2D<Float>(filterShape: (3, 3, 3, 32), padding: .same, activation: relu)
```
and remove `init()`.

jekbradbury: Oh, right, that works now :)

jekbradbury:
Unfortunately this is giving me a memory leak at head. Going to investigate later...

rxwei:
Make gradient(at:) take images (and even labels) as arguments being differentiated wrt, then you won’t have memory leaks I think.

jekbradbury:
I don't think it can take labels as that's a Tensor<Int32>

jekbradbury:
And making it take only model and images appears not to be enough to avoid the leak.

PR #47, opened 2019-03-02
Run swift-format on all Gym sources.

rxwei:
Use 4-space indentation.


rxwei:
I had some code edits on a colab. Will copy over that version directly.

PR #49, opened 2019-03-04
Update for release 0.2

dave-fernandes:
The code to download the CIFAR images does not run on macOS, but the example trains if you download the data manually. A fix for downloading can probably wait until Dataset functionality is complete.

### DIFF
--- a/CIFAR/main.swift
+++ b/CIFAR/main.swift
@@ -6,8 +6,8 @@ let (trainingDataset, testDataset) = loadCIFAR10()
 let trainingBatches = trainingDataset.batched(Int64(batchSize))
 let testBatches = testDataset.batched(Int64(batchSize))
 
-let trainContext = Context(learningPhase: LearningPhase.training)
-let testContext = Context(learningPhase: LearningPhase.inference)
+let trainContext = Context(learningPhase: .training)
+let testContext = Context(learningPhase: .inference)
 var model = CIFARModel()
 
 // optimizer used in the PyTorch code
--- a/CIFAR/main.swift
+++ b/CIFAR/main.swift
@@ -6,8 +6,9 @@ let (trainingDataset, testDataset) = loadCIFAR10()
 let trainingBatches = trainingDataset.batched(Int64(batchSize))
 let testBatches = testDataset.batched(Int64(batchSize))
 
-let learningPhaseIndicator = LearningPhaseIndicator()
-var model = CIFARModel() // learningPhaseIndicator: LearningPhaseIndicator)
+let trainContext = Context(learningPhase: LearningPhase.training)
+let testContext = Context(learningPhase: LearningPhase.inference)
+var model = CIFARModel()
 
 // optimizer used in the PyTorch code
 let optimizer = SGD<CIFARModel, Float>(learningRate: 0.001, momentum: 0.9)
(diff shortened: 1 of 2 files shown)


CIFAR/main.swift
```diff
@@ -6,8 +6,9 @@ let (trainingDataset, testDataset) = loadCIFAR10()
 let trainingBatches = trainingDataset.batched(Int64(batchSize))
 let testBatches = testDataset.batched(Int64(batchSize))
 
-let learningPhaseIndicator = LearningPhaseIndicator()
-var model = CIFARModel() // learningPhaseIndicator: LearningPhaseIndicator)
+let trainContext = Context(learningPhase: LearningPhase.training)
```
rxwei:
```suggestion
let trainContext = Context(learningPhase: .training)
```

CIFAR/main.swift
```diff
@@ -6,8 +6,9 @@ let (trainingDataset, testDataset) = loadCIFAR10()
...
+let testContext = Context(learningPhase: LearningPhase.inference)
```
rxwei: You can change call sites that use this to calling `inferring(from:)`.

CIFAR/models.swift
```diff
@@ -46,49 +46,44 @@ public struct KerasModel : Layer {
     var dense1: Dense<Float>
     var dropout3: Dropout<Float>
     var dense2: Dense<Float>
-    public init(learningPhaseIndicator: LearningPhaseIndicator) {
+    public init(learningPhase: LearningPhase) {
```
rxwei: You can remove this parameter completely.

```suggestion
    public init() {
```

CIFAR/models.swift
```diff
@@ -46,49 +46,44 @@ public struct KerasModel : Layer {
...
+        tmp = dense2.applied(to: tmp, in: context)
         return tmp
     }
 }
 
 public typealias CIFARModel = PyTorchModel
 
 @differentiable(wrt: model)
-public func loss(model: CIFARModel, images: Tensor<Float>, labels: Tensor<Int32>)
+public func loss(model: CIFARModel, images: Tensor<Float>, labels: Tensor<Int32>, in context: Context)
```
rxwei: Could you help inline this function body into the call site? I don't think this function should be defined as a standalone function. Perhaps it was defined this way to workaround some bugs.

CIFAR/main.swift
dave-fernandes: We would have to create a `CIFARModel.inferring(from:)` method to call the `Layer.inferring(from:)` methods. This seems to make more work than simply using `applied(to:in:)` everywhere. And `CIFARModel.inferring(from:)` would have to create a Context to call its `applied(to:in:)` or else recreate the whole model again using `Layer.inferring(from:)` calls.

Happy to make the change if this is the intent, but `inferring(from:)` seems more trouble than it is worth to me.

CIFAR/models.swift
dave-fernandes: It is actually used in two places -- training and test batch evaluation. I could move it to the main.swift file though, if that seems better.

CIFAR/main.swift
dave-fernandes: See comment below on loss function. That is what the Context is needed for.

CIFAR/models.swift
rxwei: Okay, I'm going to merge as is for now!

PR #50, opened 2019-03-04
Update MNIST to use the deep learning library.

rxwei:


MNIST/MNIST.swift
```diff
@@ -12,115 +12,86 @@
...
+            let loss = softmaxCrossEntropy(logits: ŷ, labels: y)
+            totalLoss += loss.scalarized()
+            return loss
         }
-        print("""
-              [Epoch \(epoch)] Accuracy: \(correctGuesses)/\(totalGuesses) \
-              (\(Float(correctGuesses) / Float(totalGuesses)))
-              """)
+        // Update the model's differentiable variables along the gradient vector.
+        optimizer.update(&classifier.allDifferentiableVariables, along: 𝛁model)
```
dan-zheng: Should this be negative?
```suggestion
        optimizer.update(&classifier.allDifferentiableVariables, along: -𝛁model)
```

rxwei: No, because optimizers are responsible for doing minimization given gradients. This is also why `RiemannSGD.update(_:along:)` is doing `.zero - tangent`, speaking of which, the `along:` argument label is not super appropriate.

dan-zheng: Aha, I see. I agree that `along:` is not quite accurate.

PR #51, opened 2019-03-04
Update Catch to use the deep learning library.

rxwei:
This PR contains minimal changes required to make Catch make use of the deep learning library, and minimal changes that attempt to correct the algorithm.

- We do not yet have a VJP for `Tensor.max()` in the library, so `𝛁loss` is not computed via automatic differentiation.
- Use reward from the previous step for correctness. It only improves learning a little bit, however.
- Change `Environment.observation()` and `Environment.reward()` to computed properties as their computation is trivial.
- Not print rewards.

### DIFF
--- a/Catch/catch.swift
+++ b/Catch/catch.swift
@@ -19,14 +19,8 @@ import TensorFlow
 // - Adopt a more principled reinforcement learning algorithm (e.g. policy
 //   gradients). The algorithm should perform some tensor computation (not a
 //   purely table-based approach).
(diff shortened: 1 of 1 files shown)


dan-zheng:
I got these final results:
```
Win rate (total): 0.6368 [3184/5000]
Win rate (final): 0.6368 [3184/5000]
```
I guess more work is needed to fix the algorithm.

rxwei:
Yes, it stabilizes at 0.6x.

PR #60, opened 2019-03-13
LR /= 10 to account for cross entropy fix

jekbradbury:


MNIST/MNIST.swift
```diff
@@ -64,7 +64,7 @@ let labels = Tensor<Float>(oneHotAtIndices: numericLabels, depth: 10)
 
 var classifier = Classifier()
 let context = Context(learningPhase: .training)
-let optimizer = SGD<Classifier, Float>(learningRate: 0.2)
+let optimizer = SGD<Classifier, Float>(learningRate: 0.02)
```
saeta: Isn't this now the default learning rate? If so, can we just remove the parameter? Or do we want to keep it for some reason?

jekbradbury: We and Keras both default to 0.01, which might be a better choice here anyway. Personally I'm a little against the idea of a "default learning rate" for a non-adaptive method (because the natural gradient scale depends on all sorts of things) but 0.01 is as good as any.

PR #64, opened 2019-03-19
Fixes README in CIFAR

valeriyvan:


brettkoonce:
🍀

CIFAR/README.md
```diff
@@ -13,5 +13,5 @@ Python 2.7.
 To train the default model, run:
 
 ```
-swift -O models.swift data.swift main.swift
+cat models.swift data.swift main.swift | swift -O -
```
dan-zheng: Hmm, I wonder if there's some `swift` flag that we can use to avoid piping.
I briefly tried `--primary-file` but couldn't figure it out.

valeriyvan: No, if there are multiple files to compile, compiler doesn’t attemp to run compiled code. If you think of it, there is some sence in it. Input files might have some code on root level. In which order run this code? Undefined. That why compiler doesn’t try to do what isn’t defined. Merging several files into one we just remove undefined behavior.

dan-zheng: `cat ... | swift -O -` seems reasonable!
Changing CIFAR to a SwiftPM project could enable `swift run`, but that's rather heavyweight.

dan-zheng:
Thanks @valeriyvan!

rxwei:
We are going to turn the entire model repo to a SPM package, once autoencoder is fixed.

PR #69, opened 2019-03-24
convert mnist to use cnn approach

brettkoonce:


MNIST/MNIST.swift
```diff
@@ -64,7 +82,7 @@ let labels = Tensor<Float>(oneHotAtIndices: numericLabels, depth: 10)
 
 var classifier = Classifier()
 let context = Context(learningPhase: .training)
-let optimizer = SGD<Classifier, Float>(learningRate: 0.02)
+let optimizer = RMSProp(for: classifier, scalarType: Float.self)
```
rxwei:
```suggestion
let optimizer = RMSProp<Classifier, Float>()
```

MNIST/MNIST.swift
```diff
@@ -30,28 +30,46 @@ func readMNIST(imagesFile: String, labelsFile: String) -> (images: Tensor<Float>
     let images = readFile(imagesFile).dropFirst(16).map { Float($0) }
     let labels = readFile(labelsFile).dropFirst(8).map { Int32($0) }
     let rowCount = Int32(labels.count)
-    let columnCount = Int32(images.count) / rowCount
+    let imgRows : Int32 = 28, imgCols : Int32 = 28
```
rxwei: * Avoid abbreviations.
* Use a "count" suffix to denote a count.
* Remove whitespaces before colons. (See [Google Swift Style Guide - Horizontal Whitespace](https://google.github.io/swift/#horizontal-whitespace)).
```suggestion
    let rowCount: Int32 = 28, columnCount: Int32 = 28
```

MNIST/MNIST.swift
```diff
@@ -30,28 +30,46 @@ func readMNIST(imagesFile: String, labelsFile: String) -> (images: Tensor<Float>
...
 
     print("Constructing data tensors.")
     return (
-        images: Tensor(shape: [rowCount, columnCount], scalars: images) / 255,
+        images: Tensor(shape: [rowCount, 1, imgRows, imgCols], scalars: images)
+		              .transposed(withPermutations: [0, 2, 3, 1]) / 255, // TF: NHWC
```
rxwei: * Indent from the expression by 2 from the previous line. See [Google Swift Style Guide - Line Wrapping (#4)](https://google.github.io/swift/#line-wrapping).

    > A continuation line that is part of a vertically-oriented comma-delimited list is indented exactly +2 from the original line.

* It looks like you used tabs instead of whitespaces on this line.

```suggestion
          .transposed(withPermutations: [0, 2, 3, 1]) / 255, // NHWC
```

MNIST/MNIST.swift
```diff
@@ -30,28 +30,46 @@ func readMNIST(imagesFile: String, labelsFile: String) -> (images: Tensor<Float>
...
         labels: Tensor(labels)
     )
 }
 
 /// A classifier.
 struct Classifier: Layer {
-    var layer1 = Dense<Float>(inputSize: 784, outputSize: 30, activation: relu)
-    var layer2 = Dense<Float>(inputSize: 30, outputSize: 10, activation: relu)
+    /// original: https://github.com/keras-team/keras/blob/master/examples/mnist_cnn.py
```
rxwei: Only use 3 slashes for doc comments. For code comments that is not intended to be exported to documentation, use 2 slashes. That said, this comment seems unnecessary in code. It is nice to have a comment that points to a reference paper or implementation in Python, but the best place to do it is in a README file, not in code.

We plan to turn the entire swift-models repo into a SwiftPM package, but it is currently blocked because the autoencoder model is not converted to use the latest deep learning API yet. After that we'd like to re-organize README files so that every model can include their documentation.

MNIST/MNIST.swift
brettkoonce: moved to imageHeight/imageWidth, removed space

MNIST/MNIST.swift
brettkoonce: tweak

MNIST/MNIST.swift
brettkoonce: removed this line, happy to help! 🚀

MNIST/MNIST.swift
brettkoonce: done

brettkoonce:
@rxwei let me know what you think! 🍀

MNIST/MNIST.swift
rxwei: Ah, I had a typo--it should be indented by 4. We following the Google Swift Style Guide except that all indentations take 4 columns. I'll fix that for you.

PR #74, opened 2019-03-28
Changes command line to make swift compiler compile and run

valeriyvan:


rxwei:
What's the current behavior? Did the compiler hang or something?

valeriyvan:
It just compiles and neither runs nor produces executable. It was discussed recently in #64

rxwei:
That looks like an optimizer bug that we also ran into this week. Have you tried turning off `-O`?

valeriyvan:
Yes, I have tried with and without `-O` and behavior is the same - compiler just compiles and doesn't attempt to run compiled code. I don't think it's a bug. I think it's expected behavior. Input swift source code files might have some code on root level. In which order compile has to run this code? Undefined. That why compiler doesn’t try to do what isn’t defined. Merging several files into one we just remove undefined behavior.

brettkoonce:
@valeriyvan are you on linux?
on a mac, the above code works, but on linux, i see what you're seeing, eg:

    (base) skoonce@squid:~/swift/swift-models/CIFAR$ swift -O models.swift data.swift main.swift
    (base) skoonce@squid:~/swift/swift-models/CIFAR$

valeriyvan:
I am on mac. It only compiles. Doesn't run.

```
swift -version
Swift version 5.0-dev (LLVM dcb9eb74a7, Clang 95cdf7c9af, Swift 434442b530)
Target: x86_64-apple-darwin18.2.0
```

brettkoonce:
i am seeing the same:

dhcp209:CIFAR koonce$ swift -O models.swift data.swift main.swift
dhcp209:CIFAR koonce$ 

swift --version
Swift version 5.0-dev (LLVM dcb9eb74a7, Clang 95cdf7c9af, Swift 434442b530)
Target: x86_64-apple-darwin18.5.0

valeriyvan:
@brettkoonce, so for you it doesn't run as well, just compiles. Did you try `cat models.swift data.swift main.swift | swift -O -`?

rxwei:
If you run it with `swiftc`, it actually shows a compiler error here: 
```console
➜  CIFAR git:(stable) swiftc -O models.swift data.swift main.swift
data.swift:51:36: error: extra argument 'filename' in call
    return loadCIFARFile(filename: "test_batch")
                                   ^~~~~~~~~~~~
```

So I believe it's `swift` is errored and not doing the right thing (printing the error).

rxwei:
Ah, actually my branch was old. `swift -O models.swift data.swift main.swift` does terminate for me.

PR #76, opened 2019-03-29
add resnet 20/32/44/56 variants

brettkoonce:
this all works, but if anybody has any suggestions on how to refactor it I would love to hear them!

### DIFF
--- a/CIFAR/resnet.swift
+++ b/CIFAR/resnet.swift
@@ -43,11 +43,9 @@ struct BasicBlock20: Layer {
             filterShape: (kernelSize, kernelSize, featureCounts.0, featureCounts.1),
             strides: strides,
             padding: .same)
-
         self.layer2 = Conv2DBatchNorm(
             filterShape: (kernelSize, kernelSize, featureCounts.1, featureCounts.2),
             padding: .same)
-
         self.shortcut = Conv2DBatchNorm(
             filterShape: (1, 1, featureCounts.0, featureCounts.3),
             strides: strides,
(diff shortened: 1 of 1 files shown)


brettkoonce:
@rxwei over here i can refactor the inner layers slightly to be:

res20: 1/1/1  -->  1 + 1 + 1
res32: 1/3/1  -->  1 + 2 + 1
res44: 1/5/1  -->  1 + 2 + 2 + 1
res56: 1/7/1  -->  1 + 2 + 2 + 2  + 1

Would that help?  I agree that we need an array based approach long term.

CIFAR/resnet.swift
```diff
@@ -65,26 +65,270 @@ struct BasicBlock: Layer {
...
+    init(
+        featureCounts: (Int, Int, Int, Int),
+        kernelSize: Int = 3,
+        strides: (Int, Int) = (2, 2)
+    ) {
+        self.layer1 = Conv2DBatchNorm(
+            filterShape: (kernelSize, kernelSize, featureCounts.0, featureCounts.1),
+            strides: strides,
+            padding: .same)
+
```
rxwei: The empty lines in between these layer initializations are not very helpful and do not increase clarity. Could you remove all of them?

CIFAR/resnet.swift
```diff
@@ -65,26 +65,270 @@ struct BasicBlock: Layer {
...
```
rxwei: Same here.

CIFAR/resnet.swift
brettkoonce: tweaked!

CIFAR/resnet.swift
brettkoonce: tweaked!

PR #79, opened 2019-04-02
add resnet 18/34 variants, cleanup

brettkoonce:


ResNet/ResNet50.swift
```diff
@@ -19,6 +25,66 @@ struct ConvBN: Layer {
     }
 }
 
+struct ResidualBasicBlock: Layer {
+    var layer1: ConvBN
+    var layer2: ConvBN
+
+    public init(
```
rxwei: The `public` here has no effect. The parent type is internal. Also, since all models are essentially executable targets, not library targets, `public` should be dropped from everything.

brettkoonce: roger, tweaked!

rxwei:
Thanks!

PR #80, opened 2019-04-03
Convert MNIST to Swift Manager

xiejw:
1. Rename the mnist.swift to main.swift so swiftpm can find it.
2. Improve the readFile method to search folders. This is required as we are changing the folder to run it. 
3. Give a better error message when file cannot be found.

### DIFF
--- a/MNIST/main.swift
+++ b/MNIST/main.swift
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import Foundation
 import TensorFlow
 import Python
 
@@ -19,8 +20,19 @@ let np = Python.import("numpy")
 
 /// Reads a file into an array of bytes.
 func readFile(_ filename: String) -> [UInt8] {
-    let d = Python.open(filename, "rb").read()
-    return Array(numpy: np.frombuffer(d, dtype: np.uint8))!
+    let possibleFolders = [".", "MNIST"]
+    for folder in possibleFolders {
+        let parent = URL(fileURLWithPath: folder)
+        let filePath = parent.appendingPathComponent(filename).path
+
+        guard FileManager.default.fileExists(atPath: filePath) else {
+            continue
+        }
+
+        let d = Python.open(filePath, "rb").read()
+        return Array(numpy: np.frombuffer(d, dtype: np.uint8))!
+    }
+    fatalError("Failed to find file with name \(filename) in following folders: \(possibleFolders)")
 }
 
 /// Reads MNIST images and labels from specified file paths.
--- a/MNIST/main.swift
+++ b/MNIST/main.swift
@@ -24,11 +24,9 @@ func readFile(_ filename: String) -> [UInt8] {
     for folder in possibleFolders {
         let parent = URL(fileURLWithPath: folder)
         let filePath = parent.appendingPathComponent(filename).path
-
         guard FileManager.default.fileExists(atPath: filePath) else {
             continue
         }
-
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -0,0 +1,14 @@
+// swift-tools-version:4.2
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "Swift-Models",
```
rxwei:
```suggestion
    name: "TensorFlowModels",
```

MNIST/main.swift
```diff
@@ -12,15 +12,27 @@
...
 
 /// Reads a file into an array of bytes.
 func readFile(_ filename: String) -> [UInt8] {
-    let d = Python.open(filename, "rb").read()
-    return Array(numpy: np.frombuffer(d, dtype: np.uint8))!
+    let possibleFolders = [".", "MNIST"]
+    for folder in possibleFolders {
+        let parent = URL(fileURLWithPath: folder)
+        let filePath = parent.appendingPathComponent(filename).path
+
```
rxwei: I think this extra line is not improving clarity and could be removed.

MNIST/main.swift
```diff
@@ -12,15 +12,27 @@
...
+        guard FileManager.default.fileExists(atPath: filePath) else {
+            continue
+        }
+
```
rxwei: This too.

MNIST/main.swift
```diff
@@ -12,15 +12,27 @@
...
+        let d = Python.open(filePath, "rb").read()
+        return Array(numpy: np.frombuffer(d, dtype: np.uint8))!
+    }
+    fatalError("Failed to find file with name \(filename) in following folders: \(possibleFolders)")
```
rxwei: Fit within 100 columns.

Package.swift
```diff
@@ -0,0 +1,14 @@
...
+    targets: [
```
rxwei: Maybe explicitly define `products:` as well?

MNIST/main.swift
xiejw: Deleted.

MNIST/main.swift
xiejw: deleted.

MNIST/main.swift
xiejw: This line is exactly 100 columns.

Package.swift
xiejw: Done.

MNIST/main.swift
```diff
@@ -12,15 +12,25 @@
...
+    for folder in possibleFolders {
+        let parent = URL(fileURLWithPath: folder)
+        let filePath = parent.appendingPathComponent(filename).path
+        guard FileManager.default.fileExists(atPath: filePath) else {
+            continue
+        }
+        let d = Python.open(filePath, "rb").read()
+        return Array(numpy: np.frombuffer(d, dtype: np.uint8))!
+    }
+    fatalError("Failed to find file with name \(filename) in following folders: \(possibleFolders)")
```
rxwei:
```suggestion
    preconditionFailure("Failed to find file with name \(filename) in the following folders: \(possibleFolders)")
```

Now you'll need to break the line :)

xiejw: As discussed, we want an error message so we will use the old one.

rxwei: I did also suggest a grammar-related change though.

PR #81, opened 2019-04-03
add resnet 101/152 variants

brettkoonce:


rxwei:
I think it would be better to do add more ResNet variants when arrays can be differentiable and when things can be generalized to aggregate types instead of being raw layers. Otherwise this is a lot of code bloat and is making code less readable.

brettkoonce:
kk, what needs to be done to make that happen?

rxwei:
For instance, to reduce code bloat, you can define a struct that includes all the repeating subcomponents. This file is really unnecessarily large right now because of the lack of generalization.

```swift
    var l4l = ResidualIdentityBlock(featureCounts: (1024, 256, 256, 1024))
    var l4m = ResidualIdentityBlock(featureCounts: (1024, 256, 256, 1024))
    var l4n = ResidualIdentityBlock(featureCounts: (1024, 256, 256, 1024))
    var l4o = ResidualIdentityBlock(featureCounts: (1024, 256, 256, 1024))
    var l4p = ResidualIdentityBlock(featureCounts: (1024, 256, 256, 1024))
```

ResNet/ResNet50.swift
```diff
@@ -145,6 +145,33 @@ struct ResidualIdentityBlock: Layer {
     }
 }
 
+struct ResidualIdentityBlockStack: Layer {
```
brettkoonce: @rxwei let me do some testing/cleanup, but is this pattern closer to what you're after?

ResNet/ResNet50.swift
```diff
@@ -145,6 +145,33 @@ struct ResidualIdentityBlock: Layer {
...
+    var layer1: ResidualIdentityBlock
```
rxwei: Call it “block1” instead?

rxwei:
Yeah, the improved pattern is what i was expecting. It's much better! I think other existing resnet variants can be generalized this way as well.

ResNet/ResNet50.swift
brettkoonce: okay!  let me do some testing and cleanup and then i will ping you again!

PR #84, opened 2019-04-04
convert cifar to package manager

brettkoonce:


Package.swift
```diff
@@ -7,11 +7,16 @@ let package = Package(
...
         .target(
             name: "MNIST",
             dependencies: [],
             path: "MNIST"),
-    ]
+        .target(
+            name: "CIFAR",
+            dependencies: [],
+            path: "CIFAR"),
+]
```
rxwei: Fix indentation of ]

brettkoonce: tweaked!

PR #87, opened 2019-04-05
WIP: Enabling function-level XLA compilation

bgogul:
This is a WIP and needs some changes in TensorFlow as well as the CompilerRuntime.

Please take a look at xla.compile.swift to get a sense for the API to invoke xla compilation. The API is based on tracing. 

This PR also has a version of batch norm that uses the API above. It is not working yet:  I have got it compiling, but I am encountering some runtime failures. (Including it in this PR, just as FYI.)

### DIFF
--- a/XLABatchNorm/bnorm.swift
+++ b/XLABatchNorm/bnorm.swift
@@ -1,0 +1,226 @@
+import TensorFlow
+
+struct PullbackArgs<T : TensorGroup, U : TensorGroup> : TensorGroup {
+  let input: T
+  let cotangent: U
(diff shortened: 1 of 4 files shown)


XLABatchNorm/bnorm.swift
```diff
@@ -0,0 +1,258 @@
+import TensorFlow
```
rxwei: Use 4-space indentation.

BradLarson:
Talking with Gogul, this is now obsolete, so I'll close it out.

PR #96, opened 2019-04-16
update cartpole for 0.3

brettkoonce:
@rxwei okay, the new problem is that there are models both in stable and then new dev has been done against master.  how do you want to get things in sync?  happy to redo my pr's if that is simpler.  otherwise would be better to do a sync before doing significant cleanup that will have to be rebased later.

### DIFF
--- a/Gym/CartPole.swift
+++ b/Gym/CartPole.swift
@@ -48,8 +48,8 @@ struct Net: Layer {
     }
 
     @differentiable
-    func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
-        return l2.applied(to: l1.applied(to: input, in: context), in: context)
+    func applied(to input: Tensor<Float>) -> Tensor<Float> {
+        return input.sequenced(through: l1, l2)
     }
 }
 
@@ -122,7 +122,6 @@ func nextBatch(
 ) -> [Episode] {
     var observationNumpy = env.reset()
 
-    let context = Context(learningPhase: .inference)
     var episodes: [Episode] = []
 
     // We build up a batch of observations and actions.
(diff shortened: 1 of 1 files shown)


Gym/CartPole.swift
```diff
@@ -181,9 +180,8 @@ while true {
       episodes: episodes, actionCount: actionCount)
 
     let gradients = gradient(at: net) { model -> Tensor<Float> in
-        let context = Context(learningPhase: .training)
-        let logits = model.applied(to: input, in: context)
-        let loss = softmaxCrossEntropy(logits: logits, oneHotLabels: target)
+        let logits = model.applied(to: input)
+        let loss = softmaxCrossEntropy(logits: logits, probabilities: target)
```
brettkoonce: also a note here that this keyword changed for reference

Gym/CartPole.swift
```diff
@@ -181,9 +180,8 @@ while true {
       episodes: episodes, actionCount: actionCount)
 
     let gradients = gradient(at: net) { model -> Tensor<Float> in
-        let context = Context(learningPhase: .training)
```
rxwei: Outside the `while true` block, you need to add a:
```swift
```

rxwei:
It would be great to pull in all changes from stable to master now and lock down stable. When v0.3 gets released, we then push master to stable.

brettkoonce:
added flag, see #97 for stable rebase!

Gym/CartPole.swift
brettkoonce: added at L174, not L131, right?

rxwei: right

rxwei:
Thanks so much!

PR #98, opened 2019-04-17
update cifar models for 0.3

brettkoonce:
@rxwei this all works, but could be refactored down more.

I am getting this error when I try to use through on my Conv2DBatchNorm Layer, should I be doing something differently?

```
resnet.swift:133:26: error: cannot invoke 'sequenced' with an argument list of type '(through: Conv2DBatchNorm
)'                                                                                                            
        let test = input.sequenced(through: inputLayer)                                                       
                         ^                                                                                    
resnet.swift:133:26: note: overloads for 'sequenced' exist with these partially matching parameter lists: (thr
ough: L1, L2), (through: L1, L2, L3), (through: L1, L2, L3, L4), (through: L1, L2, L3, L4, L5), (through: L1, 
L2, L3, L4, L5, L6)                                                                                           
        let test = input.sequenced(through: inputLayer)                                                       
                         ^                                                                                    

```

### DIFF
--- a/CIFAR/main.swift
+++ b/CIFAR/main.swift
@@ -25,7 +25,7 @@ for epoch in 1...100 {
     for batch in trainingShuffled.batched(Int64(batchSize)) {
         let (labels, images) = (batch.label, batch.data)
         let (loss, gradients) = valueWithGradient(at: model) { model -> Tensor<Float> in
-            let logits = model.applied(to: images, in: Context(learningPhase: .training))
+            let logits = model.applied(to: images)
             return softmaxCrossEntropy(logits: logits, labels: labels)
         }
         trainingLossSum += loss.scalarized()
--- a/CIFAR/main.swift
+++ b/CIFAR/main.swift
@@ -1,16 +1,16 @@
 import TensorFlow
 import Python
 PythonLibrary.useVersion(3)
+//needed to use batch norm in ResNet model layers
+Context.local.learningPhase = .training
 
 let batchSize: Int32 = 100
 
 let cifarDataset = loadCIFAR10()
 let testBatches = cifarDataset.test.batched(Int64(batchSize))
 
-//needed to use batch norm in ResNet model layers
-//Context.local.learningPhase = .training
-//var model = ResNet20()
 
+//var model = ResNet20()
 //var model = PyTorchModel()
 var model = KerasModel()
 
--- a/CIFAR/main.swift
+++ b/CIFAR/main.swift
@@ -1,8 +1,6 @@
 import TensorFlow
 import Python
 PythonLibrary.useVersion(3)
-//needed to use batch norm in ResNet model layers
-Context.local.learningPhase = .training
 
 let batchSize: Int32 = 100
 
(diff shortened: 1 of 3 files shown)


CIFAR/models.swift
```diff
@@ -3,28 +3,18 @@ import TensorFlow
 // Ported from pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html
 public struct PyTorchModel: Layer {
     var conv1 = Conv2D<Float>(filterShape: (5, 5, 3, 6), activation: relu)
-    var pool = MaxPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    var pool1 = MaxPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
```
brettkoonce: this pool was being reused, made a copy

rxwei:
> I am getting this error when I try to use through on my Conv2DBatchNorm Layer, should I be doing something differently?

The sequential helper isn't necessary if you need to apply one layer. You can just call `applied(to:)`.

brettkoonce:
@rxwei ahh, doh, give me a sec!

CIFAR/main.swift
```diff
@@ -25,7 +25,7 @@ for epoch in 1...100 {
     for batch in trainingShuffled.batched(Int64(batchSize)) {
         let (labels, images) = (batch.label, batch.data)
         let (loss, gradients) = valueWithGradient(at: model) { model -> Tensor<Float> in
-            let logits = model.applied(to: images, in: Context(learningPhase: .training))
```
rxwei: This also needs a `Context.local.learningPhase = .training` outside the training loop, otherwise batchnorm won't work.

brettkoonce:
@rxwei actually, this is as far as it can go for now.  because each middle layer is using a relu in this model i have I can't use the pass-through trick.  I could rewrite the class/structs themselves (e.g move the relu logic up a class to Conv2DBatchNorm and add a custom layer for the initial layer and then use .sequenced(through: on the middle layers in the basicblocks) but I would have to sit down and go line by line with the original model and make sure that the logic is the same.  can do, but probably should be in a different pass.  I need to rework the padding logic as well, there's some extraneous calls.  i pinky-promise to revisit this!

CIFAR/main.swift
brettkoonce: okay

brettkoonce: added a line, but commented it out since resnet is commented out by default as well, the other two models don't use batchnorm, so we shouldn't enable it for them?  or we could just enable it for everything? (but then will it get cargo-cult copied around?)

rxwei: We should enable training mode on top level.

rxwei:
No problem. Thanks!

CIFAR/main.swift
brettkoonce: tweaked!

CIFAR/main.swift
```diff
@@ -1,16 +1,16 @@
 import TensorFlow
 import Python
 PythonLibrary.useVersion(3)
+//needed to use batch norm in ResNet model layers
```
rxwei: Could you remove the comment and move it a bit closer to the training loop?

brettkoonce: put it at head of for loop!

PR #103, opened 2019-04-18
Updated Transformer model to latest Swift For TensorFlow build

leoxzhao:
Tested on swift-tensorflow-DEVELOPMENT-2019-04-16-a-osx.pkg.

### DIFF
--- a/Transformer/Model.swift
+++ b/Transformer/Model.swift
@@ -9,12 +9,14 @@ public struct TimeDistributed: Layer {
     }
 
     @differentiable(wrt: (self, input))
-    public func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
+    public func applied(to input: Tensor<Float>) -> Tensor<Float> {
         let (batchSize, timeSteps, features) = (input.shape[0], input.shape[1], input.shape[2])
-        let reshaped = input.reshaped(toShape: Tensor<Int32>([batchSize * timeSteps, features]))
-        let output = dense.applied(to: reshaped, in: context)
+        let reshaped = input.reshaped(
+            toShape: Tensor([Int32(batchSize * timeSteps), Int32(features)]))
+        let output = dense.applied(to: reshaped)
         let outputFeatures = output.shape[1]
-        return output.reshaped(toShape: Tensor<Int32>([batchSize, timeSteps, outputFeatures]))
+        return output.reshaped(
+            toShape: Tensor([Int32(batchSize), Int32(timeSteps), Int32(outputFeatures)]))
     }
 }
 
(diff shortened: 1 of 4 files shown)


Transformer/Model.swift
```diff
@@ -101,45 +103,45 @@ struct Attention: Layer {
...
 
 @differentiable(wrt: input)
-func splitHeads(_ input: Tensor<Float>, headCount: Int32) -> Tensor<Float> {
+func splitHeads(_ input: Tensor<Float>, headCount: Int) -> Tensor<Float> {
     let (batchSize, timeSteps, features) = (input.shape[0], input.shape[1], input.shape[2])
     let featuresPerHead = features / headCount
-    let splitLastDim = input.reshaped(toShape: Tensor<Int32>(
-        [batchSize, timeSteps, headCount, featuresPerHead]))
+    let splitLastDim = input.reshaped(
+        toShape: Tensor([Int32(batchSize), Int32(timeSteps), Int32(headCount), Int32(featuresPerHead)]))
```
rxwei:
```suggestion
        to: [batchSize, timeSteps, headCount, featuresPerHead])
```
Also, you can move this to the end of the previous line since it's short enough.

Transformer/Model.swift
```diff
@@ -101,45 +103,45 @@ struct Attention: Layer {
...
     let movedToFront = splitLastDim.transposed(withPermutations: 0, 2, 1, 3)
-    return movedToFront.reshaped(toShape: Tensor<Int32>(
-        [batchSize * headCount, timeSteps, featuresPerHead]))
+    return movedToFront.reshaped(
+        toShape: Tensor([Int32(batchSize * headCount), Int32(timeSteps), Int32(featuresPerHead)]))
```
rxwei: Use `reshaped(to:)` instead of `reshaped(toShape:)` since the former takes a `TensorShape` instead of a `Tensor`.

```suggestion
        to: Tensor([batchSize * headCount, timeSteps, featuresPerHead])
```
Also move this to end of the previous line.  Same for all other occurrences of `reshaped(toShape:)`.

Transformer/Model.swift
```diff
@@ -9,12 +9,14 @@ public struct TimeDistributed: Layer {
...
         let (batchSize, timeSteps, features) = (input.shape[0], input.shape[1], input.shape[2])
-        let reshaped = input.reshaped(toShape: Tensor<Int32>([batchSize * timeSteps, features]))
-        let output = dense.applied(to: reshaped, in: context)
+        let reshaped = input.reshaped(
+            toShape: Tensor([Int32(batchSize * timeSteps), Int32(features)]))
+        let output = dense.applied(to: reshaped)
         let outputFeatures = output.shape[1]
-        return output.reshaped(toShape: Tensor<Int32>([batchSize, timeSteps, outputFeatures]))
+        return output.reshaped(
+            toShape: Tensor([Int32(batchSize), Int32(timeSteps), Int32(outputFeatures)]))
```
jekbradbury: There are still a lot of `Int32`s. Are all of them necessary? (I was under the impression that most/almost all need for them has been removed)

leoxzhao: I was just trying to make it compile using the April 16 Xcode build. Then @rxwei pointed me the right direction, to use `reshape(to:)` which accepts TensorShape, saves us from all those eye sore conversions.

PR #106, opened 2019-04-18
Gets rid of Python in MNIST example

valeriyvan:
We don't need numpy to read files in Swift.

### DIFF
--- a/MNIST/MNIST.swift
+++ b/MNIST/MNIST.swift
@@ -26,8 +26,8 @@ func readFile(_ pathname: String) -> [UInt8] {
 func readMNIST(imagesFile: String, labelsFile: String) -> (images: Tensor<Float>,
                                                            labels: Tensor<Int32>) {
     print("Reading data.")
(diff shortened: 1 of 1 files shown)


MNIST/MNIST.swift
```diff
@@ -12,23 +12,22 @@
...
 import TensorFlow
-import Python
-
-let np = Python.import("numpy")
 
 /// Reads a file into an array of bytes.
-func readFile(_ filename: String) -> [UInt8] {
-    let d = Python.open(filename, "rb").read()
-    return Array(numpy: np.frombuffer(d, dtype: np.uint8))!
+func readFile(_ pathname: String) -> [UInt8] {
```
rxwei:
```suggestion
func readFile(_ path: String) -> [UInt8] {
```

rxwei:
Please merge this into `master` instead of `stable`.

valeriyvan:
done

PR #107, opened 2019-04-18
rework autoencoder for 0.3

### LINKED ISSUE #93: Autoencoder example uses deprecated API and doesn't compile anymore

brettkoonce:
Closes #93.

@rxwei Reverted to #19, then backported as many tweaks/fixes on top as I could.  Will need to be re-modernized, so to speak, but works.  Happy to redo the formatting if you are okay with this approach.

### DIFF
--- a/Autoencoder/Autoencoder.swift
+++ b/Autoencoder/Autoencoder.swift
@@ -37,7 +37,8 @@ func plot(image: [Float], name: String) {
     let pixels = array.reshape([imageHeight, imageWidth])
     if !FileManager.default.fileExists(atPath: outputFolder) {
         try! FileManager.default.createDirectory(atPath: outputFolder,
-        withIntermediateDirectories: false, attributes: nil)
+        withIntermediateDirectories: false,
+        attributes: nil)
     }
     ax.imshow(pixels, cmap: "gray")
     plt.savefig("\(outputFolder)\(name).png", dpi: 300)
--- a/Autoencoder/Autoencoder.swift
+++ b/Autoencoder/Autoencoder.swift
@@ -24,24 +24,25 @@ let plt = Python.import("matplotlib.pyplot")
 matplotlib.use("Agg")
 
 let outputFolder = "/tmp/mnist-test/"
+enum AutoencoderError: Error {
+    case noDatasetFound
+}
 
-func readDataset() -> (images: Tensor<Float>, labels: Tensor<Int32>) {
+func readDataset() throws -> (images: Tensor<Float>, labels: Tensor<Int32>) {
     print("Reading the data.")
-    let swiftFileURL = URL(fileURLWithPath: #file)
+    guard let swiftFile = CommandLine.arguments.first else { throw AutoencoderError.noDatasetFound }
+    let swiftFileURL = URL(fileURLWithPath: swiftFile)
     var imageFolderURL = swiftFileURL.deletingLastPathComponent()
     var labelFolderURL = swiftFileURL.deletingLastPathComponent()
     imageFolderURL.appendPathComponent("Resources/train-images-idx3-ubyte")
     labelFolderURL.appendPathComponent("Resources/train-labels-idx1-ubyte")
 
-    guard let imageData = try? Data(contentsOf: imageFolderURL).dropFirst(16),
-          let labelData = try? Data(contentsOf: labelFolderURL).dropFirst(8) else {
-        print("Error: could not read dataset.")
-        exit(-1)
-    }
+    let imageData = try Data(contentsOf: imageFolderURL).dropFirst(16)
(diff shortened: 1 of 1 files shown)


Autoencoder/Autoencoder.swift
```diff
@@ -24,24 +24,25 @@ let plt = Python.import("matplotlib.pyplot")
 matplotlib.use("Agg")
 
 let outputFolder = "/tmp/mnist-test/"
+enum AutoencoderError: Error {
```
rxwei: It's better to make `readDataset()` return an optional, or even print-and-exit on error. Since we are not designing a library here, throwing doesn't buy us any modularity and having an error enum with exactly one case is kind of an antipattern.

Autoencoder/Autoencoder.swift
```diff
@@ -147,36 +149,41 @@ extension Autoencoder {
         let predictions = sigmoid(z4)
 
         // Backward pass
-        let dz4 = (predictions - input) / batchSize
-        let dw4 = h3.transposed() • dz4
-        let dz3 = matmul(dz4, w4.transposed()) * (1 - h3.squared())
-        let dw3 = h2.transposed() • dz3
-        let dz2 = matmul(dz3, w3.transposed())
-        let dw2 = h1.transposed() • dz2
+        let dz4 = ((predictions - input) / batchSize)
```
rxwei: This training loop was written before automatic differentiation worked. Would you like to update it to use a few `Dense<Float>` layers and automatic differentiation instead?

Autoencoder/Autoencoder.swift
```diff
@@ -147,36 +149,41 @@ extension Autoencoder {
...
+        b2 -= db2 * learningRate
 
         return loss
     }
 }
 
 extension Autoencoder {
-    mutating public func train(on dataset: (images: Tensor<Float>, labels: Tensor<Int32>),
-                               iterationCount: Int) {
+    @inline(never)
```
rxwei: We don't need `@inline(never)` or `@inlinable` or anything like that now. Feel free to remove them. Also, it'd be good to remove all `public`s from this file.

Autoencoder/Autoencoder.swift
```diff
@@ -222,17 +228,24 @@ extension Autoencoder {
...
+    do {
+        let dataset = try readDataset()
+        var autoencoder = Autoencoder()
 
-for i in 1...5 {
-    autoencoder.train(on: dataset, iterationCount: 100)
-    embedding = autoencoder.embedding(from: dataset, shouldSaveInput: false, elementCount: 300, step: i)
-    plot(image: embedding.tensor, labels: embedding.labels, step: i)
-}
+        var embedding = autoencoder.embedding(from: dataset, shouldSaveInput: true, elementCount: 300, step: 0)
```
rxwei: Fit all lines with 100 columns.

Autoencoder/Autoencoder.swift
brettkoonce: see below!

Autoencoder/Autoencoder.swift
brettkoonce: see below!

Autoencoder/Autoencoder.swift
brettkoonce: 😉

Autoencoder/Autoencoder.swift
brettkoonce: all gone!

Autoencoder/Autoencoder.swift
```diff
@@ -20,219 +20,117 @@ import Python
...
 
 func plot(image: [Float], name: String) {
     // Create figure
     let ax = plt.gca()
     let array = np.array([image])
-    let pixels = array.reshape([Autoencoder.imageEdge * 5, Autoencoder.imageEdge * 5])
+    let pixels = array.reshape([imageHeight, imageWidth])
     if !FileManager.default.fileExists(atPath: outputFolder) {
-        try! FileManager.default.createDirectory(atPath: outputFolder, withIntermediateDirectories: false, attributes: nil)
+        try! FileManager.default.createDirectory(atPath: outputFolder,
```
rxwei: I'd suggest putting the second and the third arguments on their own lines and aligning them with the first argument.

brettkoonce: updated, let me know what you think!

rxwei: Could you align them with `atPath:`? Thanks!

brettkoonce: old obj-c style, you mean?  done!

rxwei: I meant the Swift style actually: left-aligning all argument labels. But the ObjC style does look nicer here!

PR #108, opened 2019-04-18
Gets rid of Python in MNIST example

valeriyvan:
We don't need numpy to read files in Swift.


rxwei:
Thanks!

rxwei:
Please cherrypick this to the master branch.

rxwei:
Thanks!

PR #109, opened 2019-04-18
update mnist for new layers api

brettkoonce:
@rxwei if this is the correct pattern, I can bang out the other ones!

### DIFF
--- a/MNIST/main.swift
+++ b/MNIST/main.swift
@@ -86,7 +86,7 @@ let (images, numericLabels) = readMNIST(imagesFile: "train-images-idx3-ubyte",
 let labels = Tensor<Float>(oneHotAtIndices: numericLabels, depth: 10)
 
 var classifier = Classifier()
-let optimizer = RMSProp(for:classifier)
+let optimizer = RMSProp(for: classifier)
 
 // The training loop.
 for epoch in 1...epochCount {
--- a/MNIST/main.swift
+++ b/MNIST/main.swift
@@ -53,6 +53,9 @@ func readMNIST(imagesFile: String, labelsFile: String) -> (images: Tensor<Float>
 
 /// A classifier.
 struct Classifier: Layer {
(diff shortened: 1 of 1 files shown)


MNIST/main.swift
```diff
@@ -64,7 +67,7 @@ struct Classifier: Layer {
     var layer1b = Dense<Float>(inputSize: 128, outputSize: 10, activation: softmax)
 
     @differentiable
-    func applied(to input: Tensor<Float>) -> Tensor<Float> {
+    public func call(_ input: Input) -> Output {
```
rxwei: Remove `public`.

MNIST/main.swift
```diff
@@ -83,7 +86,7 @@ let (images, numericLabels) = readMNIST(imagesFile: "train-images-idx3-ubyte",
 let labels = Tensor<Float>(oneHotAtIndices: numericLabels, depth: 10)
 
 var classifier = Classifier()
-let optimizer = RMSProp<Classifier, Float>()
+let optimizer = RMSProp(for:classifier)
```
rxwei:
```suggestion
let optimizer = RMSProp(for: classifier)
```

MNIST/main.swift
```diff
@@ -53,6 +53,9 @@ func readMNIST(imagesFile: String, labelsFile: String) -> (images: Tensor<Float>
 
 /// A classifier.
 struct Classifier: Layer {
+    public typealias Input = Tensor<Float>
```
rxwei: Remove `public` everywhere.

brettkoonce: done!

MNIST/main.swift
brettkoonce: got it!

MNIST/main.swift
brettkoonce: done!

PR #110, opened 2019-04-18
convert cifar to new layers api

brettkoonce:


CIFAR/wide_resnet.swift
```diff
@@ -219,9 +240,10 @@ struct WideResNet28: Layer {
...
     @differentiable
-    func applied(to input: Tensor<Float>) -> Tensor<Float> {
+    func call(_ input: Input) -> Output {
         let inputLayer = input.sequenced(through: l1, l2, l3, l4)
-        let finalNorm = relu(norm.applied(to: inputLayer))
+        let finalNorm = relu(norm(inputLayer))
         return finalNorm.sequenced(through: avgPool, flatten, classifier)
     }
 }
+
```
brettkoonce: i always add a line at the end, but if you don't like I can remove...

PR #113, opened 2019-04-19
Updated Transformer model to use `callable`.

leoxzhao:
I have to use `.call` explicitly for callable with `inout`. Once "Static callable does not support `inout`" (https://bugs.swift.org/browse/TF-443) get fixed, we can simply remove `.call` in the code.

### DIFF
--- a/Transformer/Model.swift
+++ b/Transformer/Model.swift
@@ -194,7 +194,7 @@ struct MultiHeadAttention: Layer {
             key: splitHeads(qkvSplit.key, headCount: headCount),
             value: splitHeads(qkvSplit.value, headCount: headCount)
         )
-        let outputs = attention.call(attentionInput, state: &state)
+        let outputs = attention(attentionInput, state: &state)
         return wo(joinHeads(outputs, headCount: headCount))
     }
 }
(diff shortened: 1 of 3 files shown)


Transformer/Model.swift
```diff
@@ -95,26 +95,27 @@ struct Attention: Layer {
...
+    func call(_ input: AttentionInput) -> Tensor<Float> {
         var dotProducts = batchedMatmul(input.query, input.key, adjointRight: true)
         dotProducts = causallyMasked(dotProducts, enable: causal) / scale
-        return batchedMatmul(dropout.applied(to: softmax(dotProducts)), input.value)
+        return batchedMatmul(dropout(softmax(dotProducts)), input.value)
     }
-    func applied(to input: AttentionInput, state: inout AttentionContext)
-        -> Tensor<Float> {
+    
+    func call(_ input: AttentionInput, state: inout AttentionContext) -> Tensor<Float> {
```
leoxzhao: `call` functions with `state` do not have `@differentiable`. I keep it that way.

brettkoonce:
Upstream: apple/swift#24157.

rxwei:
apple/swift#24157 has been merged. If you could update it to the callable syntax for the call sites that take an `inout`, that'd be great!

rxwei:
Thanks Leo!

Issue #101, opened 2019-04-17
0.3 models pass


brettkoonce:
@rxwei Did a quick pass:

  - ~~Transformer has a bunch of Int32 calls that will need to be updated~~ v2: #113 thanks @leoxzhao!
  - ~~MiniGo will require some ❤️ -> #114~~
  - ~~Autoencoder: see #93~~

Rest seem to be working!

rxwei:
Thanks a ton, Brett!

leoxzhao:
I updated Transformer model to work with April 16, 2019 Xcode toolchain (swift-tensorflow-DEVELOPMENT-2019-04-16-a-osx.pkg).

brettkoonce:
Transformer will require some tweaks to work with new layers api, very specifically in making AttentionInput play nice with being defined as a differentiable type.

brettkoonce:
@rxwei Transformer will require some tweaks to work with new layers api, very specifically in making AttentionInput play nice with being defined as a differentiable type.

#109, #110, #111 and #112 bring everything else forward (sans above).

leoxzhao:
I just read the callable proposal a few days ago. Now `Layer` is using it in April 18 Development Snapshots. This is fast. I am working on updating Transformer.

brettkoonce:
@leoxzhao cool!

leoxzhao:
Static callable (April 18 Xcode 10 Development Snapshot) has issue to handle `inout` parameter, which blocks Transformer model since it uses `inout` in a few places. I filed a bug at bugs.swift.org (https://bugs.swift.org/browse/TF-443).

brettkoonce:
@leoxzhao good find, nice testcase!

leoxzhao:
I played Go in university, but I was not an enthusiast. I left the MiniGo model for a few days to see if someone who really loved the game would claim it. A few days have passed, and nobody's claimed the model, so I've decided to take it.

rxwei:
The main author of Swift MiniGo (@xiejw) is also working on migrating it AFAIK. @xiejw, what's the current progress and is there anything we can help with? Perhaps someone can start a PR and we can iterate on it together?

rxwei:
Thanks for pushing this forward!

rxwei:
Now that everything's done, I'm going to close this issue now! Huge thanks to @brettkoonce for coordinating this.

[...]

PR #114, opened 2019-04-19
Updated MinoGo to compile and run using latest Swift for TensorFlow

leoxzhao:
#101

### DIFF
--- a/MiniGo/Models/GoModel.swift
+++ b/MiniGo/Models/GoModel.swift
@@ -175,15 +175,9 @@ public struct GoModel: Layer {
             outputSize: 1,
             activation: tanh)
     }
-
-    @differentiable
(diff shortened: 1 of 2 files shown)


rxwei:
Nice, thanks!

MiniGo/Models/GoModel.swift
```diff
@@ -176,52 +176,53 @@ public struct GoModel: Layer {
             activation: tanh)
     }
 
+    @differentiable
+    // This is just a placeholder to satisfy Layer's spec.
+    public func call(_ input: Tensor<Float>) -> Tensor<Float> {
```
rxwei: This should instead be renaming `applied(to:)` to `call(_:)`.

PR #115, opened 2019-04-19
cherry-pick 21ae62e6de2298074de153782d66a32a36b10f58

brettkoonce:
@rxwei clean rebase of #106 on master:

Gets rid of Python in MNIST example (#108)

* Get rid of Python in MNIST example. We don't need numpy to read files in Swift

* Gets rid of closures in map

* Renames parameter

### DIFF
--- a/MNIST/main.swift
+++ b/MNIST/main.swift
@@ -14,32 +14,20 @@
 
 import Foundation
 import TensorFlow
-import Python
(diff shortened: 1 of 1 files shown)


rxwei:
This is cherry-picking a commit that previously passed CLA. Merging.

PR #116, opened 2019-04-19
tweak checkpoint directory to match README?

brettkoonce:
Doing 

    swift run -Xlinker -ltensorflow -c release MiniGo

like the README suggests in the parent directory is currently failing because the logic assumes it is in the MiniGo directory.  On the flip side doing a ```swift run``` inside the MiniGo directory will fail if this get merged.  Not sure what the correct/desired behavior is.  Can also write logic to check both locations but figured I would ask first.

### DIFF
--- a/MiniGo/README.md
+++ b/MiniGo/README.md
@@ -27,9 +27,10 @@ To get a toolchain, you can:
 
 Currently, inference uses a pre-trained checkpoint. To download it, please
 install [`gsutil`](https://cloud.google.com/storage/docs/gsutil_install) and run
-the following:
+the following in the parent directory:
 
 ```sh
+cd swift-models
 mkdir -p MiniGoCheckpoint
 gsutil cp 'gs://minigo-pub/v15-19x19/models/000939-heron.*' MiniGoCheckpoint/
 ```


xiejw:
Thanks for catching this!

I think there are two ways to fix this. 
1. Assumes the MiniGoCheckpoint folder is under swift-models/MiniGo, i.e., sub-folder.
2. Assumes the MiniGoCheckpoint folder is under swift-models, i.e., the parent folder.

Both need to update the README file to make it super clear to users. 

I personally prefer the second one as we recommend users to run the code in parent folder via swiftPM. but open to the first one. 

Assuming we fix it by taking the second one. I think we could update the README as

To get the ckpt:

    cd swift-models
    mkdir -p MiniGoCheckpoint
    gsutil cp 'gs://minigo-pub/v15-19x19/models/000939-heron.*' MiniGoCheckpoint/

To run
    cd swift-models
    swift run -Xlinker -ltensorflow -c release MiniGo

Then this will align with other models, like ResNet. 

What do you think?

brettkoonce:
@xiejw I like that solution better, please see update!

PR #118, opened 2019-04-20
Uses isEmpty to check if collection is empty

valeriyvan:
[It's preferred way to check collection for emptiness instead of checking that the `count` property is equal to zero](https://developer.apple.com/documentation/swift/collection/3017690-isempty).

### DIFF
--- a/MiniGo/Sources/GameLib/LibertyTracker.swift
+++ b/MiniGo/Sources/GameLib/LibertyTracker.swift
@@ -150,7 +150,7 @@ extension LibertyTracker {
             fatalErrorForGroupsInvariance(groupID: newGroupID)
         }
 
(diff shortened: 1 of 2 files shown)


rxwei:
Thanks. Since we are using the master branch for new releases, please make this change and future changes to the master branch instead.

valeriyvan:
Sounds like `master` should be default branch, not `stable`.

valeriyvan:
`master` and `stable` are VERY different.
MiniGo in `master` doesn't even compile.

brettkoonce:
@valeriyvan it does, but you'll need either a nightly or to build tensorflow from source.  Once 0.3 comes out there will be updated packages you can download.

PR #119, opened 2019-04-20
Uses .isEmpty instead of .count > 0 for checking collections for emptiness

valeriyvan:


rxwei:
Thanks!

MiniGo/Strategies/MCTS/MCTSNode.swift
```diff
@@ -155,33 +155,29 @@ extension MCTSNode {
...
         }
 
-        let candidateCount = candidateIndexes.count
-        assert(candidateCount > 0)
-
         // Breaks the tie randomly.
-        let candidateIndex = candidateIndexes[Int.random(in: 0..<candidateCount)]
-        return elements[candidateIndex]
+        assert(!elements.isEmpty)
+        return elements.randomElement()!
```
xiejw: This line is wrong. Can you file it or rollback?

xiejw: s/file/fix

valeriyvan: @xiejw, why do you think it is wrong?

xiejw: The new change returns a random element from `elements`. It ignored the `candidateIndexes`. 

The old change picked up an `index` from `candidateIndexes` and returned the element from   `elements` according to that `index`.

So, the logic is different.

xiejw: I have fixed the tests at here 
https://github.com/tensorflow/swift-models/pull/121#pullrequestreview-229665641

PR #123, opened 2019-04-24
Implements random choose who plays black MiniGo. Solves TODO.

valeriyvan:


MiniGo/Play/PlayOneGame.swift
```diff
@@ -21,9 +21,10 @@ public func playOneGame(gameConfiguration: GameConfiguration, participants: [Pol
         participants[0].participantName !=  participants[1].participantName,
         "Participants' names should not be same.")
 
-    // TODO(xiejw): Choose a random participant to play black.
-    let blackPlayer = participants[0]
-    let whitePlayer = participants[1]
+    // Choose a random participant to play black.
+    let shuffled = participants.shuffled()
+    let blackPlayer = shuffled.first! // precondition makes safe force unwrap
```
rxwei: I don't feel the use of `first!` and `last!` makes the code more readable here. The Swift API Design Guidelines emphasize fluency, and I think `[0]` and `[1]` are the most fluent way of expressing "getting the first element" and "getting the second element" in an array that is known to have exactly 2 elements. Plus, it does not need any force unwrapping.

Could you elaborate on why you feel `.first!` and `.second!` are better?

valeriyvan: TL;DR for particular this use case I would agree that `shuffled[0]` and `shuffled[1]` are better than `shuffled.first!` and `shuffled.last!`.

Talking about readability. Why `participants` is array here at all? Is it possible to have it empty? No. Is it possible to have 5 elements? No. There will be always 2 elements. Why is it array then? Shouldn't it be a tuple? Firstly, it will make all preconditions redundant. Secondly, it will invalidate the question what is better `shuffle.first!` or `shuffle[0]`.

Why I prefer `array.first!` and `array.last!`. I dunno. May be because `array.first!` and `array.last!` are symmetrical to each other. And what's symmetrical to `array[0]`? `array[array.count-1]`? Thank you, no. I would prefer `array.last!`.

@rxwei, should I make another PR rolling back `shuffled[0]` and `shuffled[1]`?

rxwei: Please go ahead and roll them back. Thanks!

valeriyvan: done with PR #130

PR #135, opened 2019-04-29
Remove deprecated comments +`@noDerivative`

jekbradbury:
Comments no longer apply (AD bug has been fixed and arrays now conform to `Differentiable`)


rxwei:
I think that we can use [`differentiableReduce(_:_:)`](https://colab.research.google.com/drive/1gF5Z05Uf10Ox2NxLdrK73eDuhVw6B0lT#scrollTo=lqfaQ2Ozusdi) in `call(_:)` and remove the custom VJP now!

PR #137, opened 2019-04-29
Makes all classes final

valeriyvan:


BradLarson:
Thanks for suggesting this, but after further review we've decided to leave the classes as non-final to allow for potential extension in case someone wants to build on these models.

PR #139, opened 2019-04-30
rebuild wideresnet using reduce operator

brettkoonce:


CIFAR/wide_resnet.swift
```diff
@@ -144,118 +133,84 @@ struct WideResNet16: Layer {
...
-        self.block4 = BatchNormConv2DBlock(
-            filterShape: (kernelSize, kernelSize,
-                featureCounts.1 * widenFactor, featureCounts.1 * widenFactor),
-            strides: (1, 1))
-        self.shortcut = Conv2D(
-            filterShape: (1, 1, featureCounts.0, featureCounts.1 * widenFactor),
-            strides: initialStride)
+extension WideResNet {
+    enum Kind {
+        case wideresnet16
```
rxwei: How about using camel case: `wideResNet16`?

For consistency, it's better to change `ResNet.Kind` to use camel case as well.

CIFAR/wide_resnet.swift
```diff
@@ -47,93 +47,82 @@ struct BatchNormConv2DBlock: Layer {
     }
 }
 
-struct WideResnet16FirstBasicBlock: Layer {
+struct WideResnetBasicBlock: Layer {
```
rxwei:
```suggestion
struct WideResNetBasicBlock: Layer {
```

rxwei: s/Resnet/ResNet/g

brettkoonce: done!

CIFAR/wide_resnet.swift
brettkoonce: reworked!

rxwei:
Thanks!

rxwei:
Hi @brettkoonce, this patch broke the SwiftPM build. Since we made all models use SwiftPM as the primary method of building, `swift run` is strongly preferred over running individual files as a script. Please help make sure everything builds and passes tests via SwiftPM. Thanks!

PR #143, opened 2019-04-30
Fix up models so they can run.

saeta:
As part of the migration to a single SwiftPM-based package for the models
repository, the working directory where people run `swift run` has changed.
As a result, a number of relative-path lookups were broken when actually
running models. This commit fixes them up so they all properly run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 .build
 *.xcodeproj
 .DS_Store
+output
+cifar-10-batches-py/
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,6 @@
 
 .build
 *.xcodeproj
+*.png
 .DS_Store
-output
 cifar-10-batches-py/
--- a/MNIST/main.swift
+++ b/MNIST/main.swift
@@ -15,12 +15,8 @@
 import Foundation
 import TensorFlow
 
-struct FileNotFound: Error {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -4,3 +4,5 @@
 .build
 *.xcodeproj
 .DS_Store
+output
```
rxwei: Ignoring "output" feels a bit too generic here. Which model/script is producing this file?

MNIST/main.swift
```diff
@@ -15,19 +15,31 @@
 import Foundation
 import TensorFlow
 
+struct FileNotFound: Error {
```
rxwei: Since this is a simple script instead of a library, I recommend printing the error message within `readFile` and just call `exit(-1)` to gracefully exit for simplicity. Force-`try`ing will also produce a bad-looking trap on failure.

If we are writing a utility library, I would totally agree that a proper throwing function is most appropriate.

.gitignore
saeta: Yeah, unfortunately, this is what the folder produced by `Autoencoder`. I've switched this to `*.png` which should hopefully be better.

MNIST/main.swift
saeta: Switched to just using `fatalError`.

PR #154, opened 2019-05-08
Load config variable from `hparams.json` file so that Transformer can…

leoxzhao:
… work with the bigger GPT-2 models (It is called "staged release". as of now, only 117M and 345M are available).

### DIFF
--- a/Transformer/PythonCheckpointReader.swift
+++ b/Transformer/PythonCheckpointReader.swift
@@ -14,31 +14,21 @@
 
 import TensorFlow
 
-struct Config {
+struct Config : Codable {
     let vocabSize: Int
     let contextSize: Int
     let embeddingSize: Int
     let headCount: Int
     let layerCount: Int
-}
 
-extension Config {
-    init(dictionary: [String: Int]) {
-        vocabSize = dictionary["n_vocab"]!
-        contextSize = dictionary["n_ctx"]!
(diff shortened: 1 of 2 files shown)


Transformer/main.swift
```diff
@@ -14,13 +14,19 @@
...
 sys.path = sys.path + ["."]
-let encoder = Python.import("encoder").get_encoder("117M")
-
-let checkpoint = "models/117M/model.ckpt"
-let model = TransformerLM(contentsOfPythonCheckpointFile: checkpoint, scope: "model")
+let encoder = Python.import("encoder").get_encoder(modelName)
+
+let checkpoint = "models/\(modelName)/model.ckpt"
+let configFile = "models/\(modelName)/hparams.json"
+let configData = try Data.init(contentsOf: URL(fileURLWithPath: configFile))
```
rxwei:
```suggestion
let configData = try Data(contentsOf: URL(fileURLWithPath: configFile))
```

Use the initializer syntax instead of explicit `.init`.

Transformer/PythonCheckpointReader.swift
```diff
@@ -14,32 +14,22 @@
 
 import TensorFlow
 
-struct Config {
+struct Config : Codable {
```
rxwei: The space before `:` is unnecessary according to the [Google Swift Style Guide](https://google.github.io/swift/).

PR #156, opened 2019-05-09
rebuild resnet using blocks

brettkoonce:


brettkoonce:
re 1) Not sure if this is just my system (running semi-current tf-swift master), but the existing version in models repo takes 43 seconds to compile here.  Using this code takes ~3-4 seconds to compile.
re 2) Not sure how the current codebase handles arbitrary input, but certainly happy to rework things however desired.
re 3) That is fine, but having said that it would probably be good to refactor this a bit more first.  If there is a way to parameterize the BasicBlock/ResidualIdentityBlock logic then this file could be cut in half again and everything reduced to a single ResNet class, which I think would be simpler for people new to computer vision.  This torchvision logic is nice: https://github.com/pytorch/vision/blob/50d54a82d1479ffb6dd7469ed05fccdf290a1d84/torchvision/models/resnet.py#L216

Let me know!

rxwei:
> re 1) Not sure if this is just my system (running semi-current tf-swift master), but the existing version in models repo takes 43 seconds to compile here. Using this code takes ~3-4 seconds to compile.

This is expected because the compiler is bugged. The implementation detail is that we've turned off a part of the Swift compiler that optimizes large structures in order to suppress other bugs. We are actively working on this and expect this to be fixed within a week or two.

jekbradbury:
> 2\. could you make `imageSize` optional

This would be a good change. The reason `imageSize` exists at all is to distinguish between ResNets designed for CIFAR image sizes and those designed for ImageNet sizes; my impression is that the ImageNet-sized networks also work fine for images of other similar sizes.

brettkoonce:
I can work on coming up with a way to add an input type to specify cifar, imagenet and generic variants.  In the same vein, can come back to refactoring things more down the road.

ResNet/ResNet.swift
```diff
@@ -219,16 +224,18 @@ struct ResNetBasic: Layer {
...
-        l1 = ConvBN(filterShape: (7, 7, 3, 64), strides: (2, 2), padding: .same)
-        maxPool = MaxPool2D(poolSize: (3, 3), strides: (2, 2))
-        avgPool = AvgPool2D(poolSize: (7, 7), strides: (7, 7))
-        if imageSize == 32 {
-            l1 = ConvBN(filterShape: (3, 3, 3, 64), padding: .same)
-            maxPool = MaxPool2D(poolSize: (1, 1), strides: (1, 1)) // no-op
-            avgPool = AvgPool2D(poolSize: (4, 4), strides: (4, 4))
+    init(input: InputType, layerBlockCounts: (Int, Int, Int, Int)) {
+        switch input {
+            case .imagenet:
```
rxwei: `case`s should have the same indentation level as `switch`.

ResNet/ResNet.swift
```diff
@@ -20,6 +20,11 @@ import TensorFlow
 // https://arxiv.org/abs/1512.03385
 // using shortcut layer to connect BasicBlock layers (aka Option (B))
 
+enum InputType {
```
rxwei: Normally we use `kind` instead of `type` since "type" may connote the notion of "type" in programming languages. Maybe `InputKind`?

ResNet/ResNet.swift
```diff
@@ -293,16 +298,18 @@ struct ResNet: Layer {
...
-        l1 = ConvBN(filterShape: (7, 7, 3, 64), strides: (2, 2), padding: .same)
-        maxPool = MaxPool2D(poolSize: (3, 3), strides: (2, 2))
-        avgPool = AvgPool2D(poolSize: (7, 7), strides: (7, 7))
-        if imageSize == 32 {
-            l1 = ConvBN(filterShape: (3, 3, 3, 64), padding: .same)
-            maxPool = MaxPool2D(poolSize: (1, 1), strides: (1, 1)) // no-op
-            avgPool = AvgPool2D(poolSize: (4, 4), strides: (4, 4))
+    init(input: InputType, layerBlockCounts: (Int, Int, Int, Int)) {
+        switch input {
+            case .imagenet:
```
rxwei: Same regarding the indentation level of `case`s.

ResNet/ResNet.swift
```diff
@@ -260,12 +265,12 @@ extension ResNetBasic {
         case resNet34
     }
 
-    init(kind: Kind, imageSize: Int, classCount: Int) {
+    init(kind: Kind, type: InputType) {
```
rxwei: `type: InputType` -> `kind: InputKind`

ResNet/ResNet.swift
brettkoonce: done!

ResNet/ResNet.swift
brettkoonce: tweaked!

ResNet/ResNet.swift
brettkoonce: done!

ResNet/ResNet.swift
brettkoonce: tweaked!

ResNet/ResNet.swift
brettkoonce: done!

brettkoonce: done!

ResNet/ResNet.swift
```diff
@@ -0,0 +1,353 @@
...
+}
+
+extension ResNet {
+    enum Kind {
+        case resNet50
+        case resNet101
+        case resNet152
+    }
+
+    init(kind: Kind, type: InputKind) {
```
rxwei: If `kind` conflicts with the first argument label, you can call it `inputKind:`. Relatedly, would `DataKind` be a better name for the type? If so, `dataKind:` would be a good label name. Same for other initializers that take `DataKind`.

brettkoonce: moved to:  ```inputKind: Kind, dataKind: DataKind```, let me know what you think!

PR #158, opened 2019-05-16
debug filterShapes and match network to citation

mikowals:
If widenFactor is greater than 1, the input shape of the first residual block had the wrong number of input channels.  Similarly the same filterShape cannot be used for both conv1 and conv2 when the number of channels is being increased.

Added identity connections, dropout between convolutions, and the use of preactivation in the 1x1 conv2D shortcuts.  

The paper also uses a weight decay of 5e-4.  I can add a function for this but I have not seen any implementations of L2 decay in any of the models so I wondered if there was some plan to add it somewhere else in the API.

### DIFF
--- a/CIFAR/WideResNet.swift
+++ b/CIFAR/WideResNet.swift
@@ -20,14 +20,46 @@ import TensorFlow
 // https://arxiv.org/abs/1605.07146
 // https://github.com/szagoruyko/wide-residual-networks
 
-struct BatchNormConv2DBlock: Layer {
+struct IdentityLayer: Layer {
     typealias Input = Tensor<Float>
     typealias Output = Tensor<Float>
 
     var norm1: BatchNorm<Float>
     var conv1: Conv2D<Float>
     var norm2: BatchNorm<Float>
+    let dropout = Dropout<Float>(probability: 0.3)
     var conv2: Conv2D<Float>
+    
+
+    init(
+        filterShape: (Int, Int, Int, Int),
+        padding: Padding = .same
+    ) {
+        self.norm1 = BatchNorm(featureCount: filterShape.3)
+        self.conv1 = Conv2D(filterShape: filterShape, strides: (1,1), padding: padding)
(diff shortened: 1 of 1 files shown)


brettkoonce:
Thanks for debugging this + adding dropout!  I was working off of [this](https://github.com/meliketoy/wide-resnet.pytorch/blob/30a8b7bdcc7aa3c5115fe2ce4c22c8cf5457b79c/networks/wide_resnet.py#L31) implementation for reference, is that where I missed the identity layer?

Re weight decay, I haven't seen a pattern yet for defining model-specific optimizer values, if you have any suggestions/ideas I'm sure they're welcome!

mikowals:
I just spent a couple of days trying to train the network.  Ultimately I discovered a [bug in _vjpRsqrt](https://github.com/apple/swift/pull/24888) which is called in the gradient calculation of the BatchNorm layer.  

Now it looks like the changes to autodifferentiation code (possibly related to the removal of CotangentVector) is breaking `Array.differentiableReduce`.  I attempted a quick fix replacing all `ContangentVector` references with `TangentVector` but no luck.  I am working in Colab with "(LLVM 082dec2e22, Swift f0e88643cb)".  The Colab session crashes with no errors displayed building the call function with `Blocks.differentiableReduce`, so I assume it is the attempt to autodiff that `call` function that is causing the crash.

Anyway by manually iterating through blocks of a `depthFactor 6, widthFactor 4` model.  I can see that the network does now train as expected.

rxwei:
I’ll take a look at the differentiableReduce breakage tonight.

mikowals:
This pull request runs with the differentiable reduce as updated by Brett.  

The crash  I was seeing was because I had added an L2 loss locally.  It appears to be also related to the  differentiableReduce and the updated autodiff code also but I will try to create a minimal reproduction and file the bug separately.

mikowals:
I have created [TF-533](https://bugs.swift.org/browse/TF-533) for the differentiableReduce bug mentioned above.

mikowals:
replaced by #193.

Issue #165, opened 2019-06-05
Rename 'call' to 'callAsFunction'.


rxwei:
`Layer.call(_:)` has been renamed to `Layer.callAsFunction(_:)` (https://github.com/tensorflow/swift-apis/pull/172) as the result of [SE-0253](https://github.com/apple/swift-evolution/blob/master/proposals/0253-callable.md) proposal review. A toolchain with the new `Layer` definition is [ready for download](https://github.com/tensorflow/swift/blob/master/Installation.md#development-snapshots).

Task: Update all models and layers in this repo to define `func callAsFunction` instead of `func call`.

Shashi456:
@rxwei I think this should be closed, given #166

PR #168, opened 2019-06-07
match api change: call --> callAsFunction

brettkoonce:
see also #165.

rest work with this pr, following are broken:

    - Catch (needs love)
    - CIFAR (tensorhandles)
    - ResNet (tensorhandles)
    - Transformer (_vjpGathering)


brettkoonce:
move to #166

PR #172, opened 2019-06-21
Add Optimization-based Trilateration example

tanmayb123:
I'm still working on some documentation around this, but I'd appreciate any feedback on the code so far!

### DIFF
--- a/Trilateration/Trilateration.swift
+++ b/Trilateration/Trilateration.swift
@@ -1,0 +1,72 @@
+import TensorFlow
+
+struct Coordinate: Differentiable {
+    var x: Tensor<Float>
+    var y: Tensor<Float>
+    
+    @differentiable
+    func euclideanDistance(to other: Coordinate) -> Tensor<Float> {
+        let squaredDiffX = (x - other.x).squared()
+        let squaredDiffY = (y - other.y).squared()
+        let sumSquaredDiff = squaredErrorX + squaredErrorY
+        return sqrt(sumSquaredError)
+    }
+}
+
+extension Coordinate {
+    init(x: Float, y: Float) {
+        self.x = Tensor(x)
+        self.y = Tensor(y)
+    }
+}
+
+struct TrilaterationReferences: Differentiable {
+    struct ReferenceCoordinate: Differentiable {
+        var location: Coordinate
+        var expectedDistance: Tensor<Float>
+        
+        init(location: Coordinate, expectedDistance: Float) {
+            self.location = location
+            self.expectedDistance = Tensor(expectedDistance)
+        }
+    }
+    
+    var ref1: ReferenceCoordinate
+    var ref2: ReferenceCoordinate
+    var ref3: ReferenceCoordinate
+    
+    @differentiable
+    func error(for guess: Coordinate) -> Tensor<Float> {
+        let error1 = (ref1.expectedDistance - guess.euclideanDistance(to: ref1.location)).squared()
+        let error2 = (ref2.expectedDistance - guess.euclideanDistance(to: ref2.location)).squared()
+        let error3 = (ref3.expectedDistance - guess.euclideanDistance(to: ref3.location)).squared()
+        return error1 + error2 + error3
+    }
+}
+
+let references = TrilaterationReferences(
+    ref1: .init(
+        location: .init(x: 2, y: -2),
+        expectedDistance: 4
+    ),
+    ref2: .init(
+        location: .init(x: 10, y: 8),
+        expectedDistance: 10
+    ),
+    ref3: .init(
+        location: .init(x: -1, y: 6),
+        expectedDistance: 5
+    )
+)
+
+var guess = Coordinate(x: 1, y: 1)
+
+for _ in 1...100 {
+    let grad = guess.gradient { guess -> Tensor<Float> in
+        return references.error(for: guess)
+    }
+    guess.x += grad.x * -0.1
+    guess.y += grad.y * -0.1
+}
+
+print(guess) // Expected to be [2, 2]
--- a/Trilateration/Trilateration.swift
+++ b/Trilateration/Trilateration.swift
@@ -8,8 +8,8 @@ struct Coordinate: Differentiable {
     func euclideanDistance(to other: Coordinate) -> Tensor<Float> {
         let squaredDiffX = (x - other.x).squared()
         let squaredDiffY = (y - other.y).squared()
-        let sumSquaredDiff = squaredErrorX + squaredErrorY
-        return sqrt(sumSquaredError)
+        let sumSquaredDiff = squaredDiffX + squaredDiffY
+        return sqrt(sumSquaredDiff)
     }
 }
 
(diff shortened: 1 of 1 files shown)


tanmayb123:
@rxwei How hard would it be to define a differentiable map? If I'm not mistake, all the VJP would need to do is find the gradient of the differentiable function passed to map and apply it element wise to the array?

tanmayb123:
For `Tensor<Float>` specifically, maybe something like this (rough code):

```
@differentiable(vjp: _vjpDMap)
func dMap(_ f: @escaping @differentiable (Tensor<Float>) -> (Tensor<Float>), _ x: [Tensor<Float>]) -> Array<Tensor<Float>>.DifferentiableView {
    return Array.DifferentiableView(x.map({ f($0) }))
}

func _vjpDMap(_ f: @escaping @differentiable (Tensor<Float>) -> (Tensor<Float>), _ x: [Tensor<Float>]) -> (Array<Tensor<Float>>.DifferentiableView, (Array<Tensor<Float>>.DifferentiableView) -> (Array<Tensor<Float>>.DifferentiableView)) {
    return (dMap(f, x), { v in
        var values: Array<Tensor<Float>> = []
        let grad = gradient(of: f)
        for i in 0..<v.base.count {
            values.append(grad(x[i]) * v.base[i])
        }
        return Array.DifferentiableView(values)
    })
}
```

Trilateration/Trilateration.swift
```diff
@@ -0,0 +1,72 @@
...
+        location: .init(x: -1, y: 6),
+        expectedDistance: 5
+    )
+)
+
+var guess = Coordinate(x: 1, y: 1)
+
+for _ in 1...100 {
+    let grad = guess.gradient { guess -> Tensor<Float> in
+        return references.error(for: guess)
```
rxwei:
```suggestion
        references.error(for: guess)
```

Trilateration/Trilateration.swift
```diff
@@ -0,0 +1,72 @@
+import TensorFlow
```
rxwei: Could you please add a license header?

Trilateration/Trilateration.swift
```diff
@@ -0,0 +1,72 @@
...
+        var location: Coordinate
+        var expectedDistance: Tensor<Float>
+        
+        init(location: Coordinate, expectedDistance: Float) {
+            self.location = location
+            self.expectedDistance = Tensor(expectedDistance)
+        }
+    }
+    
+    var ref1: ReferenceCoordinate
```
rxwei: ref -> reference
```suggestion
    var reference1: ReferenceCoordinate
```

Trilateration/Trilateration.swift
```diff
@@ -0,0 +1,72 @@
+import TensorFlow
+
+struct Coordinate: Differentiable {
+    var x: Tensor<Float>
```
rxwei: Since this is supposed to be initialized with a scalar, we can just store two `Float`s, right?

rxwei:
> For `Tensor<Float>` specifically, maybe something like this (rough code):
> 
> ```
> @differentiable(vjp: _vjpDMap)
> func dMap(_ f: @escaping @differentiable (Tensor<Float>) -> (Tensor<Float>), _ x: [Tensor<Float>]) -> Array<Tensor<Float>>.DifferentiableView {
>     return Array.DifferentiableView(x.map({ f($0) }))
> }
> 
> func _vjpDMap(_ f: @escaping @differentiable (Tensor<Float>) -> (Tensor<Float>), _ x: [Tensor<Float>]) -> (Array<Tensor<Float>>.DifferentiableView, (Array<Tensor<Float>>.DifferentiableView) -> (Array<Tensor<Float>>.DifferentiableView)) {
>     return (dMap(f, x), { v in
>         var values: Array<Tensor<Float>> = []
>         let grad = gradient(of: f)
>         for i in 0..<v.base.count {
>             values.append(grad(x[i]) * v.base[i])
>         }
>         return Array.DifferentiableView(values)
>     })
> }
> ```

Yep, this is right. We should define it in an extension for `Array` like [`differentiableReduce(_:_:)`].

tanmayb123:
> > For `Tensor<Float>` specifically, maybe something like this (rough code):
> > ```
> > @differentiable(vjp: _vjpDMap)
> > func dMap(_ f: @escaping @differentiable (Tensor<Float>) -> (Tensor<Float>), _ x: [Tensor<Float>]) -> Array<Tensor<Float>>.DifferentiableView {
> >     return Array.DifferentiableView(x.map({ f($0) }))
> > }
> > 
> > func _vjpDMap(_ f: @escaping @differentiable (Tensor<Float>) -> (Tensor<Float>), _ x: [Tensor<Float>]) -> (Array<Tensor<Float>>.DifferentiableView, (Array<Tensor<Float>>.DifferentiableView) -> (Array<Tensor<Float>>.DifferentiableView)) {
> >     return (dMap(f, x), { v in
> >         var values: Array<Tensor<Float>> = []
> >         let grad = gradient(of: f)
> >         for i in 0..<v.base.count {
> >             values.append(grad(x[i]) * v.base[i])
> >         }
> >         return Array.DifferentiableView(values)
> >     })
> > }
> > ```
> 
> Yep, this is right. We should define it in an extension for `Array` like [`differentiableReduce(_:_:)`](https://gist.github.com/rxwei/2739515f77a62d26add66947cb179911).

For some reason, I can only get it to work when the generic type of the function defines the `Scalar` type of the Tensor. I can't replace `Tensor<Float>` with a generic type with a constraint of conforming to Differentiable like so:

```swift
@differentiable(vjp: _vjpDifferentiableMap)
func differentiableMap<T: Differentiable, U: Differentiable>(_ f: @escaping @differentiable (T) -> (U), _ x: [T]) -> Array<U>.DifferentiableView {
    return Array.DifferentiableView(x.map({ f($0) }))
}

func _vjpDifferentiableMap<T: Differentiable, U: Differentiable>(_ f: @escaping @differentiable (T) -> (U), _ x: [T]) -> (Array<U>.DifferentiableView, (Array<U>.DifferentiableView) -> (Array<T>.DifferentiableView)) {
    return (differentiableMap(f, x), { v in Array.DifferentiableView(Array<T>()) }) // todo: real vip logic
}
```

rxwei:
`Tensor` is differentiable where `Scalar: TensorFlowFloatingPoint` (not just `Differentiable`).

tanmayb123:
No, I'm trying to define a map function that works on arrays of any type, as long as it's differentiable - not just Tensors.. Like this:

```swift
struct Coord: Differentiable {
    var x: Tensor<Float>
    var y: Tensor<Float>
    // insert Euclidean distance function here
}

var ref = Coord(x: 2, y: -2)
var others: [Coord] = [.init(x: 3, y: -1), .init(x: 10, y: 8)]

print(others.differentiableMap({ $0.distance(to: ref) }))
```

BradLarson:
We're revisiting this after a long time away (sorry about that), and it looks like this might be able to be incorporated with a few small changes.

In addition to what's been suggested (adding a license header, converting Coordinate to use scalar parameters), could I also request that it be added as an executable target in the project's Package.swift so that it can be run?

shabalind:
Hi @tanmayb123, thanks a lot for the contribution, but given a lack of activity on this PR we are closing it for the time being. Feel free to resubmit!

PR #180, opened 2019-07-24
Replace Autoencoder's tanh activation with sigmoid

### LINKED ISSUE #179: Autoencoder's activation should be sigmoid
It's using `tanh` in output layer.

https://github.com/tensorflow/swift-models/blob/08c80a5d797ed771df899169914f84d8dcaa9eec/Autoencoder/main.swift#L93-L94

But input image is [0, 1] ranged.

https://github.com/tensorflow/swift-models/blob/08c80a5d797ed771df899169914f84d8dcaa9eec/Autoencoder/main.swift#L74

In v0.4.0-rc2, if we use sigmoid, this model doesn't learn maybe because there's a bug I fixed but not included yet.
https://github.com/tensorflow/swift-apis/pull/341

It should be replaced in next release.

t-ae:
This PR resolves #179 

In v0.4.0-rc2 `sigmoid` has a bug in its gradient. So the model doesn't learn correctly.
It'll be fixed in future release.

close #179

### DIFF
--- a/Autoencoder/main.swift
+++ b/Autoencoder/main.swift
@@ -91,7 +91,7 @@ struct Autoencoder: Layer {
     var decoder2 = Dense<Float>(inputSize: 12, outputSize: 64, activation: relu)
     var decoder3 = Dense<Float>(inputSize: 64, outputSize: 128, activation: relu)
     var decoder4 = Dense<Float>(inputSize: 128, outputSize: imageHeight * imageWidth,
-        activation: tanh)
+        activation: sigmoid)
 
     @differentiable
     func callAsFunction(_ input: Input) -> Output {


BradLarson:
As a note, I reverted this change in the following pull request:

https://github.com/tensorflow/swift-models/pull/182

because I found that the sigmoid activation still generated nonsense during training on 0.4.0-rc4 and the nightly build I had here. Do you see otherwise?

t-ae:
Weird.
In swift-apis, this test is passing.
https://github.com/tensorflow/swift-apis/blob/2f75a2dbe9eedddc1178ce0a2fff9a8b7a595729/Tests/TensorFlowTests/TensorAutoDiffTests.swift#L233-L236

But It returns wrong values on REPL.

```
$ swift
Welcome to Swift version 5.1-dev (LLVM af1f73e9e9, Swift 405b76fca1).
Type :help for assistance.
  1> import TensorFlow
  2> func f(a: Tensor<Float>) -> Tensor<Float> { sigmoid(a).sum() } 
  3. print(gradient(at: [-1, 0, 1], in: f))
[-2.0,  0.0,  0.0]
```

t-ae:
I copied sigmoid definition below, then autoencoder learned correctly.
https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Operators/Math.swift#L1038-L1052

Perhaps my sigmoid fix is not included on 0.4.0-rc4...?
Which commit is it using?

t-ae:
I also tried with Development Snapshots on macOS.
https://storage.googleapis.com/swift-tensorflow/mac/swift-tensorflow-DEVELOPMENT-2019-07-25-a-osx.pkg

Then autoencoder generates MNIST-like images even if I use sigmoid as activation.
Seems 0.4.0-rc4 and Development Snapshots uses different commit of swift-apis.

@BradLarson 
The nightly build you mentioned is ubuntu one?
I suspect it doesn't use latest commit of swift-apis.

rxwei:
@BradLarson 0.4 does not include the latest swift-apis. In fact it’s quite old :)

PR #182, opened 2019-07-28
First steps in repository reorganization: extracting common MNIST dataset code

BradLarson:
This is a start on implementing some of the elements of [the reorganization proposed](https://docs.google.com/document/d/1eIFrogM7HYn-hIMvrVcoOGjC15KCxDEyn60GNX1_DJI) for this repository.

As a first step, this extracts the MNIST dataset, for which duplicate code and data existed across three separate examples, and places it into a shared Dataset library. This also reworks the generic "MNIST" example into an example training the classic LeNet-5 model against MNIST. LeNet-5 has been created as a standalone model, with the LeNet-MNIST example bringing together the MNIST dataset and the LeNet model with a custom training loop.

This illustrates the general organization that I have in mind for the remaining image-based models, and is a first step to demonstrate and test the process.

(Sorry for the odd initial pull request title and commentary, not sure what happened there.)


t-ae:
CIFAR examples use `Dataset` but MNIST examples don't.
https://github.com/tensorflow/swift-models/blob/master/CIFAR/Data.swift#L142-L148
Are there any reason not to use it?

Also, MNIST examples don't shuffle data.
I think they'll be more plausible examples with shuffling.
(`Dataset` has shuffling method, so it makes sense to use it.)

Datasets/MNIST/MNIST.swift
```diff
@@ -0,0 +1,103 @@
...
+        }
+        let data = try! Data(contentsOf: filePath, options: [])
+        return [UInt8](data)
+    }
+    print("File not found: \(path)")
+    exit(-1)
+}
+
+/// Reads MNIST images and labels from specified file paths.
+func readMNIST(imagesFile: String, labelsFile: String, flatten: Bool, normalize: Bool) -> (images: Tensor<Float>,
```
rxwei: * Fit within 100 columns. Break lines according to Google Swift Style Guide.
* I'd suggest using gerunds for boolean flags: `flattening` and `normalizing`.

Datasets/MNIST/MNIST.swift
```diff
@@ -0,0 +1,103 @@
...
+    print("Reading data from files: \(imagesFile), \(labelsFile)")
+    let images = readFile(imagesFile, possibleDirectories: [".", "./Datasets/MNIST"]).dropFirst(16).map(Float.init)
+    let labels = readFile(labelsFile, possibleDirectories: [".", "./Datasets/MNIST"]).dropFirst(8).map(Int32.init)
+    let rowCount = labels.count
+    let imageHeight = 28, imageWidth = 28
+
+    print("Constructing data tensors.")
+                                                                            
+    if flatten {
+        if normalize {
```
rxwei: * Instead of nesting multiple conditional statements, how about:
    ```swift
    var images = Tensor(shape: [rowCount, imageHeight * imageWidth], scalars: images) / 255.0
    if normalizing {
        images = images * 2 - 1
    }
    return (images: images, labels: Tensor(labels))
    ```
* Fix indentation. Indent by 4 instead of 8.

Datasets/MNIST/MNIST.swift
```diff
@@ -0,0 +1,103 @@
...
+    public let trainLabels: Tensor<Int32>
+    public let testImages: Tensor<Float>
+    public let testLabels: Tensor<Int32>
+
+    public let trainingSize: Int
+    public let testSize: Int
+
+    let batchSize: Int
+
+    public init(batchSize: Int, flatten: Bool = false, normalize: Bool = false) {
```
rxwei:
```suggestion
    public init(batchSize: Int, flattening: Bool = false, normalizing: Bool = false) {
```

Models/ImageClassification/LeNet-5.swift
```diff
@@ -0,0 +1,44 @@
...
+// Original Paper:
+// "Gradient-Based Learning Applied to Document Recognition"
+// Yann LeCun, Léon Bottou, Yoshua Bengio, and Patrick Haffner
+// http://yann.lecun.com/exdb/publis/pdf/lecun-01a.pdf
+//
+// Note: this implementation connects all the feature maps in the second convolutional layer.
+// Additionally, ReLU is used instead of sigmoid activations.
+
+public struct LeNet: Layer {
+    var conv1 = Conv2D<Float>(filterShape: (5, 5, 1, 6), padding: .same, activation: relu)
```
rxwei: How about making these fields public?

Models/ImageClassification/LeNet-5.swift
```diff
@@ -0,0 +1,44 @@
...
+    var conv2 = Conv2D<Float>(filterShape: (5, 5, 6, 16), activation: relu)
+    var pool2 = AvgPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    var flatten = Flatten<Float>()
+
+    var fc1 = Dense<Float>(inputSize: 400, outputSize: 120, activation: relu)
+    var fc2 = Dense<Float>(inputSize: 120, outputSize: 84, activation: relu)
+    var fc3 = Dense<Float>(inputSize: 84, outputSize: 10, activation: softmax)
+
+    public init() {
+    }
```
rxwei: Move `}` to the end of the previous line.

Datasets/MNIST/MNIST.swift
BradLarson: I've reworked the wording and applied swift-format throughout this file.

Datasets/MNIST/MNIST.swift
BradLarson: Good call on the restructuring here. I've done that. No idea what happened with the indentation, but that's been handled by swift-format now.

Models/ImageClassification/LeNet-5.swift
BradLarson: I've made all the internal layers public. I had originally made them private to encapsulate the internal functioning of the model, but it makes sense to have them be public. Eventually, people will want to freeze individual layers internal to a model during training, and they'll need to get at them to do so.

I'll make this the style going forward for models.

BradLarson:
@t-ae - Good points about Dataset. The current examples were largely developed independent of one another, so there are some inconsistencies here that we're working to resolve. One of those is the use of Dataset. I also did notice the lack of shuffling, which is something we'll want to address.

I do think we'll want to work towards using the Dataset type in the datasets. My first step here was to exactly replicate functionality and pull out repeated code. Next, I'd like to do the same for CIFAR-10 in the two examples that use that, and after that see what can be done to make these two share an interface and more functionality.

Datasets/MNIST/MNIST.swift
```diff
@@ -0,0 +1,104 @@
...
+    print("File not found: \(path)")
+    exit(-1)
+}
+
+/// Reads MNIST images and labels from specified file paths.
+func readMNIST(imagesFile: String, labelsFile: String, flattening: Bool, normalizing: Bool) -> (
+    images: Tensor<Float>,
+    labels: Tensor<Int32>
+) {
+    print("Reading data from files: \(imagesFile), \(labelsFile)")
```
rxwei: End messages with a period like other print statements.
```suggestion
    print("Reading data from files: \(imagesFile), \(labelsFile).")
```

Examples/LeNet-MNIST/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Setup
+
+To begin, you'll need the [latest version of Swift for
+TensorFlow](https://github.com/tensorflow/swift/blob/master/Installation.md)
+installed. Make sure you've added the correct version of `swift` to your path.
+
+To train the model, run:
+
+```
```
rxwei:
```suggestion
```sh
```

GAN/README.md
```diff
@@ -16,6 +16,8 @@ To begin, you'll need the [latest version of Swift for
 TensorFlow](https://github.com/tensorflow/swift/blob/master/Installation.md)
 installed. Make sure you've added the correct version of `swift` to your path.
 
+This example requires matplotlib and numpy to be installed, for use in image output.
```
rxwei:
```suggestion
This example requires Matplotlib and NumPy to be installed, for use in image output.
```
Nit: Use official product names, "Matplotlib" and "NumPy".

Datasets/MNIST/MNIST.swift
```diff
@@ -0,0 +1,104 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import TensorFlow
+
+public struct MNIST {
+    public let trainImages: Tensor<Float>
```
rxwei: I'm not sure "train images" is the best term we should use in a Swift API design context. Albeit common in Python, "train images" which intends to be a noun phrase starts with a bare form verb that doesn't read fluently as recommended by Swift API Design Guidelines. How about using "training images" as the standard instead?

Datasets/MNIST/MNIST.swift
```diff
@@ -0,0 +1,104 @@
...
+    public let trainLabels: Tensor<Int32>
+    public let testImages: Tensor<Float>
+    public let testLabels: Tensor<Int32>
+
+    public let trainingSize: Int
+    public let testSize: Int
+
+    let batchSize: Int
```
rxwei: I believe this should also be public, because:
1. The initializer takes this as an argument. Since the user already explicitly provides it, it's weird to hide it from the user.
2. Some pilot users like Jeremy Howard have strongly recommend making model properties be public, which is considered important for research flexibility.

```suggestion
    public let batchSize: Int
```

Package.swift
```diff
@@ -9,27 +9,32 @@ let package = Package(
...
+        .target(name: "Datasets", path: "Datasets"),
+        .target(name: "Autoencoder", dependencies: ["Datasets"], path: "Autoencoder"),
         .target(name: "CIFAR", path: "CIFAR"),
         .target(name: "Catch", path: "Catch"),
         .target(name: "Gym-FrozenLake", path: "Gym/FrozenLake"),
         .target(name: "Gym-CartPole", path: "Gym/CartPole"),
         .target(name: "Gym-Blackjack", path: "Gym/Blackjack"),
-        .target(name: "MNIST", path: "MNIST"),
+        .target(name: "LeNet-MNIST", dependencies: ["ImageClassificationModels", "Datasets"], 
+				path: "Examples/LeNet-MNIST"),
```
rxwei: Fix indentation.

Datasets/MNIST/MNIST.swift
BradLarson: You're right, that name bothered me. I've left "testImages" as-is, because I'm thinking about the name as it would be in a sentence like "these are training images" or "these are test images".

Datasets/MNIST/MNIST.swift
BradLarson: I'll keep those guidelines in mind, as well.

Datasets/MNIST/MNIST.swift
rxwei: “Test images” sounds totally fine to me as a noun phrase. Thanks!

PR #188, opened 2019-08-02
Add JPEG loading / saving via Raw TF operations, removing Matplotlib dependencies

BradLarson:
This adds a wrapper Image struct for encapsulating JPEG (and potentially other formats) image loading and saving, using Raw TensorFlow operators. This allows for JPEG loading and saving without any external dependencies.

To date, Matplotlib had been used for image export, but this provides a means of performing this directly in Swift. As a result, the Matplotlib (and Python, in general) dependencies have been removed for the Autoencoder and GAN examples.

This is a first step for providing general image dataset loading for training and tests.

### DIFF
--- a/Support/FileManagement.swift
+++ b/Support/FileManagement.swift
@@ -14,11 +14,10 @@
 
 import Foundation
 
-public func createDirectoryIfMissing(path: String) {
-    if !FileManager.default.fileExists(atPath: path) {
-        try! FileManager.default.createDirectory(
-            atPath: path,
-            withIntermediateDirectories: false,
-            attributes: nil)
-    }
+public func createDirectoryIfMissing(path: String) throws {
+    guard !FileManager.default.fileExists(atPath: path) else { return }
+    try FileManager.default.createDirectory(
+        atPath: path,
+        withIntermediateDirectories: false,
+        attributes: nil)
 }
--- a/Support/FileManagement.swift
+++ b/Support/FileManagement.swift
@@ -14,7 +14,7 @@
 
 import Foundation
 
-public func createDirectoryIfMissing(path: String) throws {
+public func createDirectoryIfMissing(at path: String) throws {
     guard !FileManager.default.fileExists(atPath: path) else { return }
     try FileManager.default.createDirectory(
         atPath: path,
--- a/Support/FileManagement.swift
+++ b/Support/FileManagement.swift
@@ -1,0 +1,24 @@
+// Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+public func createDirectoryIfMissing(path: String) {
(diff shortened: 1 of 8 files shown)


Support/Image.swift
```diff
@@ -0,0 +1,91 @@
...
+        if byteOrdering == .bgr {
+            self.imageData = .uint8(
+                data: Raw.reverse(loadedJpeg, dims: Tensor<Bool>([false, false, false, true])))
+        } else {
+            self.imageData = .uint8(data: loadedJpeg)
+        }
+    }
+
+    public func save(to url: URL, quality: Int64 = 95) {
+        // Currently only saving in grayscale
```
rxwei: End all comments with a period.

Support/Image.swift
```diff
@@ -0,0 +1,91 @@
...
+import Foundation
+import TensorFlow
+
+public struct Image {
+    public enum ByteOrdering {
+        case bgr
+        case rgb
+    }
+
+    enum ImageTensorFormat {
```
rxwei:
```suggestion
    enum ImageTensor {
```

I think this should be renamed to `ImageTensor` because it is semantically just a tensor except that the value depends on a format tag. It itself isn't the format -- only the enum case tag is.

Support/Image.swift
```diff
@@ -0,0 +1,91 @@
...
+                    images: Tensor<Float>([data]),
+                    size: Tensor<Int32>([Int32(size.0), Int32(size.1)])))
+        }
+
+    }
+}
+
+public func saveImage(tensor: Tensor<Float>, size: (Int, Int), directory: String, name: String) {
+    createDirectoryIfMissing(path: directory)
+
```
rxwei: Remove the empty line since this function is small?

Support/Image.swift
```diff
@@ -0,0 +1,91 @@
...
+            return Image(
+                tensor: Raw.resizeBilinear(
+                    images: Tensor<Float>([data]),
+                    size: Tensor<Int32>([Int32(size.0), Int32(size.1)])))
+        }
+
+    }
+}
+
+public func saveImage(tensor: Tensor<Float>, size: (Int, Int), directory: String, name: String) {
```
rxwei:
```suggestion
public func saveImage(_ tensor: Tensor<Float>, size: (Int, Int), directory: String, name: String) {
```
The `tensor` argument label is merely repeating the argument type information, so it should be omitted according to the Swift API Design Guidelines.

Support/FileManagement.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+public func createDirectoryIfMissing(path: String) {
```
rxwei: Should this throw?

Support/FileManagement.swift
```diff
@@ -0,0 +1,24 @@
...
+    if !FileManager.default.fileExists(atPath: path) {
```
rxwei: Consider flipping the condition to early-return.

BradLarson:
Good call on making failure of directory creation throw. I believe I've incorporated the remainder of the changes here.

Support/FileManagement.swift
```diff
@@ -14,11 +14,10 @@
...
 import Foundation
 
-public func createDirectoryIfMissing(path: String) {
-    if !FileManager.default.fileExists(atPath: path) {
-        try! FileManager.default.createDirectory(
-            atPath: path,
-            withIntermediateDirectories: false,
-            attributes: nil)
-    }
+public func createDirectoryIfMissing(path: String) throws {
```
rxwei: How about making the argument label be `at:`?

BradLarson: Done.

PR #192, opened 2019-08-07
[Models] Fix enumeration of blocks in `ResidualBasicBlockStack` `init`

jon-tow:
Small fix for correctly populating `blocks` with `blockCount` number of `ResidualBasicBlock`s and `ResidualIdentityBlock`s as opposed to `blockCount - 1`.

### DIFF
--- a/Models/ImageClassification/ResNet50.swift
+++ b/Models/ImageClassification/ResNet50.swift
@@ -99,7 +99,7 @@ public struct ResidualBasicBlockStack: Layer {
     public var blocks: [ResidualBasicBlock] = []
 
     public init(featureCounts: (Int, Int, Int, Int), kernelSize: Int = 3, blockCount: Int) {
-        for _ in 1...blockCount {
+        for _ in 0..<blockCount {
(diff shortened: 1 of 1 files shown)


Models/ImageClassification/ResNet50.swift
```diff
@@ -99,7 +99,7 @@ public struct ResidualBasicBlockStack: Layer {
     public var blocks: [ResidualBasicBlock] = []
 
     public init(featureCounts: (Int, Int, Int, Int), kernelSize: Int = 3, blockCount: Int) {
-        for _ in 1..<blockCount {
+        for _ in 1...blockCount {
```
rxwei: How about 0..<blockCount instead?

jon-tow: I guess it'd keep Dijkstra happy. Sounds good to me.

BradLarson:
Will this also need to be changed in `ResidualIdentityBlockStack`?

jon-tow:
@BradLarson Sorry I didn't notice that! Thanks for letting me know.

PR #194, opened 2019-08-08
Remove deprecated allDifferentiableVariables

BradLarson:
This removes the now-deprecated allDifferentiableVariables from the various training loops. This is a follow-on from this change in swift-apis: https://github.com/tensorflow/swift-apis/pull/419 .

### DIFF
--- a/Autoencoder/main.swift
+++ b/Autoencoder/main.swift
@@ -93,6 +93,6 @@ for epoch in 1...epochCount {
             return meanSquaredError(predicted: image, expected: x)
         }
 
-        optimizer.update(&autoencoder.allDifferentiableVariables, along: 𝛁model)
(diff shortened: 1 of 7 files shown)


Ayush517:
I tried running the updated code on Colab without `allDifferentiableVariable` and it gave me the following error:

> error: cannot convert value of type 'Autoencoder' to expected argument type 'Autoencoder.AllDifferentiableVariables'
>         optimizer.update(&autoencoder, along: 𝛁model)

It runs absolutely fine when I added updated `optimizer.update(&autoencoder, along: 𝛁model)` to `optimizer.update(&autoencoder.allDifferentiableVariables, along: 𝛁model)`

rxwei:
Colab is running Swift for TensorFlow 0.4, not the latest implementation.

Issue #199, opened 2019-08-24
Convert MNIST to use 'Sequential'


rxwei:

BradLarson:
Would it make sense to convert all linear models to use Sequential, or leave some one way and some the other as examples / test cases?

Should we try to apply Sequential to ResNet and other branching models?

rxwei:
I think it's fine to use `Sequential` wherever it's applicable, since it's already design-reviewed.  The reason I'm opening this ticket just for MNIST is because I wanted to see how things shape up in practice with the smallest step possible :)

Shashi456:
@BradLarson, what other linear models would you want to be converted to sequential? Now that we've decided not to convert the structure, should we still go ahead with that move as well?

PR #200, opened 2019-08-25
Convert MNIST classifier to use sequential

Shashi456:
Fix: #199 
Just a few questions, my first time contributing to this repo. Do I build this in the same `swift build` and `swift test` fashion?
Also now that the model was updated, do I remove the Le-Net file, or should I change the Le-net file to use Sequential?

### DIFF
--- a/Examples/LeNet-MNIST/main.swift
+++ b/Examples/LeNet-MNIST/main.swift
@@ -20,7 +20,16 @@ let epochCount = 12
 let batchSize = 128
 
 let dataset = MNIST(batchSize: batchSize)
-var classifier = LeNet()
+var classifier = Sequential {
+    Conv2D<Float>(filterShape: (5, 5, 1, 6), padding: .same, activation: relu)
+    AvgPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    Conv2D<Float>(filterShape: (5, 5, 6, 16), activation: relu)
+    AvgPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    Flatten<Float>()
+    Dense<Float>(inputSize: 400, outputSize: 120, activation: relu)
+    Dense<Float>(inputSize: 120, outputSize: 84, activation: relu)
+    Dense<Float>(inputSize: 84, outputSize: 10, activation: softmax)
+    }
 
 let optimizer = SGD(for: classifier, learningRate: 0.1)
 
@@ -73,7 +82,7 @@ for epoch in 1...epochCount {
     print("""
           [Epoch \(epoch)] \
           Training Loss: \(trainStats.totalLoss), \
-          Training Accuracy: \(trainStats.correctGuessCount)/\(trainStats.totalGuessCount) \ 
+          Training Accuracy: \(trainStats.correctGuessCount)/\(trainStats.totalGuessCount) \
           (\(trainAccuracy)), \
           Test Loss: \(testStats.totalLoss), \
           Test Accuracy: \(testStats.correctGuessCount)/\(testStats.totalGuessCount) \


Examples/LeNet-MNIST/main.swift
```diff
@@ -20,7 +20,16 @@ let epochCount = 12
...
+var classifier = Sequential {
+    Conv2D<Float>(filterShape: (5, 5, 1, 6), padding: .same, activation: relu)
+    AvgPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    Conv2D<Float>(filterShape: (5, 5, 6, 16), activation: relu)
+    AvgPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    Flatten<Float>()
+    Dense<Float>(inputSize: 400, outputSize: 120, activation: relu)
+    Dense<Float>(inputSize: 120, outputSize: 84, activation: relu)
+    Dense<Float>(inputSize: 84, outputSize: 10, activation: softmax)
+    }
```
rxwei:
```suggestion
}
```

Examples/LeNet-MNIST/main.swift
```diff
@@ -20,7 +20,16 @@ let epochCount = 12
 let batchSize = 128
 
 let dataset = MNIST(batchSize: batchSize)
-var classifier = LeNet()
+var classifier = Sequential {
```
rxwei:
```suggestion
// The LeNet-5 model, equivalent to `LeNet` in `ImageClassificationModels`.
var classifier = Sequential {
```

rxwei:
> Also now that the model was updated, do I remove the Le-Net file, or should I change the Le-net file to use Sequential?

Please keep the LeNet file. LeNet should be kept as is because other models and clients may use it. A `struct` implementation is better than `Sequential` in various ways, one of which is that users can access properties (component layers) easily for debugging purposes.

The MNIST model is simple enough that implementing the model using `Sequential` makes it very familiar to people who come from Keras.

rxwei:
Please also remove `import ImageClassificationModels` since it's no longer used in this model.

Shashi456:
There's a curl error:
```
curl: (22) The requested URL returned error: 404
The command '/bin/sh -c curl -fSsL $swift_tf_url -o swift.tar.gz     && mkdir usr     && tar -xzf swift.tar.gz --directory=usr --strip-components=1     && rm swift.tar.gz' returned a non-zero code: 22

```

Shashi456:
@rxwei are there any other models that need to ported to Sequential?

BradLarson:
I guess I'd interpreted the request as asking for the LeNet-5.swift model to be reworked to be Sequential, and everything else remain the same. I'd like to keep the models separate from the example code, for organization and to allow for reuse in testing or external projects.

Would it make sense to have a separate Sequential LeNet implementation within the LeNet-5.swift file, and then call that implementation here? That way, when [this pull request](https://github.com/tensorflow/swift-models/pull/198) lands, the LeNet-Sequential (or however the model is identified) can be placed in the inference tests alongside the traditionally specified LeNet model. That would provide for a side-by-side comparison.

Shashi456:
@BradLarson do you mean something like this?

Shashi456:
@BradLarson would that look something like 
``` swift
public func LeNetSequential() {
     return Sequential(Layers)
} 
```

Shashi456:
I'm +1 on Richard's mentioned direction, We don't modify `ImageClassificationModels` and let the struct remain the same in case people want to use it. 

While we go on and change the LeNet usage in the example to `Sequential`, because I believe that showing the entire LeNet makes it readable and allow referees to check the LeNet implementation.  

Finally, to make the docs thorough, we could offer a short example on how to use it in the `Lenet-5.swift` file.

PR #204, opened 2019-08-28
Update squeezenet and add V1.1 version

Shashi456:
So the default activation in conv layers is identity, while the squeezenet paper mentions a relu activation after the 1st and last convolution.

Also adding the v1.1 version of squeezenet, its 2.4x less computation without sacrificing accuracy. For more check [here](https://github.com/DeepScale/SqueezeNet/tree/master/SqueezeNet_v1.1).

### DIFF
--- a/Models/ImageClassification/SqueezeNet.swift
+++ b/Models/ImageClassification/SqueezeNet.swift
@@ -177,6 +177,7 @@ public struct SqueezeNetV1_1: Layer {
         let fired1 = convolved1.sequenced(through: fire2, fire3, maxPool3, fire4, fire5)
         let fired2 = fired1.sequenced(through: maxPool5, fire6, fire7, fire8, fire9)
         let convolved2 = fired2.sequenced(through: dropout, conv10, avgPool10)
+            .reshaped(to: [input.shape[0], conv10.filter.shape[3]])
         return convolved2
     }
 }
--- a/Models/ImageClassification/SqueezeNet.swift
+++ b/Models/ImageClassification/SqueezeNet.swift
@@ -53,8 +53,11 @@ public struct Fire: Layer {
 }
 
 public struct SqueezeNetV1_0: Layer {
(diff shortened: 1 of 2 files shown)


BradLarson:
Good catch on the missing ReLUs in the model. I missed that in the initial review.

Because this renames SqueezeNet to SqueezeNetV1_0, it will break the newly added inference tests here:     https://github.com/tensorflow/swift-models/blob/master/Tests/ImageClassificationTests/Inference.swift#L95 . Could you also rename the model in the inference test, and maybe add the new SqueezeNetV1_0 as a new test there?

Models/ImageClassification/SqueezeNet.swift
```diff
@@ -114,3 +115,67 @@ public struct SqueezeNet: Layer {
         return convolved2
     }
 }
+
+public struct SqueezeNetV1_1: Layer {
+    public var conv1 = Conv2D<Float>(filterShape: (3, 3, 3, 96), strides: (2, 2), padding: .same,
```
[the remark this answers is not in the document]
Shashi456: Yeah, it's 64, i wonder how I missed that.

Shashi456:
@BradLarson done :)

Models/ImageClassification/SqueezeNet.swift
```diff
@@ -115,3 +116,68 @@ public struct SqueezeNet: Layer {
         return convolved2
     }
 }
+
+public struct SqueezeNetV1_1: Layer {
+    public var conv1 = Conv2D<Float>(filterShape: (3, 3, 3, 64), strides: (2, 2), padding: .same,
```
rxwei:
```suggestion
    public var conv1 = Conv2D<Float>(
        filterShape: (3, 3, 3, 64), 
        strides: (2, 2), 
        padding: .same,
        activation: relu)
```
Please follow the Google Swift Style Guide or run swift-format.

Shashi456: swift-format doesn't modify it this way :P, I'll update it as you said.

PR #205, opened 2019-08-29
Update deprecated APIs in Catch

Shashi456:
[swift-apis/#497](https://github.com/tensorflow/swift-apis/pull/497) deprecates layer apis, now that we are moving onward to v0.5. This PR updates the catch file which has a deprecated dense layer.

### DIFF
--- a/Catch/main.swift
+++ b/Catch/main.swift
@@ -20,8 +20,6 @@ import TensorFlow
 //   gradients). The algorithm should perform some tensor computation (not a
 //   purely table-based approach).
 
-var rng = PhiloxRandomNumberGenerator(seed: 0xdeadbeef)
-
 extension Sequence {
     /// Returns elements' descriptions joined by a separator.
     func description(joinedBy separator: String) -> String {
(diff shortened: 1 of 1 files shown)


BradLarson:
This removes the constant seed for the random number generator. Usually, we'd only use a constant seed for testing where we wanted deterministic behavior, are there any side effects for removing this seed in this example?

rxwei:
This model is not being unit-tested, so I think removing the seed is fine. Finding a TensorFlow seed would be good, but it's a lower priority than deprecating legacy APIs for a new toolchain release. Let's check this in and resolve this later.

dan-zheng:
CI failed:
```
/swift-models/Catch/main.swift:101:48: error: use of unresolved identifier 'rng'
        let id = Int.random(in: 0..<3, using: &rng)
                                               ^~~
/swift-models/Catch/main.swift:160:68: error: use of unresolved identifier 'rng'
        let randomColumn = Int.random(in: 0..<columnCount, using: &rng)
                                                                   ^~~
```

rxwei:
@Shashi456 Please make sure to always *compile and test* the code before you push. Thanks!

rxwei:
tensorflow/swift-apis#497 is now unblocked. Thank you @Shashi456 !

PR #208, opened 2019-09-20
Fix '@noDerivative' warnings.

rxwei:


ematejska:
Actually, I think this might have to go the tensorflow-0.5 branch.  That's what we're testing from for the release

ematejska:
The master one is still on 0.4 and that doesn't have warnings as far as I know.

rxwei:
> Actually, I think this might have to go the tensorflow-0.5 branch. That's what we're testing from for the release

How about merging into master and cherry-picking into 0.5?

PR #213, opened 2019-09-28
DenseNet Implementaion

Andr0id100:
Any feedback would be awesome.
@rxwei @BradLarson

### DIFF
--- a/Models/ImageClassification/DenseNet.swift
+++ b/Models/ImageClassification/DenseNet.swift
@@ -82,7 +82,6 @@ extension DenseNet {
     }
 
     /// A pair of a 1x1 `Conv` layer and a 3x3 `Conv` layer.
-    // make a dense block
     public struct ConvPair: Layer {
         public var conv1x1: Conv
         public var conv3x3: Conv
--- a/Models/ImageClassification/DenseNet.swift
+++ b/Models/ImageClassification/DenseNet.swift
@@ -81,7 +81,7 @@ extension DenseNet {
         }
     }
 
-    // A pair of a 1x1 `Conv` layer and a 3x3 `Conv` layer.
+    /// A pair of a 1x1 `Conv` layer and a 3x3 `Conv` layer.
     public struct ConvPair: Layer {
         public var conv1x1: Conv
         public var conv3x3: Conv
@@ -119,7 +119,7 @@ extension DenseNet {
 
         @differentiable
         public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
-            return pairs.differentiableReduce(input) { last, layer in
+            pairs.differentiableReduce(input) { last, layer in
                 layer(last)
             }
         }
--- a/Models/ImageClassification/DenseNet.swift
+++ b/Models/ImageClassification/DenseNet.swift
@@ -31,13 +31,13 @@ public struct DenseNet: Layer {
         strides: (2, 2),
         padding: .same
     )
-    public var denseBlock1 = DenseBlock(repitionCount: 6, inputFilterCount: 64)
+    public var denseBlock1 = DenseBlock(repetitionCount: 6, inputFilterCount: 64)
     public var transitionLayer1 = TransitionLayer(inputFilterCount: 256)
-    public var denseBlock2 = DenseBlock(repitionCount: 12, inputFilterCount: 128)
+    public var denseBlock2 = DenseBlock(repetitionCount: 12, inputFilterCount: 128)
     public var transitionLayer2 = TransitionLayer(inputFilterCount: 512)
-    public var denseBlock3 = DenseBlock(repitionCount: 24, inputFilterCount: 256)
+    public var denseBlock3 = DenseBlock(repetitionCount: 24, inputFilterCount: 256)
     public var transitionLayer3 = TransitionLayer(inputFilterCount: 1024)
-    public var denseBlock4 = DenseBlock(repitionCount: 16, inputFilterCount: 512)
+    public var denseBlock4 = DenseBlock(repetitionCount: 16, inputFilterCount: 512)
     public var globalAvgPool = GlobalAvgPool2D<Float>()
     public var dense: Dense<Float>
 
@@ -58,7 +58,7 @@ public struct DenseNet: Layer {
 
 extension DenseNet {
     public struct Conv: Layer {
-        public var bNorm: BatchNorm<Float>
+        public var batchNorm: BatchNorm<Float>
         public var conv: Conv2D<Float>
 
         public init(
@@ -67,7 +67,7 @@ extension DenseNet {
             inputFilterCount: Int,
             outputFilterCount: Int
         ) {
-            bNorm = BatchNorm(featureCount: inputFilterCount)
+            batchNorm = BatchNorm(featureCount: inputFilterCount)
             conv = Conv2D(
                 filterShape: (filterSize, filterSize, inputFilterCount, outputFilterCount),
                 strides: (stride, stride),
@@ -77,7 +77,7 @@ extension DenseNet {
 
         @differentiable
         public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
-            return conv(relu(bNorm(input)))
+            conv(relu(batchNorm(input)))
         }
     }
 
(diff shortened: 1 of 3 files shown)


Andr0id100:
Is that alright @rxwei ?

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,141 @@
...
+// limitations under the License.
+
+import TensorFlow
+
+// Original Paper:
+// Densely Connected Convolutional Networks
+// Gao Huang, Zhuang Liu, Laurens van der Maaten, Kilian Q. Weinberger
+// https://arxiv.org/pdf/1608.06993.pdf
+
+public struct Conv: Layer {
```
rxwei: This public struct should not be defined at top level because it's very specific to this model. When defined at top level, all users importing `ImageClassification` would get a public `Conv`, which is a leaky abstraction that would cause confusion. This has been sub-optimal in other `ImageClassification` models too, e.g. the public [`BatchNormConv2DBlock `](https://github.com/tensorflow/swift-models/blob/a60989d48bce41066de33dc2af280813b54232e1/Models/ImageClassification/WideResNet.swift#L23) struct.

Let's think about what's the right solution moving forward. I'd suggest doing a major refactoring to put any auxiliary layers in an extension of the primary model type. So this model could look like:

```swift
public struct DenseNet: Layer {
    public var conv = Conv(...)
    public var maxpool = MaxPool2D<Float>(...)
    ...
}

extension DenseNet {
    public struct Conv: Layer {
        ...
    }
}
```

What do you think about this approach? If we reach consensus on this, we can start a PR to refactor all other models this way.

Andr0id100: I think that would be a good idea.

Andr0id100: Does something like this looks good?

Andr0id100:
Does something like this looks good?

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+            )
+        }
+
+        @differentiable
+        public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+            return conv(relu(bNorm(input)))
+        }
+    }
+
+    // Seperate structure for the 1x1 and 3x3 conv layers whose collection
```
rxwei: For a documentation comment, use a formal sentence to describe `ConvPair`. I'd suggest something like the following:

_A pair of a 1x1 `Conv` layer and a 3x3 `Conv` layer._

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+            let conv3Output = conv3x3(conv1Output)
+            return conv3Output.concatenated(with: input, alongAxis: -1)
+        }
+    }
+
+    public struct DenseBlock: Layer {
+        public var pairs: [ConvPair] = []
+
+        public init(repitionCount: Int, growthRate: Int = 32, inputFilterCount: Int) {
+            for i in 0 ..< repitionCount {
```
rxwei:
```suggestion
            for i in 0..<repetitionCount {
```

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+                let filterCount = inputFilterCount + i * growthRate
+                pairs.append(ConvPair(inputFilterCount: filterCount, growthRate: growthRate))
+            }
+        }
+
+        @differentiable
+        public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+            let pairsReduced = pairs.differentiableReduce(input) { last, layer in
```
rxwei: Return this directly.
```suggestion
            pairs.differentiableReduce(input) { last, layer in
```

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+                layer(last)
+            }
+            return pairsReduced
```
rxwei:
```suggestion
```

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+                filterSize: 1,
+                inputFilterCount: inputFilterCount,
+                outputFilterCount: inputFilterCount / 2
+            )
+            pool = AvgPool2D(poolSize: (2, 2), strides: (2, 2), padding: .same)
+        }
+
+        @differentiable
+        public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+            return input.sequenced(through: conv, pool)
```
rxwei:
```suggestion
            input.sequenced(through: conv, pool)
```
We can omit the `return` in functions whose body has a single expression.

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+            conv = Conv2D(
+                filterShape: (filterSize, filterSize, inputFilterCount, outputFilterCount),
+                strides: (stride, stride),
+                padding: .same
+            )
+        }
+
+        @differentiable
+        public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+            return conv(relu(bNorm(input)))
```
rxwei:
```suggestion
            conv(relu(bNorm(input)))
```

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+            let conv1Output = conv1x1(input)
+            let conv3Output = conv3x3(conv1Output)
+            return conv3Output.concatenated(with: input, alongAxis: -1)
+        }
+    }
+
+    public struct DenseBlock: Layer {
+        public var pairs: [ConvPair] = []
+
+        public init(repitionCount: Int, growthRate: Int = 32, inputFilterCount: Int) {
```
rxwei: Typo: "repition" -> "repetition"

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        @differentiable
+        public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+            return conv(relu(bNorm(input)))
+        }
+    }
+
+    /// A pair of a 1x1 `Conv` layer and a 3x3 `Conv` layer.
+    // make a dense block
```
rxwei:
```suggestion
```

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,148 @@
...
+        let level2 = level1.sequenced(through: denseBlock2, transitionLayer2)
+        let level3 = level2.sequenced(through: denseBlock3, transitionLayer3)
+        let output = level3.sequenced(through: denseBlock4, globalAvgPool, dense)
+        return output
+    }
+}
+
+extension DenseNet {
+    public struct Conv: Layer {
+        public var bNorm: BatchNorm<Float>
```
rxwei: Avoid abbreviations, especially single-character ones.
```suggestion
        public var batchNorm: BatchNorm<Float>
```

Andr0id100:
Thanks for all the feedback.

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,146 @@
...
+            )
+        }
+
+        @differentiable
+        public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+            conv(relu(batchNorm(input)))
+        }
+    }
+
+    // A pair of a 1x1 `Conv` layer and a 3x3 `Conv` layer.
```
rxwei: Please use three slashes for documentation comments.

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,146 @@
...
+        public init(repetitionCount: Int, growthRate: Int = 32, inputFilterCount: Int) {
+            for i in 0..<repetitionCount {
+                let filterCount = inputFilterCount + i * growthRate
+                pairs.append(ConvPair(inputFilterCount: filterCount, growthRate: growthRate))
+            }
+        }
+
+        @differentiable
+        public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+            return pairs.differentiableReduce(input) { last, layer in
```
rxwei: Omit the ‘return’ keyword for consistency.

Andr0id100:
I am thinking of giving it a structure similar to the ResNet file to incorporate all the different types.
I could change the name to DenseNet121 for now.

Tests/ImageClassificationTests/Inference.swift
```diff
@@ -21,6 +21,15 @@ final class ImageClassificationInferenceTests: XCTestCase {
     override class func setUp() {
         Context.local.learningPhase = .inference
     }
+    
+    func testDenseNet() {
```
rxwei:
```suggestion
    func testDenseNet121() {
```

Andr0id100:
I think this can be merged now.

PR #218, opened 2019-10-19
use enable in _vjpCausallyMasked

### LINKED ISSUE #217: Transformer model: _vjpCausallyMasked does not take the enable flag into account
The enable flag in function _vjpCausallyMasked is not used when calling causallyMasked.
Since the flag defaults to false on the latter function, masking is never executed.

https://github.com/tensorflow/swift-models/blob/41c4a7a782debc8361d1ba7c2de176c9c028b75e/Transformer/Model.swift#L102-L106

Solution:
```
func _vjpCausallyMasked(_ dotProducts: Tensor<Float>, enable: Bool)
    -> (Tensor<Float>, (Tensor<Float>) -> Tensor<Float>) {
    return (causallyMasked(dotProducts, enable:enable), identity)
}
```

koen-dejonghe:
Fixes #217


BradLarson:
As Richard said, thanks for finding this, that was a really subtle problem. Greatly appreciate the fix.

koen-dejonghe:
@BradLarson My pleasure.

PR #219, opened 2019-11-07
Uses switch instead of ifs

valeriyvan:


Catch/main.swift
```diff
@@ -89,12 +89,11 @@ extension CatchAgent {
     func perfectAction(for observation: Observation) -> Action {
         let paddleX = observation.scalars[0]
         let ballX = observation.scalars[1]
-        if paddleX > ballX {
-            return .right
-        } else if paddleX < ballX {
-            return .left
+        switch paddleX {
+        case ..<ballX: return .left
+        case ballX:    return .none
```
dan-zheng: Minor: how about moving `case ballX: return .none` to be first in the `switch` statement?

dan-zheng: Done in https://github.com/tensorflow/swift-models/pull/219/commits/8adee079df8881551b0d216981eaf0b15f980c55.

PR #222, opened 2019-11-08
Initial training and inference benchmarks

BradLarson:
This creates a new executable target called Benchmarks that can be used as a basis for running training and inference benchmarks. It introduces callback-based benchmarking functions that currently only log values to the console, but which could be extended to report values to other backends.

As an initial test case, benchmarks have been configured for training of LeNet against the MNIST dataset and performing inference using a randomly initialized LeNet model against MNIST-sized random image tensors.

Future plans include the ability to selectively perform specific benchmarks via command-line options.

I'm open to any suggestions about architecture and implementation of these benchmarks.


Benchmarks/Benchmark.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    let (samplesPerSecond, standardDeviation) = statistics(for: timings)
+    let results = InferenceBenchmarkResults(
+        iterations: iterations, samplesPerSecond: samplesPerSecond,
+        standardDeviation: standardDeviation)
+    callback(results)
+}
+
+func logResults(name: String) -> (BenchmarkResults) -> Void {
```
dan-zheng: Rather than adding a callback for printing benchmark name, how about adding a "name" property to `BenchmarkResults`?

This seems pretty standard, see [`BenchmarkInfo.name`](https://github.com/apple/swift/blob/5f21c123583bb06540748002e516875153cf2067/benchmark/utils/TestsUtils.swift#L110) from the Swift benchmark suite.

.gitignore
```diff
@@ -9,3 +9,7 @@
 cifar-10-batches-py/
 cifar-10-batches-bin/
 output/
+t10k-labels-idx1-ubyte
+t10k-images-idx3-ubyte
+train-labels-idx1-ubyte
+train-images-idx3-ubyte
```
saeta: nit: trailing newline.

Benchmarks/Benchmark.swift
```diff
@@ -0,0 +1,114 @@
...
+    var description: String { get }
+}
+
+struct TrainingBenchmarkResults: BenchmarkResults {
+    let iterations: Int
+    let averageTime: Double
+    let standardDeviation: Double
+
+    var description: String {
+        get {
```
saeta: I think you can drop the extra `get {}`, here and below.

Benchmarks/Benchmark.swift
```diff
@@ -0,0 +1,114 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+protocol BenchmarkResults: CustomStringConvertible {
+    var description: String { get }
+}
+
+struct TrainingBenchmarkResults: BenchmarkResults {
```
saeta: Can you merge this and the inference result into a single struct with an enum of training vs inference? Is there a reason why they're separate?

Benchmarks/Benchmark.swift
```diff
@@ -0,0 +1,114 @@
...
+    let iterations: Int
+    let averageTime: Double
```
saeta: What is average time? Is this "step-time" for a single training step? If so, this probably can be merged with the examples per second used in inference.

Out of curiosity, what about having the struct be something to the effect of:

```swift
struct BenchmarkResult {
  var name: String  // Maybe an enum of known models, and name becomes computed?
  var kind: BenchmarkKindEnum
  var sampleCount: Int
  var batchSize: Int  // aka sampleCountPerBatch
  var runTimes: [Int]
  var examplesPerBatch: Int { sampleCount / batchSize }  // Verify this divides evenly!
  var runTimesStdDev: Double { ... }
  var iterations: Int { runTimes.count }
  var meanSamplesPerSecond: Double { runTimes.map { ... }.reduce(0, +) / iterations }
  var description: String { ... }
}
```

Benchmarks/Benchmark.swift
```diff
@@ -0,0 +1,114 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+protocol BenchmarkResults: CustomStringConvertible {
+    var description: String { get }
```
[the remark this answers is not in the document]
BradLarson: Thanks for the pointer. Done.

Benchmarks/Benchmark.swift
```diff
@@ -0,0 +1,114 @@
...
+        get {
+            return """
+                  \tAfter \(iterations) iterations:
+                  \tSamples per second: \(samplesPerSecond), standard deviation: \(standardDeviation)
+                """
+        }
+    }
+}
+
+func timeExecution(_ operation: () -> Void) -> Double {
```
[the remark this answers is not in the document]
BradLarson: I got so used to Dispatch not being available on Linux in the early days that I keep forgetting we have this available to us now. I've reworked the timing per your suggestion.

Benchmarks/Benchmark.swift
BradLarson: I restructured the whole BenchmarkResult type to have shared common values and an enum for the kind of benchmark, so I was able to move this into a shared value and simplify the logging function like you describe.

Benchmarks/Benchmark.swift
BradLarson: Thanks for pointing that out. I did this as part of a larger rework on the BenchmarkResult type.

Benchmarks/Benchmark.swift
BradLarson: That was a much better way of handling the results, so I've restructured them in the manner you describe. It simplified much of the code here.

BradLarson:
In general, I reworked the core benchmark functions into using a common BenchmarkResults struct that has an enum with associated values for the kind of benchmark result we're after. The description switches on this, and I can see having JSON, etc. that we'd need for other callbacks be selectively generated from that.

`benchmark()` itself has been combined into a single function that varies its behavior based on an enum for the kind of benchmark performed. It uses a common notion of iterations and some timed operation within each iteration, then specializes that based on the variety.

I'd like to benchmark the initialization of the model, because that can be a pretty slow operation at present, depending on how weights are initialized. That's a definite target for optimization and should be benchmarked.

I can take a stab at pulling out boilerplate from the model training code to make that more easily extensible to other models.

.gitignore
BradLarson: Done.

Benchmarks/Benchmark.swift
```diff
@@ -0,0 +1,104 @@
...
+
+struct BenchmarkResults {
+    let name: String
+    let iterations: Int
+    let timings: [Double]
+    let variety: BenchmarkVariety
+}
+
+extension BenchmarkResults {
+    func interpretTimings() -> [Double] {
```
dan-zheng:
```suggestion
    var interpretedTimings: [Double] {
```

PR #226, opened 2019-11-20
Add initial support for command-line argument parsing

shabalind:
This PR adds support for command-line argument parsing to the Benchmarks. The implementation is build upon Commander library and introduces two commands:

* `measure`. Starts a benchmarking of a single benchmark with given settings.
* `measure-all`. Run all benchmarks with default settings.
* `list-defaults`. See the list of all benchmarks and their default settings.

### DIFF
--- a/Benchmarks/Benchmark.swift
+++ b/Benchmarks/Benchmark.swift
@@ -14,27 +14,13 @@
 
 import Foundation
 
+protocol Benchmark {
+    func run()
+}
+
 enum BenchmarkVariety {
-    case inferenceThroughput(batches: Int, batchSize: Int)
+    case inferenceThroughput
     case trainingTime
-}
-
-struct BenchmarkResults {
-    let name: String
-    let iterations: Int
-    let timings: [Double]
-    let variety: BenchmarkVariety
-}
-
-extension BenchmarkResults {
-    var interpretedTimings: [Double] {
-        switch self.variety {
-        case let .inferenceThroughput(batches, batchSize):
-            return timings.map { Double(batches * batchSize) / ($0 / 1000.0) }
-        case .trainingTime:
-            return timings
-        }
-    }
 }
 
 struct BenchmarkSettings {
@@ -44,23 +30,42 @@ struct BenchmarkSettings {
     let epochs: Int
 }
 
+struct BenchmarkResults {
+    let name: String
+    let timings: [Double]
+    let settings: BenchmarkSettings
+    let variety: BenchmarkVariety
+}
+
+extension BenchmarkResults {
+    var interpretedTimings: [Double] {
+        switch self.variety {
+        case .inferenceThroughput:
+            let batches = settings.batches
+            let batchSize = settings.batchSize
+            return timings.map { Double(batches * batchSize) / ($0 / 1000.0) }
+        case .trainingTime:
+            return timings
+        }
+    }
+}
+
 /// Performs the specified benchmark over a certain number of iterations and provides the result to a callback function.
 func benchmark(
     name: String,
-    iterations: Int, variety: BenchmarkVariety,
-    operation: () -> Void,
+    settings: BenchmarkSettings,
+    variety: BenchmarkVariety,
+    benchmark: Benchmark,
     callback: (BenchmarkResults) -> Void
 ) {
     var timings: [Double] = []
-    for _ in 0..<iterations {
-        let timing = time(operation)
+    for _ in 0..<settings.iterations {
+        let timing = time(benchmark.run)
         timings.append(timing)
     }
 
     let results = BenchmarkResults(
-        name: name, iterations: iterations,
-        timings: timings, variety: variety
-    )
+        name: name, timings: timings, settings: settings, variety: variety)
     callback(results)
 }
 
@@ -97,14 +102,14 @@ func logResults(_ result: BenchmarkResults) {
         print(
             """
                 Benchmark: \(result.name):
-                \tAfter \(result.iterations) iterations:
+                \tAfter \(result.settings.iterations) iterations:
                 \tSamples per second: \(String(format: "%.2f", average)), standard deviation: \(String(format: "%.2f", standardDeviation))
             """)
     case .trainingTime:
         print(
             """
                 Benchmark: \(result.name):
-                \tAfter \(result.iterations) iterations:
+                \tAfter \(result.settings.iterations) iterations:
                 \tAverage: \(String(format: "%.2f", average)) ms, standard deviation: \(String(format: "%.2f", standardDeviation)) ms
             """)
     }
--- a/Benchmarks/Benchmark.swift
+++ b/Benchmarks/Benchmark.swift
@@ -37,6 +37,13 @@ extension BenchmarkResults {
     }
 }
 
+struct BenchmarkSettings {
+    let batches: Int
+    let batchSize: Int
+    let iterations: Int
+    let epochs: Int
+}
+
 /// Performs the specified benchmark over a certain number of iterations and provides the result to a callback function.
 func benchmark(
     name: String,
--- a/Benchmarks/Benchmark.swift
+++ b/Benchmarks/Benchmark.swift
@@ -18,6 +18,13 @@ protocol Benchmark {
     func run()
 }
 
+protocol BenchmarkModel {
+    func inferenceDefaults() -> BenchmarkSettings
+    func inferenceBenchmark(_ settings: BenchmarkSettings) -> Benchmark
+    func trainingDefaults() -> BenchmarkSettings
+    func trainingBenchmark(_ settings: BenchmarkSettings) -> Benchmark
+}
+
 enum BenchmarkVariety {
     case inferenceThroughput
     case trainingTime
@@ -28,6 +35,16 @@ struct BenchmarkSettings {
     let batchSize: Int
     let iterations: Int
     let epochs: Int
+
+    func withDefaults(_ defaults: BenchmarkSettings) -> BenchmarkSettings {
+        let newBatches = batches == -1 ? defaults.batches : batches
+        let newBatchSize = batchSize == -1 ? defaults.batchSize : batchSize
+        let newIterations = iterations == -1 ? defaults.iterations : iterations
+        let newEpochs = epochs == -1 ? defaults.epochs : epochs
+        return BenchmarkSettings(
+            batches: newBatches, batchSize: newBatchSize,
+            iterations: newIterations, epochs: newEpochs)
+    }
 }
 
 struct BenchmarkResults {
--- a/Benchmarks/Models/ImageClassificationInference.swift
+++ b/Benchmarks/Models/ImageClassificationInference.swift
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import TensorFlow
 import Datasets
 import ImageClassificationModels
+import TensorFlow
 
 protocol ImageClassificationModel: Layer where Input == Tensor<Float>, Output == Tensor<Float> {
     init()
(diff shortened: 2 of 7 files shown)


Benchmarks/main.swift
```diff
@@ -12,20 +12,88 @@
...
+let main =
+    Group{group ingroup.command(
+        "measure",
+        Flag("training"),
+        Flag("inference"),
+        Option("benchmark", default: ""),
+        Option("batches", default: 1000),
+        Option("batchSize", default: 1),
+        Option("iterations", default: 10),
+        Option("epochs", default: 1)
```
shabalind: @BradLarson On one hand, it would be nice to have sane defaults, but on the other the defaults can vary per model or training/inference benchmark varieties. Do you think it's better to just force people to have explicitly specify batches/batchSize/epochs on every invocation?

Benchmarks/Models/ImageClassificationTraining.swift
```diff
@@ -12,26 +12,28 @@
...
     // interface to allow for them to be interchangeable in these benchmark cases.
     let dataset: ClassificationDataset
+
     let epochs: Int
     let batchSize: Int
 
-    init(epochs: Int, batchSize: Int) {
-        self.epochs = epochs
-        self.batchSize = batchSize
+    init(withSettings settings: BenchmarkSettings) {
```
saeta: Ditto.

Benchmarks/main.swift
```diff
@@ -12,20 +12,91 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import ImageClassificationModels
+import Commander
 import Datasets
+import ImageClassificationModels
+
+let trainingBenchmarks = [
```
saeta: Is there a reason why we have made the distinction (at this level) between training & inference benchmarks? (Concretely, I suspect that pretty much every (model x dataset) that can be used for training can also be used for inference as well.) Note: this can be refactored in a future PR; don't let me slow you down!

Benchmarks/Models/ImageClassificationInference.swift
```diff
@@ -12,29 +12,30 @@
...
+
     var model: Model
     let images: Tensor<Float>
     let batches: Int
     let batchSize: Int
 
-    init(batches: Int, batchSize: Int, images: Tensor<Float>? = nil) {
-        self.batches = batches
-        self.batchSize = batchSize
+    init(withSettings settings: BenchmarkSettings, andImages images: Tensor<Float>? = nil) {
```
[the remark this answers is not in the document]
shabalind: Thanks for the suggestion, just fixed it!

Benchmarks/Models/ImageClassificationTraining.swift
shabalind: Fixed.

Benchmarks/main.swift
shabalind: It's a good point! Changes in this PR are following the previous distinction with separate benchmark instances for training vs inference. We should probably consider refactoring it further in a follow-up PR.

Benchmarks/main.swift
shabalind: Another alternative would be to make benchmarks declare their defaults on per-benchmark basis (and have any unset cli arguments default to those values).

Benchmarks/main.swift
BradLarson: Originally, I had the code together for running inference and training benchmarks in a single class, but that wasn't the cleanest. It helped to have at least the running code split into two separate types.

However, the point is valid that a combination of dataset and model should be able to run in inference as well as training. Maybe an ImageClassificationBenchmark class with lazy variables for both inference and training benchmarks (to avoid initialization overhead for unused benchmarks) could be created to encapsulate inference and training, and the appropriate benchmark used as needed?

Benchmarks/main.swift
```diff
@@ -12,20 +12,90 @@
...
+    "lenet-mnist": { ImageClassificationTraining<LeNet, MNIST>(settings: $0) },
+]
+
+let inferenceBenchmarks = [
+    "lenet-mnist": { ImageClassificationInference<LeNet, MNIST>(settings: $0) },
+]
```
BradLarson: Should we think about moving the list of specific benchmarks, with their lookup identifiers, into a file other than main.swift, and leave main.swift as just the command-line parsing code? That might make it easier to people to focus on just the code that needs to be modified to add other benchmarks to the list.

Benchmarks/Benchmark.swift
```diff
@@ -14,22 +14,35 @@
...
+    case inferenceThroughput
     case trainingTime
 }
 
+struct BenchmarkSettings {
+    let batches: Int
+    let batchSize: Int
+    let iterations: Int
+    let epochs: Int
+}
```
BradLarson: It bothers me a bit that not all of these settings are applicable to the specific classes of inference or training benchmarks. Normally, for situations like that I'd think about an enum with associated values to prevent you from specifying a value that didn't apply for a certain case. However, people can specify settings for all of these at the command line, and is there any value in warning people about that? I don't know.

Benchmarks/README.md
```diff
@@ -18,5 +18,8 @@ installed. Make sure you've added the correct version of `swift` to your path.
 To run all benchmarks, type the following while in the swift-models directory:
 
 ```sh
-swift run -c release Benchmarks
-```
\ No newline at end of file
+swift run -c release Benchmarks measure --benchmark lanet-mnist --training
+--epochs 1 --batchSize 128
+swift run -c release Benchmarks measure --benchmark lanet-mnist --training
```
BradLarson: Should this be `--inference`?

Benchmarks/main.swift
```diff
@@ -12,20 +12,90 @@
...
+        ) { (trainingFlag, inferenceFlag, name, batches, batchSize, iterations, epochs) in
+            let settings = BenchmarkSettings(
+                batches: batches,
+                batchSize: batchSize,
+                iterations: iterations,
+                epochs: epochs)
+            if !trainingFlag && !inferenceFlag {
+                print("Must specify either --training xor --inference benchmark variety.")
+            } else if trainingFlag && inferenceFlag {
+                print("Can't specify both --training and --inference benchmark variety.")
```
BradLarson: Do we want to have an option for `--all`, that runs all benchmarks for inference and training with their own model-specific default parameters? Or being able to run both inference and training benchmarks for a single model in one command?

That might require needing to supply separate parameters for the inference and training benchmark runs, which makes things a little more complicated.

Benchmarks/main.swift
BradLarson: I like the idea of per-benchmark defaults, specified in some fashion. Our benchmark specification is fairly simple (and possibly simpler, if we combine defining inference and training benchmarks into one class containing both), so adding default benchmark settings to that wouldn't make these unwieldy. Having to remember all the individual parameters at the command line to create comparable benchmarks could be made easier with some established defaults.

These defaults would vary from model to model, because a batch size of 128 might be great for a small model but exhaust GPU memory on a large ResNet. These parameters might also differ between inference and training.

Which reminds me, we should probably log out the benchmark settings used in our output from the benchmarks, to mark the conditions they were run under.

shabalind: Just added support for per-benchmark defaults in 585268df.

Benchmarks/main.swift
shabalind: In 585268df, I added BenchmarkModel to refactor this better.

Benchmarks/Benchmark.swift
shabalind: I'm afraid if we don't keep it together as a single settings struct, it will make the CLI too complicated for no good reason. If this turn out to be a problem we can always revisit this design later on.

Benchmarks/README.md
shabalind: Yep, good catch!

Benchmarks/main.swift
shabalind: Added a separate command called `measure-all` that runs all benchmarks in both training and inference modes with corresponding default settings. Unlike `measure` it does not support overriding settings. It encourages us to actually have good defaults. If a user wants to run with custom settings they should use `measure` on per-benchmark basis instead.

Benchmarks/main.swift
shabalind: Moved the list to `Models.swift`.

Benchmarks/Benchmark.swift
shabalind: I made `list-defaults` command that shows expected fully qualified invocations. I hope this makes it more clear where to use what settings (and what are the defaults).

Benchmarks/README.md
shabalind: To avoid errors like that, `list-defaults` command now shows all default flags for all benchmarks.

Benchmarks/Benchmark.swift
```diff
@@ -14,22 +14,52 @@
 
 import Foundation
 
+protocol Benchmark {
+    func run()
+}
+
+protocol BenchmarkModel {
+    func inferenceDefaults() -> BenchmarkSettings
```
dan-zheng: Naming suggestions:
* Value-producing argument-less methods (e.g. those producing default values) are naturally expressed as computed properties: `func inferenceDefaults() -> ...` -> `var defaultInferenceSettings: ... { get }`.
* Rename `func inferenceBenchmark` to `func benchmarkInference` so that the verb appears first.
```swift
protocol BenchmarkModel {
    var defaultInferenceSettings: BenchmarkSettings { get }
    func benchmarkInference(_ settings: BenchmarkSettings) -> Benchmark
    var defaultTrainingSettings: BenchmarkSettings { get }
    func benchmarkTraining(_ settings: BenchmarkSettings) -> Benchmark
}
```

Benchmarks/main.swift
```diff
@@ -12,20 +12,121 @@
...
 
-// LeNet-MNIST
-let leNetTrainingBenchmark = ImageClassificationTraining<LeNet, MNIST>(epochs: 1, batchSize: 128)
-benchmark(
-    name: "LeNet-MNIST (training)",
-    iterations: 10, variety: .trainingTime, operation: leNetTrainingBenchmark.train,
-    callback: logResults)
+func runBenchmark(
+    _ name: String,
+    withSettings settings: BenchmarkSettings,
```
dan-zheng: Naming suggestion: prepositions like `with` and `and` should be omitted unless they have critical grammatical/semantic significance. This makes call-sites read more nicely.
```swift
func runBenchmark(
    _ name: String,
    settings: BenchmarkSettings,
    variety: BenchmarkVariety
) {
```

shabalind: Thanks for the suggestion! I'll address it in one of the follow-up PRs.

Benchmarks/Benchmark.swift
shabalind: 1. Thanks! Will address in the follow-up PR.
2. Here `inferenceBenchmark` *produces* a benchmark instance, it doesn't actually run the benchmark. So `benchmark` is actually a noun.

Benchmarks/main.swift
shabalind: Thanks for the suggestion! I'll address it in one of the follow-up PRs.

Benchmarks/Benchmark.swift
dan-zheng: > 2\. Here `inferenceBenchmark` _produces_ a benchmark instance, it doesn't actually run the benchmark. So `benchmark` is actually a noun.

Thanks for pointing that out, my mistake!

I think `func makeInferenceBenchmark(settings: BenchmarkSettings) -> Benchmark` is an appropriate name; similar for `func makeTrainingBenchmark`. Factory methods in Swift begin with "make", e.g. [`Collection.makeIterator`](https://developer.apple.com/documentation/swift/collection/2883861-makeiterator).

PR #228, opened 2019-11-22
Use stderr for logging messages

shabalind:
Currently, we show log-style messages in stdout together with actual tool output (e.g., models print messages to show progress on loading the models). This makes it difficult to programmatically involve benchmarks cli from other tools. 

In this PR, we use stderr for logging-style messages instead. This is fairly standard convention in *nix world.

### DIFF
--- a/Support/Stderr.swift
+++ b/Support/Stderr.swift
@@ -9,6 +9,6 @@ extension FileHandle: TextOutputStream {
     }
 }
 
-public func printerr(_ msg: String) {
-    print(msg, to: &stderr)
+public func printError(_ message: String) {
+    print(message, to: &stderr)
 }
--- a/Support/Stderr.swift
+++ b/Support/Stderr.swift
@@ -1,0 +1,14 @@
+import Foundation
+
+var stderr = FileHandle.standardError
+
+extension FileHandle: TextOutputStream {
+    public func write(_ string: String) {
+        guard let data = string.data(using: .utf8) else { return }
+        self.write(data)
+    }
+}
+
+public func printerr(_ msg: String) {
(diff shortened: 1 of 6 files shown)


Support/Stderr.swift
```diff
@@ -0,0 +1,14 @@
...
+var stderr = FileHandle.standardError
+
+extension FileHandle: TextOutputStream {
+    public func write(_ string: String) {
+        guard let data = string.data(using: .utf8) else { return }
+        self.write(data)
+    }
+}
+
+public func printerr(_ msg: String) {
```
dan-zheng: Can this function be internal? Naming suggestions:
```suggestion
func printError(_ message: String) {
```

dan-zheng: Just for reference, here are SwiftPM's own more involved printing facilities: [`InteractiveWriter`](https://github.com/apple/swift-package-manager/blob/d923aa4e28d39c0fe48002e26fe5cb96e1ea8380/Sources/Commands/Error.swift#L103), [`print(error:)`](https://github.com/apple/swift-package-manager/blob/d923aa4e28d39c0fe48002e26fe5cb96e1ea8380/Sources/Commands/Error.swift#L66).

I prefer the name `printError(_:)` over `print(error:)` following the [Swift naming guidelines](https://swift.org/documentation/api-design-guidelines/#naming):

> Otherwise, if the first argument forms part of a grammatical phrase, omit its label, appending any preceding words to the base name, e.g. `x.addSubview(y)`

shabalind: Thanks for the suggestion! I changed the name, but visibility has to public because it's used across module boundary.

PR #234, opened 2019-12-02
work around duplicate definition of diff witness

marcrasi:
Works around https://bugs.swift.org/browse/TF-1025 by specifying explicit activations on all the layers with default activations.


saeta:
How much work would be required to actually fix the underlying issue? Alternately, can you file a follow-up task for TF-1025 to revert this change?

marcrasi:
I will look at TF-1025 soon because this will confuse anyone who runs into it. There is some related cleanup that I would like to finish first that should make the underlying issue much easier to figure out and fix, so it might be a day before I actually create a fix.

I have added a note in TF-1025 to revert this.

Issue #239, opened 2019-12-04
MobileNet v1 example


texasmichelle:
Add an example that uses the MobileNet v1 model and the ImageNette dataset.

Blocked by #238, #237.

BradLarson:
Completed via PR #240, with training being verified in PR #241. For posterity, these are the results I see (against the test set) in the first ten epochs of training with Imagenette:

> [Epoch 1] Accuracy: 215/500 (0.43) Loss: 1.7700883
> [Epoch 2] Accuracy: 282/500 (0.564) Loss: 1.4064155
> [Epoch 3] Accuracy: 250/500 (0.5) Loss: 1.664016
> [Epoch 4] Accuracy: 304/500 (0.608) Loss: 1.2813418
> [Epoch 5] Accuracy: 352/500 (0.704) Loss: 0.9346338
> [Epoch 6] Accuracy: 373/500 (0.746) Loss: 0.80547553
> [Epoch 7] Accuracy: 367/500 (0.734) Loss: 0.8665428
> [Epoch 8] Accuracy: 386/500 (0.772) Loss: 0.71155274
> [Epoch 9] Accuracy: 348/500 (0.696) Loss: 1.0265149
> [Epoch 10] Accuracy: 405/500 (0.81) Loss: 0.6340269

Issue #237, opened 2019-12-04
Add ImageNette dataset


texasmichelle:
ImageNet would be useful to include in this repo, but is very large and difficult to handle both in terms of downloading and processing. Instead, it would be great to have an adapted version of [ImageNette](https://github.com/fastai/imagenette), a much smaller dataset that roughly mirrors the accuracy of ImageNet.

/cc @BradLarson

BradLarson:
As a public note, before anyone else might start working on this, I should have a pull request in within the next day or so to add this dataset. I've been working with [Ayush Agrawal](https://github.com/Ayush517) to add this to swift-models.

BradLarson:
Completed via PR #241, although some more work is needed to reduce peak memory usage.

PR #240, opened 2019-12-04
Add MobileNet v1

### LINKED ISSUE #238: MobileNet v1 model
It would be great to have a version of MobileNet based on the [keras implementation](https://github.com/keras-team/keras-applications/blob/master/keras_applications/mobilenet.py).

Paper: [MobileNets: Efficient Convolutional Neural Networks for Mobile Vision Applications](https://arxiv.org/abs/1704.04861)

texasmichelle:
Fixes #238.

Add MobileNet v1 model based on the [keras implementation](https://github.com/keras-team/keras-applications/blob/master/keras_applications/mobilenet.py).

- [X] MobileNet v1 model
- [X] MobileNet v1 test
- [x] Conditional zero padding
- [x] Dropout probability
- [x] Depth multiplier with minimum depth
- [x] Verify layer architecture with python keras implementation for additional input sizes
- [x] Verify tensor size fed into the GlobalAvgPool2D layer

Follow-on PRs will address creating a dataset (#237) and example that uses this model (#239).

### DIFF
--- a/Models/ImageClassification/MobileNet.swift
+++ b/Models/ImageClassification/MobileNet.swift
@@ -22,190 +22,199 @@ import TensorFlow
 // https://arxiv.org/abs/1704.04861
 
 public struct ConvBlock: Layer {
-  public var zeroPad = ZeroPadding2D<Float>(padding: ((0, 1), (0, 1)))
-  public var conv: Conv2D<Float>
-  public var batchNorm: BatchNorm<Float>
+    public var zeroPad = ZeroPadding2D<Float>(padding: ((0, 1), (0, 1)))
+    public var conv: Conv2D<Float>
+    public var batchNorm: BatchNorm<Float>
 
-  public init(filterCount: Int, strides: (Int, Int)) {
-    conv = Conv2D<Float>(
-      filterShape: (3, 3, 3, filterCount),
-      strides: strides,
-      padding: .valid)
-    batchNorm = BatchNorm<Float>(featureCount: filterCount)
-  }
+    public init(filterCount: Int, strides: (Int, Int)) {
+        conv = Conv2D<Float>(
+            filterShape: (3, 3, 3, filterCount),
+            strides: strides,
+            padding: .valid)
+        batchNorm = BatchNorm<Float>(featureCount: filterCount)
+    }
 
-  @differentiable
-  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
-    let convolved = input.sequenced(through: zeroPad, conv, batchNorm)
-    return relu6(convolved)
-  }
+    @differentiable
+    public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+        let convolved = input.sequenced(through: zeroPad, conv, batchNorm)
+        return relu6(convolved)
+    }
 }
 
 public struct DepthwiseConvBlock: Layer {
-  @noDerivative
-  let depthMultiplier: Int
-  @noDerivative
-  let strides: (Int, Int)
+    @noDerivative
+    let depthMultiplier: Int
 
-  @noDerivative
-  public let zeroPad = ZeroPadding2D<Float>(padding: ((0, 1), (0, 1)))
-  public var dConv: DepthwiseConv2D<Float>
-  public var batchNorm1: BatchNorm<Float>
-  public var conv: Conv2D<Float>
-  public var batchNorm2: BatchNorm<Float>
+    @noDerivative
+    let strides: (Int, Int)
 
-  public init(filterCount: Int, pointwiseFilterCount: Int, depthMultiplier: Int,
-    strides: (Int, Int)) {
-    if depthMultiplier < 1 {
-      fatalError("Depth multiplier must be an integer greater than 0.")
+    @noDerivative
+    public let zeroPad = ZeroPadding2D<Float>(padding: ((0, 1), (0, 1)))
+    public var dConv: DepthwiseConv2D<Float>
+    public var batchNorm1: BatchNorm<Float>
+    public var conv: Conv2D<Float>
+    public var batchNorm2: BatchNorm<Float>
+
+    public init(
+        filterCount: Int, pointwiseFilterCount: Int, depthMultiplier: Int,
+        strides: (Int, Int)
+    ) {
+        if depthMultiplier < 1 {
+            fatalError("Depth multiplier must be an integer greater than 0.")
+        }
+        self.depthMultiplier = depthMultiplier
+        self.strides = strides
+
+        dConv = DepthwiseConv2D<Float>(
+            filterShape: (3, 3, filterCount, self.depthMultiplier),
+            strides: strides,
+            padding: strides == (1, 1) ? .same : .valid)
+        batchNorm1 = BatchNorm<Float>(
+            featureCount: filterCount * self.depthMultiplier)
+        conv = Conv2D<Float>(
+            filterShape: (
+                1, 1, filterCount * self.depthMultiplier,
+                pointwiseFilterCount
+            ),
+            strides: (1, 1),
+            padding: .same)
+        batchNorm2 = BatchNorm<Float>(featureCount: pointwiseFilterCount)
     }
-    self.depthMultiplier = depthMultiplier
-    self.strides = strides
 
-    dConv = DepthwiseConv2D<Float>(
-      filterShape: (3, 3, filterCount, self.depthMultiplier),
-      strides: strides,
-      padding: strides == (1, 1) ? .same : .valid)
-    batchNorm1 = BatchNorm<Float>(
-      featureCount: filterCount * self.depthMultiplier)
-    conv = Conv2D<Float>(
-      filterShape: (1, 1, filterCount * self.depthMultiplier,
-        pointwiseFilterCount),
-      strides: (1, 1),
-      padding: .same)
-    batchNorm2 = BatchNorm<Float>(featureCount: pointwiseFilterCount)
-  }
-
-  @differentiable
-  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
-    var convolved1: Tensor<Float>
-    if self.strides == (1, 1) {
-      convolved1 = input.sequenced(through: dConv, batchNorm1)
+    @differentiable
+    public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+        var convolved1: Tensor<Float>
+        if self.strides == (1, 1) {
+            convolved1 = input.sequenced(through: dConv, batchNorm1)
+        } else {
+            convolved1 = input.sequenced(through: zeroPad, dConv, batchNorm1)
+        }
+        let convolved2 = relu6(convolved1)
+        let convolved3 = relu6(convolved2.sequenced(through: conv, batchNorm2))
+        return convolved3
     }
-    else {
-      convolved1 = input.sequenced(through: zeroPad, dConv, batchNorm1)
-    }
-    let convolved2 = relu6(convolved1)
-    let convolved3 = relu6(convolved2.sequenced(through: conv, batchNorm2))
-    return convolved3
-  }
 }
 
 public struct MobileNetV1: Layer {
-  @noDerivative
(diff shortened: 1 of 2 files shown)


texasmichelle:
Looking for some initial feedback, esp. around:
1. Which additional function arguments would be useful?
2. What other functionality is missing?

Style & polish suggestions encouraged!

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,117 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import TensorFlow
+
+// Original Paper:
+// MobileNets: Efficient Convolutional Neural Networks for Mobile Vision
```
BradLarson: Minor style suggestion: I'd recommend placing the paper title in quotes, which might help make it easier to parse with the text wrapping in the comments.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @differentiable
+  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+    // TODO(michellecasbon): Add conditional zero padding
+    let convolved1 = relu6(input.sequenced(through: dConv, batchNorm1))
+    let convolved2 = relu6(convolved1.sequenced(through: conv, batchNorm2))
+    return convolved2
+  }
+
```
BradLarson: Minor formatting issue: this extra empty line can probably be removed.

Models/ImageClassification/MobileNet.swift
texasmichelle: lol - I just removed the quotes because I was looking at SqueezeNet and trying to conform to the repo standard.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,116 @@
...
+  public var dConvBlock7 = DepthwiseConvBlock(filterCount: 512, pointwiseFilterCount: 512, strides: (1, 1))
+  public var dConvBlock8 = DepthwiseConvBlock(filterCount: 512, pointwiseFilterCount: 512, strides: (1, 1))
+  public var dConvBlock9 = DepthwiseConvBlock(filterCount: 512, pointwiseFilterCount: 512, strides: (1, 1))
+  public var dConvBlock10 = DepthwiseConvBlock(filterCount: 512, pointwiseFilterCount: 512, strides: (1, 1))
+  public var dConvBlock11 = DepthwiseConvBlock(filterCount: 512, pointwiseFilterCount: 512, strides: (1, 1))
+  public var dConvBlock12 = DepthwiseConvBlock(filterCount: 512, pointwiseFilterCount: 1024, strides: (2, 2))
+  public var dConvBlock13 = DepthwiseConvBlock(filterCount: 1024, pointwiseFilterCount: 1024, strides: (1, 1))
+  public var avgPool = GlobalAvgPool2D<Float>()
+  public var reshape = Reshape<Float>(shape: [1, 1, 1, 1024])
+  public var dropout = Dropout<Float>(probability: 0.001)
```
BradLarson: In the Keras model, dropout probability is an input parameter to the model. Do we want to do the same here?

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,113 @@
...
+  public var convLast: Conv2D<Float>
+
+  public init(classCount: Int) {
```
BradLarson: In line with the above comment, in the paper and Keras implementations, a depth multiplier (and minimum depth) are specified as inputs to the model and then used to scale the number of filters to shrink or expand the model. Did we want to match that functionality? If so, the various convolution blocks would probably have to be initialized here with the respective calculated filter counts, rather than above.

Models/ImageClassification/MobileNet.swift
texasmichelle: Done.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,150 @@
...
+  public var dConv: DepthwiseConv2D<Float>
+  public var batchNorm1: BatchNorm<Float>
+  public var conv: Conv2D<Float>
+  public var batchNorm2: BatchNorm<Float>
+
+  public init(filterCount: Int, pointwiseFilterCount: Int, depthMultiplier: Int, strides: (Int, Int)) {
+    if depthMultiplier > 0 {
+        self.depthMultiplier = depthMultiplier
+    } else {
+        print("Depth multiplier must be an integer greater than 0. Setting to default value of 1.")
```
texasmichelle: Is there a better way to output this log message?

Models/ImageClassification/MobileNet.swift
texasmichelle: I'm not sure that I applied the multiplier in the correct place. It was a bit hard to tell since in python it's an argument to DepthwiseConv2D, which is missing in the swift layer. Can you confirm?

Models/ImageClassification/MobileNet.swift
texasmichelle: Would printError from ModelSupport be better here? I ran into linking issues running the test command, so prob just didn't import it properly.

texasmichelle:
Is there a preferred linter? This is ready for review.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,283 @@
...
+      print(convolved.shape)
+      print("Relu6")
+    }
+    return relu6(convolved)
+  }
+}
+
+public struct DepthwiseConvBlock: Layer {
+  @noDerivative
+  var depthMultiplier: Int
```
BradLarson: If this isn't a learned or otherwise mutated parameter, this should probably be a `let`. Likewise for `strides` and `printSummary`.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,283 @@
...
+  @noDerivative
+  var strides: (Int, Int)
+  @noDerivative
+  var printSummary: Bool
+
+  public var zeroPad = ZeroPadding2D<Float>(padding: ((0, 1), (0, 1)))
```
BradLarson: Will the zero padding by mutated in any way by training? If not, it can probably be a let.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,283 @@
...
+    }
+    return convolved3
+  }
+}
+
+public struct MobileNetV1: Layer {
+  @noDerivative
+  var classCount: Int
+  @noDerivative
+  var printSummary: Bool
```
BradLarson: Similarly, `classCount` and `printSummary` can probably be lets here.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,283 @@
...
+  public var batchNorm2: BatchNorm<Float>
+
+  public init(filterCount: Int, pointwiseFilterCount: Int, depthMultiplier: Int,
+    strides: (Int, Int), printSummary: Bool = false) {
+    self.strides = strides
+    self.printSummary = printSummary
+    if depthMultiplier > 0 {
+      self.depthMultiplier = depthMultiplier
+    } else {
+      print("""
```
BradLarson: Do we want to make this check a guard or precondition that triggers a fatalError when they specify an illegal value?

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,283 @@
...
+  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+    let convolved = input.sequenced(through: convBlock1, dConvBlock1,
+      dConvBlock2, dConvBlock3, dConvBlock4)
+    let convolved2 = convolved.sequenced(through: dConvBlock5, dConvBlock6,
+      dConvBlock7, dConvBlock8, dConvBlock9)
+    let convolved3 = convolved2.sequenced(through: dConvBlock10, dConvBlock11,
+      dConvBlock12, dConvBlock13)
+
+    // Verify shape since any discrepancies will be masked after pooling
+    let expectedWidth = input.shape[1] / 32
```
BradLarson: We typically haven't done shape checks inside the models themselves, and materializing the shape here may have adverse performance implications. If this is something we need to test against in unit tests, we may want to think about how we could extract that during a test there, rather than during normal execution.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,283 @@
...
+      print("""
+            Invalid shape before GlobalAveragePooling2D: \(convolved2.shape)
+            should be [1, \(expectedWidth), \(expectedHeight), 1024].
+            """)
+    }
+
+    let convolved4 = convolved3.sequenced(through: avgPool, reshape,
+      dropoutLayer, convLast)
+    let output = convolved4.reshaped(to: [1, classCount])
+    if self.printSummary {
```
BradLarson: Likewise, other models haven't employed shape logging as a standard operation. This might be something to verify in a unit test, but we probably don't want to have the logging internal to the model itself.

BradLarson:
When it comes to linting, we use `swift-format` to format all of our files, and there is a `.swift-format` in this project that specifies our few format tweaks.

Models/ImageClassification/MobileNet.swift
BradLarson: Sorry, ignore the above comment. This is needed for proper autodiff.

Models/ImageClassification/MobileNet.swift
texasmichelle: This was useful for verifying replication of the python implementation, but now that it's built that makes sense.

Models/ImageClassification/MobileNet.swift
texasmichelle: Removed from the execution path. I'll think about how to test for this externally.

Models/ImageClassification/MobileNet.swift
texasmichelle: Based on layer outputs and shapes, this seems correct. We can investigate further when Imagenette is complete and we have a working example.

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,220 @@
...
+            through: convBlock1, dConvBlock1,
+            dConvBlock2, dConvBlock3, dConvBlock4)
+        let convolved2 = convolved.sequenced(
+            through: dConvBlock5, dConvBlock6,
+            dConvBlock7, dConvBlock8, dConvBlock9)
+        let convolved3 = convolved2.sequenced(
+            through: dConvBlock10, dConvBlock11,
+            dConvBlock12, dConvBlock13)
+
+        // Verify shape since any discrepancies will be masked after pooling
```
BradLarson: Because the code checking for this has been removed, this comment can probably be removed as well.

texasmichelle: Good catch

Models/ImageClassification/MobileNet.swift
```diff
@@ -0,0 +1,218 @@
...
+    public var dConv: DepthwiseConv2D<Float>
+    public var batchNorm1: BatchNorm<Float>
+    public var conv: Conv2D<Float>
+    public var batchNorm2: BatchNorm<Float>
+
+    public init(
+        filterCount: Int, pointwiseFilterCount: Int, depthMultiplier: Int,
+        strides: (Int, Int)
+    ) {
+        if depthMultiplier < 1 {
```
BradLarson: Rather than an if statement here, I believe a precondition would be preferred for an invalid argument value check like this. For example, we do [a similar check for random seeds](https://github.com/tensorflow/swift-apis/blob/master/Sources/Tensor/Random.swift#L125) in the swift-apis. That could condense this down to a single line.

texasmichelle: Thanks for the example!

Issue #225, opened 2019-11-19
Remove _Raw operators from models


BradLarson:
In an effort to layer APIs and allow for runtimes other than the traditional TensorFlow eager runtime, access to Raw TensorFlow operators have been marked as _Raw. This expresses that they should be considered as internal implementation details and should not be accessed directly.

These operators are used in several places within the swift-models repository, and we should replace them with higher-level implementations of these operations or rethink how they're used. These are the operators that should be replaced:

- [x] **Datasets/CIFAR10/CIFAR10.swift**: `_Raw.concat(concatDim:)`
- [x] **MiniGo/Models/PythonCheckpointReader.swift**, **Transformer/PythonCheckpointReader.swift**: `_Raw.restoreV2(prefix:)`
- [x] **Transformer/Model.swift**: `_Raw.matrixBandPart()`
- [x] **Transformer/Model.swift**: `_Raw.concatV2()`
- [x] **Transformer/main.swift**: `_Raw.multinomial(logits:, numSamples:)`
- [x] **Transformer/Operators.swift**:  `_Raw.batchMatMul()`
- [x] **Support/Image.swift**:`_Raw.readFile(filename:)`, `_Raw.decodeJpeg(contents:, channels:, dctMethod:)`, `_Raw.reverse()`, `_Raw.encodeJpeg()`, `_Raw.writeFile(filename:, contents:)`, `_Raw.resizeBilinear()`
- [x] **FastStyleTransfer/Demo/Helpers.swift**: `_Raw.readFile(filename:)`, `_Raw.decodeJpeg(contents:, channels:, dctMethod:)`, `_Raw.clipByValue(t:, clipValueMin:, clipValueMax:)`, `_Raw.encodeJpeg()`, `_Raw.writeFile(filename:, contents:)`
- [x] **FastStyleTransfer/Utility/ImportableLayer.swift**: `_Raw.restoreV2(prefix:)`
- [x] **FastStyleTransfer/Utility/Resize.swift**: `_Raw.resizeNearestNeighbor(images:, size:)`, `_Raw.resizeNearestNeighborGrad(grads:, size:)`

The trickiest ones to replace will be the image operations, because that had provided cross-platform image loading and saving, and the checkpoint reading, which will require the addition of direct checkpoint parsing.

rickwierenga:
When I tried installing this repo in Google Colab I got the following (most likely related) error:

```swift
%install-location $cwd/swift-install
[...]

PR #242, opened 2019-12-06
Remove _Raw from CIFAR10.swift

xihui-wu:
Removing using _Raw and use Tensor initializer instead to concatenate segments of datasets.
Details about removing _Raw see https://github.com/tensorflow/swift-models/issues/225.

### DIFF
--- a/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
+++ b/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
@@ -9,11 +9,9 @@ final class CIFAR10Tests: XCTestCase {
 
 		var totalCount = 0
 		for example in dataset.trainingDataset {
-		  if totalCount == 0 {
-		    XCTAssertTrue((0..<10).contains(example.label.scalar!))
-		    XCTAssertEqual(example.data.shape, [32, 32, 3])
-		  }
-		  totalCount += 1
+	    	XCTAssertTrue((0..<10).contains(example.label.scalar!))
+	    	XCTAssertEqual(example.data.shape, [32, 32, 3])
+			totalCount += 1
 		}
 		XCTAssertEqual(totalCount, 50000)
 	}
--- a/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
+++ b/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
@@ -1,0 +1,20 @@
+import TensorFlow
+import XCTest
+
+@testable import Datasets
+
+final class CIFAR10Tests: XCTestCase {
+	func testCreateCIFAR10() {
+		let dataset = CIFAR10()
+
+		var totalCount = 0
+		for example in dataset.trainingDataset {
+		  if totalCount == 0 {
+		    XCTAssertTrue((0..<10).contains(example.label.scalar!))
+		    XCTAssertEqual(example.data.shape, [32, 32, 3])
+		  }
+		  totalCount += 1
+		}
+		XCTAssertEqual(totalCount, 50000)
+	}
+}
--- a/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
+++ b/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
@@ -1,18 +1,18 @@
 import TensorFlow
 import XCTest
-
-@testable import Datasets
+import Datasets
 
 final class CIFAR10Tests: XCTestCase {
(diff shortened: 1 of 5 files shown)


Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
```diff
@@ -0,0 +1,20 @@
...
+		for example in dataset.trainingDataset {
+		  if totalCount == 0 {
+		    XCTAssertTrue((0..<10).contains(example.label.scalar!))
+		    XCTAssertEqual(example.data.shape, [32, 32, 3])
+		  }
+		  totalCount += 1
+		}
+		XCTAssertEqual(totalCount, 50000)
+	}
+}
```
saeta: Nit: could you add an extra newline at the end?

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
```diff
@@ -0,0 +1,20 @@
...
+
+@testable import Datasets
+
+final class CIFAR10Tests: XCTestCase {
+	func testCreateCIFAR10() {
+		let dataset = CIFAR10()
+
+		var totalCount = 0
+		for example in dataset.trainingDataset {
+		  if totalCount == 0 {
```
saeta: Would it make more sense to have the assertions in the body check every example in the dataset, instead of just the first one?

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
```diff
@@ -0,0 +1,20 @@
+import TensorFlow
+import XCTest
+
+@testable import Datasets
```
marcrasi: You can remove `@testable` because that's only necessary for calling private functions, and your test only calls public ones.

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
```diff
@@ -0,0 +1,20 @@
+import TensorFlow
```
marcrasi: Indentation should be 4 spaces. There are some tabs and some 2-space indents in this file currently.

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
xihui-wu: Got it.

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
xihui-wu: Yes! It is actually not as costly as I thought.

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
xihui-wu: Got it.

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
xihui-wu: Corrected.

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
```diff
@@ -0,0 +1,18 @@
...
+import XCTest
+import Datasets
+
+final class CIFAR10Tests: XCTestCase {
```
Shashi456: @marcrasi @saeta Just a question, wouldn't the DatasetTests need an `XCTestManifest.swift`? and also adding that to `LinuxMain.swift`?

marcrasi: Yes, that might be necessary for the tests to run on linux. I'm not familiar enough with XCTest to be able to tell just by reading what this needs.

@xihui-wu: You can run the tests in Linux using kokoro by adding a label `kokoro:run` to this PR. It'll take ~5 minutes and then it'll post a test log on this PR. Look in the log to see if your new test is running on Linux. If it's not running on Linux, look at `ImageClassificationTests` and add similar things to your new test:
* `static var allTests` in the test file
* `XCTestManifest.swift` in the test directory
* Another entry in `LinuxMain.swift`.

xihui-wu: This should be resolved: https://source.cloud.google.com/results/invocations/35894fe5-e12d-4fa8-9335-e5a79baf5e61/log.

PR #250, opened 2019-12-11
add vgg16 network

brettkoonce:


Models/ImageClassification/VGG.swift
```diff
@@ -0,0 +1,65 @@
...
+
+    var flatten = Flatten<Float>()
+    var dense1 = Dense<Float>(inputSize: 512 * 7 * 7, outputSize: 4096, activation: relu)
+    var dense2 = Dense<Float>(inputSize: 4096, outputSize: 4096, activation: relu)
+    var dense3 = Dense<Float>(inputSize: 4096, outputSize: 1000, activation: softmax)
+
+    public init() {}
+
+    @differentiable
+    public func callAsFunction(_ input: Input) -> Output {
```
BradLarson: Stylistically, our current preference is to specify the input and output types in `callAsFunction()`, rather than as typealiases above. I know the compiler suggests the typealiases to conform to the protocol, and we have some examples where we've used the typealiases, but we're moving things toward specifying the input and output types and dropping the typealiases in model definitions.

Therefore, this could be replaced with 

```swift
public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
```

and the above typealiases removed.

brettkoonce: roger, my style is a bit out of date, fixed!

PR #251, opened 2019-12-11
rewrite vgg to use blocks, add vgg 19

brettkoonce:
not sure what all should have the public keyword

### DIFF
--- a/Models/ImageClassification/VGG.swift
+++ b/Models/ImageClassification/VGG.swift
@@ -40,7 +40,7 @@ public struct VGGBlock: Layer {
     }
 }
 
-public struct VGG: Layer {
+public struct VGG16: Layer {
     var layer1: VGGBlock
     var layer2: VGGBlock
     var layer3: VGGBlock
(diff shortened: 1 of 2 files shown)


Shashi456:
I personally think we should have a simple init with just vgg16 and vgg19, although I see the advantages of this approach. Ultimately when we add support for pre trained models, I think it would be neat for each network to have it's own initialization. 
Because calling for `vgg16()` or `vgg19` with a default class count parameter is neater than `vgg(kind:_classcount:_). It might add to the complexity of the api
@bradlarson wdyt?

Models/ImageClassification/VGG.swift
```diff
@@ -19,44 +19,67 @@ import TensorFlow
...
+    }
+}
+
+extension VGG {
+    public enum Kind {
+        case vgg16
+        case vgg19
+    }
+
+    public init(inputKind: Kind, classCount: Int) {
```
BradLarson: I know this is patterned after the ResNet examples, but we really should rename the `inputKind` parameter there and here to something else. `depth` might be a better choice, because that's what these variants really express. I'm not great at naming, but would `ModelDepth` or `ModelSize` be a more descriptive type name than `Kind`? Again, we can update the ResNet example to match what works here.

BradLarson:
The ResNet example: https://github.com/tensorflow/swift-models/blob/master/Models/ImageClassification/ResNet50.swift#L326 switches between depths in this manner, so it's not a terrible approach, especially for models with a lot of depth variants. In others, such as the MobileNet family, a more continuous scaling is applied as a parameter where it might be impractical to have separate instances for each.

We don't know what the story will look like around pretrained models yet, and it would seem like model-specific parameters for loading those could just as easily be attached to an enumeration as a separate initializer for each depth.

Models/ImageClassification/VGG.swift
brettkoonce: I'll agree that I don't like the inputKind parameter here as well.  and yes, the parameter namespace could be unified.

brettkoonce:
@BradLarson let me know what you think about this pattern.

Models/ImageClassification/VGG.swift
brettkoonce: just nuked the logic for now, can be revisited when we get more complicated models?

PR #255, opened 2019-12-12
Also set number of test examples on CIFAR10

rickwierenga:
[reference](https://www.cs.toronto.edu/~kriz/cifar.html)


saeta:
Note: you might want to consider adding it to the protocol: https://github.com/tensorflow/swift-models/blob/master/Datasets/ImageClassificationDataset.swift

rickwierenga:
It should be all in one PR now. (It got a little messy with different patches)

rickwierenga:
I think either `trainingDataset` or `testDataset` needs to be renamed to match the grammatical form. I prefer the present, noncontinuous tense.

PR #256, opened 2019-12-12
speed up ci by using a prebuilt docker image

marcrasi:
Uses the docker image built in https://github.com/tensorflow/swift/pull/342.

This changes CI time from ~6min to ~5m30. Not as much speedup as I was hoping, but the code deduplication is good.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,40 +1,7 @@
-# TODO: We should have a job that creates a S4TF base image so that
-# we don't have to duplicate the installation everywhere.
-FROM nvidia/cuda:9.2-cudnn7-devel-ubuntu18.04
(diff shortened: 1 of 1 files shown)


vvmnnnkv:
@marcrasi have you considered making S4TF toolchain base image? There should be even more deduplication and speed-up, especially if you can use docker build cache for swift-models Dockerfile.
Here's example of toolchain image I made for myself: https://github.com/vvmnnnkv/s4tf-docker to build an app: https://github.com/vvmnnnkv/s4tf-style-transfer-service/blob/master/Dockerfile

marcrasi:
Yes, I want to do that eventually. It's a bit harder because I'd need to reupload the image every time the automated toolchain build makes a new toolchain, and I haven't yet figured out how to authorize our internal toolchain build automation tool to upload images to GCR. I'm sure it's possible but I'll need to spend more time reading the docs and experimenting with it.

PR #261, opened 2019-12-17
Add DCGAN

rickwierenga:
Ayush Agrawal recommended me to add my DCGAN model to this repository. It currently fails to run on `master` due to the rapid changes in the APIs, but it does work on `tensorflow-0.6` as you can see in [this](https://colab.research.google.com/github/rickwierenga/s4tf-notebooks/blob/master/notebooks/MNIST_DCGAN.ipynb) colab. Should I create another PR for `tensorflow-0.6`?


Ayush517:
@BradLarson Like you mentioned, it might make sense to group the non-convolutional GAN currently present with this DCGAN implementation somehow.
Maybe we can merge both models in one script, run them both separately with names '"GAN",  "DCGAN and present graphs and images as outputs. Will this work?

BradLarson:
Thanks for converting your tutorial into a standalone model. Let's try to see if we can get this building against master, which shouldn't be too bad. We haven't deviated very far from the 0.6 release just yet.

As a first few comments, if you can get swift-format up and running on your system, I'd recommend running that against the files here to make sure they fit our formatting guidelines. At a glance, I see several long lines that might need to be reformatted. Additionally, we prefer capitalization and punctuation in comments, and you can look at other models in the repository for examples of how those tend to be formatted.

Beyond that, it would be really nice if we didn't have to depend on Matplotlib for image rendering. Would it be possible to use the project's image-handling functions and types instead, to write out the resulting images to disk? This is done in the existing Autoencoder and GAN examples, for example.

Regarding organization, we might be able to group the Autoencoder, GAN, and DCGAN examples together somehow, possibly placing them under Examples, but that could come after this is made operational in master.

rickwierenga:
So here's the ToDo list:

- [ ] format nicely
  - [ ] comments
- [ ] remove matplotlib (look at current GAN)

Is it ok with you if I do this tomorrow?

---

I think it'd be great to use jupyter notebooks for the examples, because that's what most people will be using for machine learning anyways. We could add instructions for [NotebookExport](https://github.com/latenitesoft/NotebookExport) so that those who want can still run models as a package.

rickwierenga:
I was updating the package to `master`, but got this error:

```
/Users/rickwierenga/workspace/gci/src/swift-models/DCGAN/main.swift:125:48: error: value of type 'Dataset<LabeledExample>' has no member 'minibatch'
        let realImages = mnist.trainingDataset.minibatch(at: i, batchSize: i * batchSize >= mnist.testExampleCount ? (mnist.testExampleCount - ((i - 1) * batchSize)) : batchSize)
                         ~~~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~
```

I'd already fixed a few things at that point. Please let me know if I should share them.

Is there a replacement in the current API? I can add it `ImageClassificationDataset` if you want.

rickwierenga:
All things are done and it runs correctly on my Mac (haven't tested it on Linux).

Let me know if there's anything else you want to see changed.

DCGAN/main.swift
```diff
@@ -0,0 +1,184 @@
...
+// MARK: - Models
+
+// MARK: Generator
+
+struct Generator: Layer {
+    var flatten = Flatten<Float>()
+
+    var dense1 = Dense<Float>(inputSize: zDim, outputSize: 7 * 7 * 256)
+    var batchNorm1 = BatchNorm<Float>(featureCount: 7 * 7 * 256)
+    // leaky relu
```
BradLarson: We don't have placeholder comments like this in other models within the repository, so I'd recommend removing them here and in the descriminator to condense the list of properties for these networks.

DCGAN/main.swift
```diff
@@ -0,0 +1,184 @@
...
+        let x1 = dropout(leakyRelu(conv2D1(input)))
+        let x2 = dropout(leakyRelu(conv2D2(x1)))
+        return x2.sequenced(through: flatten, dense)
+    }
+}
+
+@differentiable
+func discriminatorLoss(realLabels: Tensor<Float>, fakeLabels: Tensor<Float>) -> Tensor<Float> {
+    let realLoss = sigmoidCrossEntropy(logits: realLabels,
+                                       labels: Tensor(ones: realLabels.shape)) // should say it's real, 1
```
BradLarson: The three comments here are probably unnecessary and could be removed. If they were to stay, they'd need to have capitalization and punctuation to match the others.

rickwierenga:
Any update on this? Happy 2020!

BradLarson:
Sorry about the delay, I'd fired off the CI check and forgot to check the results. Thanks for the work in implementing this.

rickwierenga:
Thanks!

rickwierenga:
The images aren't displaying correctly in the README. I made a fix in https://github.com/tensorflow/swift-models/pull/268. Sorry about that.

PR #262, opened 2019-12-17
Use Tensor categorical initializer in Transformer/main.swift

xihui-wu:
This PR switches off _Raw from Transformer/main.swift by using Tensor categorical initializer.

More about removing _Raw: #225

Able to run the model and got reasonable output after the change.


marcrasi:
Kokoro failed because the dev build that it uses doesn't have your latest changes. The next build should pick up your changes, and if you add `kokoro:force-run` after that build is ready, it should pass.

PR #263, opened 2019-12-19
Remove init as a requirement in ImageClassificationDataset

rickwierenga:
Every type that conforms to this protocol will automatically have an initializer requirement set by Swift (`class`/`struct`).

Having it as a requirement prevents users from using the automatically generated struct initializers in Swift.

### DIFF
--- a/Datasets/ImageClassificationDataset.swift
+++ b/Datasets/ImageClassificationDataset.swift
@@ -15,7 +15,6 @@
 import TensorFlow
 
 public protocol ImageClassificationDataset {
-    init()
     var trainingDataset: Dataset<LabeledExample> { get }
     var testDataset: Dataset<LabeledExample> { get }
     var trainingExampleCount: Int { get }


BradLarson:
Unfortunately, this breaks our Benchmarks target (try a `swift run -c release Benchmarks` in your repository to see the results). We currently need a parameterless initializer for datasets to be constructed and used within that target.

I'm open to hearing alternatives to this, but we need a way to initialize generic image classification datasets with default parameters as part of our benchmarks, thus the initializer requirement. For the existing datasets, we use this to specify common defaults that work with our benchmarks.

rickwierenga:
I hadn't thought about that but it makes sense.

I'm not sure how the tests work in detail, but I understand why it would be an issue. I think Swift should "know" that each type has an initializer, but that's not our problem at the moment.

I can just write my own initializer for now. Just thought it was inconvenient to have this "useless" requirement, but apparently it serves a purpose.

PR #264, opened 2019-12-20
Use local archive for CIFAR10Tests

xihui-wu:
This PR majorly fixes CIFAR10Tests internally due to unable to download archive from internet.
We also see GitHub tests failures a few times because the archive file was corrupted due to unstable download.

The fix is to include the archive in the repo under test folder.

I think it's doable for following reasons:
1. the archive is 162M and can be managed with Git LFS
2. CIFAR10.swift assumes the archive file to be in 5 chunks, so we can't say only pick 2 or 3 chunks in test, I don't think we can shrink individual chunk either.
3. The complete dataset can be used for testing training or inferencing in future.

### DIFF
--- a/Datasets/CIFAR10/CIFAR10.swift
+++ b/Datasets/CIFAR10/CIFAR10.swift
@@ -62,7 +62,7 @@ func downloadCIFAR10IfNotPresent(to directory: URL) {
 
     guard !directoryExists else { return }
 
-    printError("Downloading CIFAR dataset...")
+    printError("Preparing CIFAR dataset...")
     let archivePath = directory.appendingPathComponent("cifar-10-binary.tar.gz").path
     let archiveExists = FileManager.default.fileExists(atPath: archivePath)
     if !archiveExists {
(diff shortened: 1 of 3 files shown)


BradLarson:
I have concerns about placing this directly within the GitHub repository itself, because that will mean that everyone who checks out this repository (including any Colab notebooks that use these models and dataset wrappers) will now have to download a ~162 MB binary in addition to the repository every time it is cloned.

Instead, could we rework the CIFAR-10, MNIST, and Imagenette datasets to allow for the user to provide a custom URL (local or remote) for the binaries to be sourced from as an initializer option, with the default falling back to the source URL where we download from now? That way, within our tests we could point to a GCS bucket for the OSS tests and a filesystem location in internal tests (I can set up rewrite rules to have it point somewhere else when the files are copied internally).

xihui-wu:
Started a new PR: https://github.com/tensorflow/swift-models/pull/270/files as per offline discussion. Closing this.

PR #266, opened 2019-12-21
Minor changes to average/standard deviation calculation.

dan-zheng:
- Use `Array.reduce(into:)` for efficiency.
- Add `Array.variance` helper property.
- Minor style edits.
  - Avoid explicit `self.` prefix when unnecessary.

---

[Verified correctness against NumPy](https://gist.github.com/dan-zheng/c46780a1068ad6345523b9f07c2bd776):
```swift
import Python
let np = Python.import("numpy")

for _ in 0..<100 {
  let array = (0..<100).map { _ in Double.random(in: (-1000...1000)) }
  let npArray = np.array(PythonObject(array))
  let npStddev = np.std(npArray, ddof: 1)
  print(array.standardDeviation)
  print(npStddev)
  assert(Bool(np.allclose(array.standardDeviation, npStddev))!)
}
```

### DIFF
--- a/Benchmarks/PrintingStyle.swift
+++ b/Benchmarks/PrintingStyle.swift
@@ -108,23 +108,24 @@ func printJSON<T: Encodable>(_ value: T) {
 
 /// Statistics-related helpers.
 extension Array where Element == Double {
-    /// Average value across all elements of the array.
+    /// The average value of elements.
     var average: Element {
-        guard self.count > 0 else { return 0.0 }
-        guard self.count > 1 else { return self.first! }
-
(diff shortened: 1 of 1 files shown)


BradLarson:
Huh, I was not aware of the efficiency gains of `reduce(into:)` over `reduce()` in this case. Learn something new every day. Thanks for the cleanup.

@saeta - Regarding tests for the benchmarks, @shabalind and I were just discussing that this week in terms of the Python interface code, but tests of the Swift logic too would definitely make sense.

PR #268, opened 2020-01-02
Fix image in DCGAN README

rickwierenga:


BradLarson:
What about adding the images directly into the example, like is done for the Autoencoder sample: https://github.com/tensorflow/swift-models/tree/master/Autoencoder ? That way, we're not dependent on your website to pull these down.

rickwierenga:
Moving this PR here: https://github.com/tensorflow/swift-models/pull/269 (GitHub didn't see the commit somehow)

PR #270, opened 2020-01-02
 Enable user inputting source url while creating CIFAR10

xihui-wu:
This will provide flexibility to download dataset files from S4TF hosted binaries or any where else that the user can specify.

### DIFF
--- a/Datasets/CIFAR10/CIFAR10.swift
+++ b/Datasets/CIFAR10/CIFAR10.swift
@@ -32,21 +32,22 @@ public struct CIFAR10: ImageClassificationDataset {
     public let testExampleCount = 10000
 
     public init() {
-        self.init(
-            remoteBinaryArchiveUrl: "https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz",
-            localStorageDirectory: FileManager.default.temporaryDirectory.appendingPathComponent(
-                "CIFAR10"))
+        self.init(downloadResourceFrom: "https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz")
     }
 
     public init(downloadResourceFrom: String) {
-        self.init(
-            remoteBinaryArchiveUrl: downloadResourceFrom,
-            localStorageDirectory: FileManager.default.temporaryDirectory.appendingPathComponent(
-                "CIFAR10"))
+        if let resourceUrl = URL(string:downloadResourceFrom) {
+            self.init(
+                remoteBinaryArchiveLocation: resourceUrl,
+                localStorageDirectory: FileManager.default.temporaryDirectory.appendingPathComponent("CIFAR10"))
+        } else {
+            printError("Invalid path, please specify another one")
+            exit(-1)
+        }
     }
 
-    public init(remoteBinaryArchiveUrl: String, localStorageDirectory: URL) {
-        downloadCIFAR10IfNotPresent(from: remoteBinaryArchiveUrl, to: localStorageDirectory)
+    public init(remoteBinaryArchiveLocation: URL, localStorageDirectory: URL) {
+        downloadCIFAR10IfNotPresent(from: remoteBinaryArchiveLocation, to: localStorageDirectory)
         self.trainingDataset = Dataset<LabeledExample>(
             elements: loadCIFARTrainingFiles(localStorageDirectory: localStorageDirectory))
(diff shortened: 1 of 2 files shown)


Datasets/CIFAR10/CIFAR10.swift
```diff
@@ -33,19 +33,28 @@ public struct CIFAR10: ImageClassificationDataset {
...
 
-    public init(localStorageDirectory: URL) {
+    public init(downloadResourceFrom: String) {
+        self.init(
+            remoteBinaryArchiveUrl: downloadResourceFrom,
+            localStorageDirectory: FileManager.default.temporaryDirectory.appendingPathComponent(
+                "CIFAR10"))
+    }
+
+    public init(remoteBinaryArchiveUrl: String, localStorageDirectory: URL) {
```
BradLarson: Rather than having `remoteBinaryArchiveUrl` as a String, I think it might make sense to use a URL type for this and a name like `remoteBinaryArchiveLocation`. Then, internally, instead of creating a URL from a string, we could use the URL directly.

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
```diff
@@ -1,11 +1,29 @@
...
+            try! FileManager.default.removeItem(atPath: path)
+        }
+    }
+
     func testCreateCIFAR10() {
         let dataset = CIFAR10()
+        verify(dataset)
+    }
+
+    func testCreateCIFAR10FromS4TFHostedBinaries() {
```
BradLarson: I'd recommend making this the default test case, and not having both this and the version with the default CIFAR10 parameters. Otherwise, the original version will still rely on the external download location instead of our new bucket, and we'll still have a potentially flaky test.

I'd just have `testCreateCIFAR10()` use our new GCS bucket.

Datasets/CIFAR10/CIFAR10.swift
xihui-wu: Good catch!

Datasets/CIFAR10/CIFAR10.swift
```diff
@@ -59,27 +69,28 @@ func downloadCIFAR10IfNotPresent(to directory: URL) {
...
-            let downloadedFile = try Data(
-                contentsOf: URL(
-                    string: "https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz")!)
+            let downloadedFile = try Data(contentsOf: location)
             try downloadedFile.write(to: URL(fileURLWithPath: archivePath))
         } catch {
             printError("Could not download CIFAR dataset, error: \(error)")
             exit(-1)
         }
+        print("Archive downloaded, processing...")
```
BradLarson:
```suggestion
        printError("Archive downloaded, processing...")
```

Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift
```diff
@@ -1,11 +1,24 @@
+import Foundation
 import TensorFlow
 import XCTest
 import Datasets
 
 final class CIFAR10Tests: XCTestCase {
+    override func setUp() {
+        super.setUp()
+        // clean up dataset files in the first place
```
BradLarson:
```suggestion
        // Force downloading of the dataset during tests by removing any pre-existing local files.
```

Datasets/CIFAR10/CIFAR10.swift
```diff
@@ -32,20 +32,30 @@ public struct CIFAR10: ImageClassificationDataset {
...
 
     public init() {
-        self.init(
-            localStorageDirectory: FileManager.default.temporaryDirectory.appendingPathComponent(
-                "CIFAR10"))
+        self.init(downloadResourceFrom: "https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz")
     }
 
-    public init(localStorageDirectory: URL) {
+    public init(downloadResourceFrom: String) {
```
BradLarson: Like with the other initializer cases, it probably makes sense to also have this take in a URL. That places a requirement on the user to provide a proper URL as input. In that case, we could change the parameter naming here to match the other instance of the initializer: `public init(remoteBinaryArchiveLocation: URL)`.

This will simplify the logic here, but will require the creation of URLs for the input to this in the tests.

PR #273, opened 2020-01-07
Reworking MiniGo and Transformer models to use a downloading checkpoint reader

BradLarson:
This creates a CheckpointReader class that currently is a wrapper around the [TensorFlowCheckpointReader](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Core/Serialization.swift) in swift-apis, but eventually should be replaced by a checkpoint reader that doesn't depend on the TensorFlow runtime. This wrapper class currently provides the convenience of being able to automatically download checkpoint files to a local temporary directory.

This greatly simplifies the process of using the MiniGo and Transformer example applications, because people no longer need set up gsutil, download the checkpoint files, place them in specific locations, and then build and run the examples from specific places. Instead, they can just call `swift run MiniGoDemo` or `swift run Transformer` and all the files will be downloaded on-demand. The Transformer example is dependent on a Python file being in a specific relative path, so it still needs to be run from the swift-models root directory.

This addresses one of the items in issue #225.


Support/Checkpoints/CheckpointReader.swift
```diff
@@ -0,0 +1,117 @@
...
+    /// The number of tensors stored in the checkpoint.
+    public var tensorCount: Int { reader.tensorCount }
+
+    /// The names of the tensors stored in the checkpoint.
+    public var tensorNames: [String] { reader.tensorNames }
+
+    /// Initializes the checkpoint reader from either a local or remote directory. If remote, 
+    /// automatically downloads the checkpoint files into a temporary directory.
+    ///
+    /// - Arguments:
```
dan-zheng: Minor: `Parameters` is the proper API doc comment that's understood by doc generation tools. 
```suggestion
    /// - Parameters:
```

Transformer/PythonCheckpointReader.swift
```diff
@@ -30,114 +31,98 @@ struct Config: Codable {
...
+        name: String,
+        scalarType: Scalar.Type
+    ) -> Tensor<Scalar> {
+        return Tensor<Scalar>(loadTensor(named: name))
+    }
 }
 
 protocol InitializableFromPythonCheckpoint {
-    init(contentsOfPythonCheckpointFile path: String, config: Config, scope: String)
+    init(reader: CheckpointReader, config: Config, scope: String)
```
dan-zheng: I know `Config` wasn't added in this PR, but how about renaming `Config` to `TransformerLMConfig` for clarity? `Config` seems like a configuration for the `TransformerLM` struct.
`

Shashi456:
@BradLarson can this `checkpointReader` idea be used for loading pretrained weights onto the image models? I'd like to try and start supporting that.

BradLarson:
@Shashi456 - Yes, it can, but I think we need to have a better or more consistent way of associating tensor names with layers in the model. The MiniGo and Transformer models have different naming schemes, as does the fast style transfer example, based on how the checkpoints were configured. Right now, this slots in under the code that associates names with layers, but I'd like to see a cleaner approach for that.

We've had this functionality inside swift-apis in the TensorFlowCheckpointReader, so this is available before this PR, but this PR makes it easy to use remotely hosted checkpoints in the models.

I have Python equivalents for some of the models in the repository, and was going to test setting up CI for inference accuracy using checkpoints generated from training those models against ImageNet, etc.

Transformer/PythonCheckpointReader.swift
BradLarson: That makes sense, I've renamed this.

Issue #253, opened 2019-12-12
Why is ResNet fixed to only two datasets?


rickwierenga:
It seems that you can only train a ResNet on these two datasets:

```swift
public enum DataKind {
    case cifar
    case imagenet
}
```

Without any useful customization options this seems unusable.

I think `DataKind` should have associated values which can be passed to the initializer to make it more flexible.

saeta:
I think it's not that you can only train on those 2 datasets, but rather that there are 2 variations of ResNet's, one that's optimized for CIFAR-sized images, and one that's optimized for bigger ImageNet images. Does that make senes? (I agree that this is a little confusing, and that a comment providing context would be very welcome as a PR!)

brettkoonce:
This is some old code from when there were only a few models in the repo that needs to be rethought.  Loosely, there are two problems this was trying to solve (see #156):

1) setting up the input filters for the network (non-obvious)
2) setting the number of output nodes (trivial)

(2) is handled better IMO by having an explicit classCount parameter, which is how the latest models work, so we should consider moving to that approach. see also: #251
(1) once we get to more complicated cv models, using a enum param name of some form is useful because many networks have multiple combined settings (eg https://github.com/tensorflow/tpu/blob/d06f0af0e48bb5e815b4db48c2d1e7168066064a/models/official/efficientnet/efficientnet_builder.py#L37)

Another route we could do is explicitly split the imagenet and cifar models into separate subtypes of  imageclassification, remove the cifar-specific logic from the current resnet variants, and then provide the resnet20/(+N*12) variants again for cifar.

BradLarson:
I think Brett and Brennan provide the right context here. The ResNet model in the repository was originally designed to test that the framework could train a model of that complexity end-to-end. The published ResNet structure was designed against the ImageNet dataset, but we lacked the pipeline to pull in and train against ImageNet, so a modified version was used that could be run against CIFAR10 ([a Python TensorFlow implementation for this](https://github.com/tensorflow/models/tree/master/official/vision/image_classification), as a reference).

[...]

PR #275, opened 2020-01-09
Making ResNet models a little more flexible

BradLarson:
Responding to comments in issue #253, this reworks the interface to the ResNet models defined in ResNet50.swift to let them generalize better to other datasets. It also renames the parameters to have them more closely reflect their use and purpose.

To generalize better, the models now take in a parameter for class counts, rather than using a hardcoded value, and replace the fixed pooling near the end with a more flexible global pooling layer. To clarify the interface, the `inputKind` and `dataKind` parameters have been replaced with `depth` and `imageSize`, which I think better illustrates their use in selecting the right size of model and the correct image size as input.


Shashi456:
@bradlarson, would renaming the file to resnet make more sense? Since there are a variety of resnet models in place.

BradLarson:
@Shashi456 - Good point, that name was left over from the original development of this model. I've renamed it to indicate its more general use.

Models/ImageClassification/ResNet.swift
```diff
@@ -20,7 +20,7 @@ import TensorFlow
 // https://arxiv.org/abs/1512.03385
 // using shortcut layer to connect BasicBlock layers (aka Option (B))
 
-public enum DataKind {
+public enum ImageSize {
```
dan-zheng: Nice renaming!

rahulbhalley: Doesn't `ImageSize` say, "size of image in `Int`?" I think it should be `DatasetKind` because:
1. We can't use name `Dataset` (already reserved for [`Dataset`](https://www.tensorflow.org/swift/api_docs/Structs/Dataset) structure).
2. And `DataKind` says something like, "data of kind either audio, image, text, and so on."

My argument is that `DatasetKind` will generalize them all as "dataset of kind either `general`, `cifar`, `vctk`, `pennTreebank`, and so on."

I included `general` case for loading other datasets not available in this pre-defined static dataset space. This option seems important for practitioners and researchers who want to use their own dataset which might contain data of any `DataKind` (audio, text, and so on).

Models/ImageClassification/ResNet.swift
```diff
@@ -317,20 +311,20 @@ public struct ResNet: Layer {
 }
 
 extension ResNet {
-    public enum Kind {
+    public enum Depth {
         case resNet50
         case resNet101
         case resNet152
```
rickwierenga: I'm sorry to interfere here.

I really like the idea of `depth` being an enum, but I think that either the cases or the enum itself should be renamed. A `resNet50` is not a depth, but an architecture so it would be more natural to call it `Architecture`. Accessing it from outside this struct would be `ResNet.Architecture` which I think is better than `ResNet.Depth`. This is also how the authors refer to it in the paper. [[ref](https://arxiv.org/pdf/1512.03385.pdf), table 1 on page 5]

This approach would also be more extendible if the team decides to implement other models because not all networks are configured by depth only.

Models/ImageClassification/ResNet.swift
```diff
@@ -273,22 +270,18 @@ public struct ResNet: Layer {
...
 
-    public var avgPool: AvgPool2D<Float>
+    public var avgPool = GlobalAvgPool2D<Float>()
     public var flatten = Flatten<Float>()
     public var classifier: Dense<Float>
 
-    public init(dataKind: DataKind, layerBlockCounts: (Int, Int, Int, Int)) {
-        switch dataKind {
+    public init(classCount: Int, imageSize: ImageSize, layerBlockCounts: (Int, Int, Int, Int)) {
+        switch imageSize {
```
rickwierenga: Consider changing this to `inputSize`. This is how Keras refers to the input and I think it would be best to keep this consistent across all TensorFlow products. [[ref](https://keras.io/applications/#resnet)]

Edit: I meant both `ImageSize` and `inputSize`.

Models/ImageClassification/ResNet.swift
```diff
@@ -242,17 +239,17 @@ public struct ResNetBasic: Layer {
...
-        switch inputKind {
+    public init(classCount: Int, depth: Depth, imageSize: ImageSize) {
+        switch depth {
         case .resNet18:
-            self.init(dataKind: dataKind, layerBlockCounts: (2, 2, 2, 2))
+            self.init(classCount: classCount, imageSize: imageSize, layerBlockCounts: (2, 2, 2, 2))
         case .resNet34:
-            self.init(dataKind: dataKind, layerBlockCounts: (3, 4, 6, 3))
+            self.init(classCount: classCount, imageSize: imageSize, layerBlockCounts: (3, 4, 6, 3))
         }
```
rickwierenga: Would it be possible to use associated values for `layerBlockCounts`? This way the initializer could be shrunken to:

```swift
public init(classCount: Int, depth: Depth, imageSize: ImageSize) {
  self.init(classCount: classCount, imageSize: imageSize, layerBlockCounts: depth.rawValue)
}
```

I'm not sure whether tuples can be used as raw values.

This would make even more sense if depth is renamed to architecture (like I mentioned in another comment).

Models/ImageClassification/ResNet.swift
BradLarson: I debated a few different names, but I settled on `depth` here because the numerical values reflect the number of convolutional layers in the network, thus their depth. The overall design is typically referred to as the ResNet architecture, so I thought that might have been a little overloaded as a term.

I don't think the design here would necessarily restrict other models. For example, in the case of MobileNet, we provide a numerical parameter to more smoothly scale the width: https://github.com/tensorflow/swift-models/blob/master/Models/ImageClassification/MobileNet.swift#L29 . That was appropriate there, where to follow the paper here we had a set number of depths and block sizes we wanted to use.

Models/ImageClassification/ResNet.swift
BradLarson: It's an interesting idea, but I don't believe a tuple can be provided as a raw value (or at least I've never tried). If you want to give that a shot, I'd be interested to see how this could be cleaned up.

Possibly a bigger win could come from somehow combining the ResNet and ResNetBasic architectures here into a single architecture that spans depths from 18 to 152.

brettkoonce: @BradLarson i got that working in an ugly manner by copy/pasting a bunch of if basicBlock { } else { <bottleneck> } logic together a little while back, would you be interested in seeing that?

PR #281, opened 2020-01-21
refactor resnet to unify namespace using block approach

brettkoonce:
feel free to rip this apart!

### DIFF
--- a/Models/ImageClassification/ResNet.swift
+++ b/Models/ImageClassification/ResNet.swift
@@ -95,24 +95,6 @@ public struct ResidualBasicBlock: Layer {
     }
 }
 
-public struct ResidualBasicBlockStack: Layer {
-    public var blocks: [ResidualBasicBlock] = []
-
-    public init(featureCounts: (Int, Int, Int, Int), kernelSize: Int = 3, blockCount: Int) {
-        for _ in 0..<blockCount {
-            blocks += [ResidualBasicBlock(featureCounts: featureCounts, kernelSize: kernelSize)]
-        }
-    }
-
-    @differentiable
-    public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
-        let blocksReduced = blocks.differentiableReduce(input) { last, layer in
-            layer(last)
-        }
-        return blocksReduced
-    }
-}
-
 public struct ResidualConvBlock: Layer {
     public var layer1: ConvBN
     public var layer2: ConvBN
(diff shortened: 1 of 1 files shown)


Models/ImageClassification/ResNet.swift
```diff
@@ -164,117 +146,74 @@ public struct ResidualIdentityBlock: Layer {
...
 }
 
 public struct ResNet: Layer {
+    @noDerivative var isBasicBlockStack: Bool
+
     public var l1: ConvBN
     public var maxPool: MaxPool2D<Float>
 
+    public var l2a_bb = ResidualBasicBlock(featureCounts: (64, 64, 64, 64))
     public var l2a = ResidualConvBlock(featureCounts: (64, 64, 64, 256), strides: (1, 1))
```
BradLarson: Having memory allocated for both the ResidualBasicBlock and ResidualConvBlock variants, when only one is used, seems inefficient. I'd like to find a way for only one layer to be set up here.

Models/ImageClassification/ResNet.swift
```diff
@@ -284,47 +223,66 @@ public struct ResNet: Layer {
...
 
     @differentiable
     public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
         let inputLayer = maxPool(relu(l1(input)))
-        let level2 = inputLayer.sequenced(through: l2a, l2b)
-        let level3 = level2.sequenced(through: l3a, l3b)
-        let level4 = level3.sequenced(through: l4a, l4b)
-        let level5 = level4.sequenced(through: l5a, l5b)
-        return level5.sequenced(through: avgPool, flatten, classifier)
+        if self.isBasicBlockStack {
```
BradLarson: I worry about the performance consequences of having a runtime switch here. As stated above, I'd like to first figure out if there's a way that we can have one layer representing the layers that are outside of the stacks and have this work without allocating both types and without a conditional check at runtime here.

brettkoonce: yeah, basically the next level of this is to redo all the sub-blocks, so to speak.  i did this as a first pass last month then started looking at going further but it's decidedly non-trivial.  my theory was this would unify the namespace as a first step.

Models/ImageClassification/ResNet.swift
brettkoonce: see below

Models/ImageClassification/ResNet.swift
```diff
@@ -164,117 +146,74 @@ public struct ResidualIdentityBlock: Layer {
...
 
-public struct ResidualIdentityBlockStack: Layer {
-    public var blocks: [ResidualIdentityBlock] = []
-
-    public init(featureCounts: (Int, Int, Int, Int), kernelSize: Int = 3, blockCount: Int) {
-        for _ in 0..<blockCount {
-            blocks += [ResidualIdentityBlock(featureCounts: featureCounts, kernelSize: kernelSize)]
+public struct ResidualBlockStack: Layer {
+    public var basicBlocks: [ResidualBasicBlock] = []
+    public var identityBlocks: [ResidualIdentityBlock] = []
```
BradLarson: Is there a way to represent both ResidualBasicBlocks and ResidualIdentityBlocks within a single array, so that we can avoid having two arrays and two parallel tracks? We'd still need to add one or the other to the array when setting things up, so that boolean value would apply, but processing should be same for both types after that point.

Models/ImageClassification/ResNet.swift
BradLarson: Inspired by an implementation in JAX, I was looking at a way of reimplementing the fundamental blocks here. If you'd like, I could try to put that together in a counterproposal PR and we could compare and contrast the two.

We're using ResNet50 as a performance benchmark, so I want to avoid any potential regressions due to architecture changes.

brettkoonce: a ground up-rewrite makes sense.  not in love with this implementation, just wanted to show something that could be done to get things closer to where they would ideally be.  this is logically the same, but not sure what the best way to profile memory/cycle differences is.

BradLarson:
As promised, here's a counterproposal PR: https://github.com/tensorflow/swift-models/pull/283 where I rework the blocks on up. Flax had a very elegant ResNet implementation, so I used that as inspiration for the basic architecture. There's one workaround and one conditional check in there that we might be able to remove at some point with additional compiler support, but I think that provides a clean and flexible implementation. I also added support for v1 vs. v1.5 variants of the architecture, to reflect both the paper and the version most benchmarks seem to use.

brettkoonce:
sounds good!

PR #283, opened 2020-01-22
Restructuring the ResNet model to merge large and small variants, choose v1 vs. v1.5

BradLarson:
This is an alternative to PR #281, in an attempt to restructure the ResNet model to simplify its overall structure and unify the 18 and 34 layer network architectures with the 50+ layer ones. It significantly reduces the overall lines of code when compared to the original model, lets one network be used for all sizes, and adds the flexibility of choosing both v1 and v1.5 variants.

The [implementation of ResNet within the Flax library](https://github.com/google-research/flax/blob/prerelease/examples/imagenet/models.py) was the source for the basic implementation structure, but this expands upon that by adding the basic blocks for ResNet 18 and 34, as well as support for both ResNet v1 and v1.5 variants.

A workaround is currently used to handle the differences between blocks that require projection and those that don't (a dummy tiny convolution / batchnorm pair is allocated but never used in non-projection case). That will eventually be removed when differentiation support for optionals is added.

### DIFF
--- a/Models/ImageClassification/ResNet.swift
+++ b/Models/ImageClassification/ResNet.swift
@@ -12,27 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import TensorFlow
 import Foundation
+import TensorFlow
 
 // Original Paper:
 // "Deep Residual Learning for Image Recognition"
 // Kaiming He, Xiangyu Zhang, Shaoqing Ren, Jian Sun
 // https://arxiv.org/abs/1512.03385
 // This uses shortcut layers to connect residual blocks (aka Option (B)).
+//
+// The structure of this implementation was inspired by the Flax ResNet example:
+// https://github.com/google-research/flax/blob/prerelease/examples/imagenet/models.py
 
-public protocol Projection: Module {
-    @differentiable func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float>    
-}
-
-public struct Passthrough: ParameterlessLayer, Projection {
-    @differentiable
-    public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
-        return input
-    }
-}
-
-public struct ConvBN: Layer, Projection {
+public struct ConvBN: Layer {
     public var conv: Conv2D<Float>
     public var norm: BatchNorm<Float>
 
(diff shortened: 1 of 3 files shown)


Models/ImageClassification/ResNet.swift
```diff
@@ -12,18 +12,17 @@
...
-// using shortcut layer to connect BasicBlock layers (aka Option (B))
-
-public enum ImageSize {
-    case cifar
-    case imagenet
-}
+// This uses shortcut layers to connect residual blocks (aka Option (B)).
+//
+// The structure of this implementation was inspired by the Flax ResNet example:
+// https://github.com/google-research/flax/blob/prerelease/examples/imagenet/models.py
```
brettkoonce: something something mystery box something something 🎁

Models/ImageClassification/ResNet.swift
```diff
@@ -44,287 +43,154 @@ public struct ConvBN: Layer {
...
 }
 
 extension ResNet {
+    public enum ImageSize {
+        case cifar
+        case imagenet
+    }
+
     public enum Depth {
+        case resNet18
```
shabalind: Can we make `Depth` a protocol to avoid hardcoded list of variants? More concretely, can we let users define their own resnet depth variant as long as it provides `usesBasicBlocks` & `layerBlockSizes`?

Models/ImageClassification/ResNet.swift
```diff
@@ -44,287 +43,154 @@ public struct ConvBN: Layer {
...
+        }
+        return blocksReduced.sequenced(through: avgPool, flatten, classifier)
     }
 }
 
 extension ResNet {
+    public enum ImageSize {
+        case cifar
+        case imagenet
+    }
```
shabalind: Similarly to `Depth`, can we provide more flexibility here with custom image sizes?

Models/ImageClassification/ResNet.swift
BradLarson: Because of the different layer groupings, I don't know that we'd be able to have a fully continuous spectrum of sizes. These are the ones defined in the paper, and the most commonly used variants. In the Flax implementation, they use a lookup table and throw an error if you provide a value outside of that.

I had wanted to leave this as an enum to provide a compile-time check to verify that you're using a supported architecture size. This could be simplified into a dictionary with entries for these sizes, and an optional check when pulling values out of the dictionary, but I generally like to tend toward compile-time errors as opposed to runtime ones.

shabalind: Yep, makes sense! 👍

Models/ImageClassification/ResNet.swift
BradLarson: This is a good point. With the use of global pooling in the last stage, we can use a pretty variable set of input sizes. The two cases we need to account for are whether or not to downsample by 4X during the first stage. ImageNet-class images, we'll want to downsample, but for CIFAR-sized images, we don't.

I've made this now an explicit boolean parameter, with it set to a default of true for the common ImageNet-scale cases.

PR #284, opened 2020-01-23
Update download link for MNIST

WilliamHYZhang:
We should use the Google storage api link to load the MNIST dataset to avoid server overload. TFDS also uses this approach. (see [here](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image/mnist.py))

cc: @rickwierenga


rickwierenga:
Awesome!

PR #288, opened 2020-01-26
Add densenet169, densenet210 models

Shashi456:
Initial addition of densenet models, If anyone wants to take it up, the densenet models can be abstracted into neater code, I'll do it in a follow up PR later this week.


Shashi456:
Also @BradLarson could you add a few Image classificaiton model checkpoints to the bucket you made for models? 
https://github.com/tensorflow/models/tree/master/research/slim#Pretrained

Most of them are tar files containing the `.ckpt` file, I'd like to start support basic pretrained model loading for the models we have.

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,220 @@
...
+        let level2 = level1.sequenced(through: denseBlock2, transitionLayer2)
+        let level3 = level2.sequenced(through: denseBlock3, transitionLayer3)
+        let output = level3.sequenced(through: denseBlock4, globalAvgPool, dense)
+        return output
+    }
+}
+
+
+public struct Conv: Layer {
+    public var batchNorm: BatchNorm<Float>
```
rickwierenga: Consider renaming this to `BatchNormConv2DLayer`. It is a little confusing right now.

[[keras reference](https://github.com/tensorflow/swift-models/blob/master/Models/ImageClassification/ResNetV2.swift#L43)]

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,220 @@
...
+            filterShape: (filterSize, filterSize, inputFilterCount, outputFilterCount),
+            strides: (stride, stride),
+            padding: .same
+        )
+    }
+
+    @differentiable
+    public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+        conv(relu(batchNorm(input)))
+    }
```
rickwierenga: Is there a reason you use chaining over `sequenced(through:)` here?

Models/ImageClassification/DenseNet.swift
BradLarson: Rick has a good point that Conv probably isn't descriptive enough of a name for a convolution + batchnorm layer. In the ResNet implementation, we refer to this as [ConvBN](https://github.com/tensorflow/swift-models/blob/master/Models/ImageClassification/ResNet.swift#L27). Would it even be possible to re-use the ResNet's implementation of this and / or extract it as a common implementation of this subunit?

Models/ImageClassification/DenseNet.swift
```diff
@@ -0,0 +1,220 @@
...
+        stride: 2,
+        inputFilterCount: 3,
+        outputFilterCount: 64
+    )
+    public var maxpool = MaxPool2D<Float>(
+        poolSize: (3, 3),
+        strides: (2, 2),
+        padding: .same
+    )
+    public var denseBlock1 = DenseBlock(repetitionCount: 6, inputFilterCount: 64)
```
BradLarson: Rather than having three discrete implementations of DenseNets, would it be possible to have one general model and three sets of block sizes that could be used to generate the three variants, in the same way as is done with the current ResNet implementation?

Models/ImageClassification/DenseNet.swift
8bitmp3: Second that. The users would get a sense of consistency.

Shashi456: @BradLarson I'm working on the general reduction of the implementations and was hoping to add that in a follow up PR. This PR was to generally support the rest of the densenet models. I will rename it to `ConvBN`.

Models/ImageClassification/DenseNet.swift
rickwierenga: That would be a nice simplification, and it's probably nice to use an enum for this.

BradLarson:
Just checking in, because it's been a little while. Are you still interested in bringing these models in? I still like the idea of generalizing the three variants into a single model, like we did with the ResNets. I think that would organize this model nicely. Additionally, the minor renaming that had been suggested still seems to make sense.

If those couple of items were addressed, we'd be still glad to bring this in.

Shashi456:
@bradlarson alright I'll push the generalized versions soon.

texasmichelle:
Reminder to add generalized versions.

marcrasi:
Hi, there hasn't been activity on this PR for a while. We're going to close it to keep our list of open PRs small. Of course, feel free to reopen any time if you have time to come back and work on this!

PR #290, opened 2020-01-27
Add reference to the bag of tricks paper

sgugger:


sgugger:
@googlebot I signed it!

PR #291, opened 2020-01-28
"Swiftify" VGG

rickwierenga:
I'm sorry this is very nitpicky.


Shashi456:
I think this is the file that comes after running `swift-format`, so are manual changes to the format necessary?

rickwierenga:
[The Swift Style Guide](https://google.github.io/swift/#line-wrapping) actually recommends to place every parameter on a separate line. But if `swift-format` prefers this way that's absolutely fine.

BradLarson:
Yes, in this case the file was formatted via `swift-format`, following the [Google Swift style guide](https://google.github.io/swift/). We'll probably want to remain with that.

Thanks for suggesting this, though.

PR #292, opened 2020-01-29
Fix the shape of the last dense layer in Resnet18 and ResNet34

sgugger:
Since there is no *4 expansion in the basic ResNets, the last number of filters is 512, not 2048, so that should be the number of inputs of the last dense layer. This PR fixes that.

### DIFF
--- a/Models/ImageClassification/ResNet.swift
+++ b/Models/ImageClassification/ResNet.swift
@@ -151,7 +151,7 @@ public struct ResNet: Layer {
             }
         }
 
-        classifier = Dense(inputSize: 2048, outputSize: classCount)
+        classifier = Dense(inputSize: depth.usesBasicBlocks ? 512 : 2048, outputSize: classCount)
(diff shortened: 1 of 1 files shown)


Models/ImageClassification/ResNet.swift
```diff
@@ -79,7 +80,7 @@ public struct ResidualBlock: Layer {
                     ConvBN(filterShape: (1, 1, inputFilters, filters), strides: strides))
                 earlyConvs.append(ConvBN(filterShape: (3, 3, filters, filters), padding: .same))
             }
-            lastConv = ConvBN(filterShape: (1, 1, filters, filters * 4))
+            lastConv = ConvBN(filterShape: (1, 1, filters, outFilter))
```
BradLarson:
```suggestion
            lastConv = ConvBN(filterShape: (1, 1, filters, outFilters))
```

PR #293, opened 2020-01-30
Made some minor edits to get the BERT classifier training to work for CoLA.

eaplatanios:
@dan-zheng this is what we discussed earlier. Please let me know if the model seems to be learning once you add the evaluation function call. :)


dan-zheng:
Merging to unblock progress. We can continue fine-tuning and evaluation in the `bert-wip` branch!

eaplatanios:
No problem! :) The loss will converge to some value so that's good. I'm not sure how to interpret the 0.695 value. Evaluation will be useful for that purpose as cross entropy values are sometimes hard to interpret.

PR #294, opened 2020-01-31
Fixed BERT.

eaplatanios:
There were a couple of issues that seem fixed now:
1. The sigmoid cross entropy loss was applied on logits of shape [B, 1] and labels of shape [B]. This forced a silent broadcast of logits to shape [B, B], which resulted in the loss not being informative for training.
2. The batch size was too small. I added a comment in the main script code explaining how batching works in my data pipelines.
3. This is a minor one but there is bug with how I was copying the prefetching iterator. This was a temporary solution but I just disabled prefetching for the dev and test sets so that the dev set is the same across runs.

This is not currently tuned but it's working. After ~20 steps MCC should be at about 0.28 and after ~200 steps it should be getting close to 0.50.

cc @dan-zheng

### DIFF
--- a/Models/Text/Architecture.swift
+++ b/Models/Text/Architecture.swift
@@ -1,9 +1,0 @@
-// Adapted from: https://github.com/eaplatanios/nca/blob/master/Sources/NCA/Architecture.swift
-
-public struct ArchitectureInput {
-  public let text: TextBatch?
-
-  public init(text: TextBatch? = nil) {
-    self.text = text
-  }
-}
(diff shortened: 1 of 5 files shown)


eaplatanios:
By the way, if you set the max sequence length to 512 and the batch size to 4096, training should be faster. I get MCC > 0.3 after 30 steps and reach ~0.5 in ~60 steps.

eaplatanios:
No problem at all. I hope everything works fine now! :)

PR #269, opened 2020-01-02
Add final image to DCGAN README

rickwierenga:

PR #297, opened 2020-02-01
Force redownloading of MNIST files when testing

rickwierenga:
Similar to https://github.com/tensorflow/swift-models/blob/master/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift.

This also shows [a commit](https://github.com/tensorflow/swift-models/commit/c726c649995644766b16700c49292e2737bbd4d5) for [an old PR](https://github.com/tensorflow/swift-models/pull/269).


BradLarson:
Talking with others here, I wonder if we should go in the opposite direction. Nominally, you want unit tests to start in the same state every time they are run, so that was our thinking in making the CIFAR-10 test wipe itself before each run.

In practice, this is getting to be a problem. The CIFAR-10 dataset is ~180 MB in size, and MNIST adds another ~20 MB. Re-downloading nearly 200 MB of data every single time the test suite is run is significantly slowing these tests, especially when I try to run them locally in areas with limited bandwidth. It's also wasteful of resources.

I'm leaning towards removing the cleaning of the cache from the CIFAR-10 tests, which would also remove the need for this pull request. Does that sound sensible?

rickwierenga:
Fair point. I agree. Another argument to be made is that this library should not depend on the state of external resources to pass tests.

Would it be possible to run certain tasks via the `swift` command line "application"? I've tried doing this but could not quite figure out how. If it is possible, we could write separate tasks to download the data.

Another thing is that we do not want the developer to download Imagenette, Imagewang, and other large datasets when they are working on MobileNet, for example. Having separate test targets would facilitate this. @WilliamHYZhang and I are also struggling with this in our s5tf-team/datasets repo. If this is not currently possible, it might be worth exploring that on the `tensorflow` branch of Swift. (I think that should be possible, but I am not 100% sure on that)

BradLarson:
Based on what we discussed, I think I'll close this out and we will work on removing the wiping of the cache from the CIFAR-10 tests, as well. Thanks for bringing it up.

PR #300, opened 2020-02-07
Add ResNet V2 and remove legacy file

sgugger:
This pull request adds the resnet V2 model that was trained internally on TPU to full accuracy adn removes the old legacy ResnetV2.swift file.
Tests have been updated to check each version of the model runs at inference.

### DIFF
--- a/Tests/ImageClassificationTests/Inference.swift
+++ b/Tests/ImageClassificationTests/Inference.swift
@@ -137,23 +137,23 @@ final class ImageClassificationInferenceTests: XCTestCase {
             randomNormal: [1, 224, 224, 3], mean: Tensor<Float>(0.5),
             standardDeviation: Tensor<Float>(0.1), seed: (0xffeffe, 0xfffe))
         
-        let resNet18 = ResNet(classCount: 1000, depth: .resNet18)
+        let resNet18 = ResNetV2(classCount: 1000, depth: .resNet18)
         let resNet18Result = resNet18(input)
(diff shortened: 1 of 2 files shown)


Tests/ImageClassificationTests/Inference.swift
```diff
@@ -136,13 +136,26 @@ final class ImageClassificationInferenceTests: XCTestCase {
...
             standardDeviation: Tensor<Float>(0.1), seed: (0xffeffe, 0xfffe))
-        let resNet18ImageNet = PreActivatedResNet18(imageSize: 224, classCount: 1000)
-        let resNet18ImageNetResult = resNet18ImageNet(input)
-        XCTAssertEqual(resNet18ImageNetResult.shape, [1, 1000])
-
-        let resNet34ImageNet = PreActivatedResNet34(imageSize: 224, classCount: 1000)
-        let resNet34ImageNetResult = resNet34ImageNet(input)
-        XCTAssertEqual(resNet34ImageNetResult.shape, [1, 1000])
+        
+        let resNet18 = ResNet(classCount: 1000, depth: .resNet18)
```
BradLarson: Should these be ResNetV2 instances, instead of ResNet?

sgugger: Oops!

sgugger: Updated the tests to use the V2, good catch!

PR #301, opened 2020-02-09
Executable for VGG-Imagewoof

kongzii:
While investigating example for VGG-Imagewoof I found that there is a line about how to execute it:

```sh
cd swift-models
swift run -c release VGG-Imagewoof
```

but actually no executable for that in `Package.swift`. So just a simple PR that adds it.


brettkoonce:
good catch! 😅

rickwierenga:
Nice!

BradLarson:
I thought maybe this had been intentionally left off, but if not, great to have it added.

PR #302, opened 2020-02-10
add fashion mnist dataset

brettkoonce:


BradLarson:
As with the previous datasets, it would be great to have an example that uses this to verify that everything was being pulled in and used correctly. I've commonly seen Fashion MNIST used with simple GAN models, so would it be possible to rework one of the existing GAN / Autoencoder examples to instead work with Fashion MNIST? That would provide a neat, visible demonstration of the dataset working.

8bitmp3:
Just a s4tf user here sharing my 2c:

@brettkoonce this is awesome, thanks! Fashion is more fun than handwritten digits.

@BradLarson Re: demo - I've just run Brett's FashionMNIST in Colab with a GPU on a s4tf GAN with all the `ModelSupport` and `Datasets` packages from `/swift-models/` as separate cells. Over 10 epochs it seems to not break, which is great. [Notebook](https://github.com/8bitmp3/testing-s4tf-things/blob/master/test_fashion_mnist_gan_s4tf_feb_12_2020.ipynb).

I'm sure the team is aware of this, so, just in case, I think MNIST is currently "hardcoded" into the s4tf versions of GAN, DCGAN and autoencoder as `let dataset = MNIST(...)`, so maybe later introducing something like an enum would make sense.

Cheers all!

brettkoonce:
@8bitmp3 glad you like it! ☺️

@BradLarson i'll modify the autoencoder demo here in a sec, my machine is offline.

i was going to add this and the kmnist dataset, which are intended as drop-in mnist replacements.  there's probably some code reuse possible as well.

brettkoonce:
@BradLarson let me know what you think!

Shashi456:
Could we pass the dataset to the Autoencoder as an option?  A standard MNIST example should definitely be kept around.

BradLarson:
I might be missing something, but I see updates to the Autoencoder Readme and images (which look great), but not to the Autoencoder source itself. Did the changes not get checked in?

@Shashi456 - I'm fine with the Autoencoder example being targeted at a specific dataset. It's merely a simple example of an autoencoder, and the behavior won't fundamentally change with which dataset it's on. I just like having examples that exercise the different datasets and models such that we can feel confident they work. We have a number of them using MNIST, it's nice to have at least one using the new FashionMNIST.

brettkoonce:
@BradLarson re main.swift, doh! 

my plan is to eventually add a 2d autoencoder example as well, fwiw.

PR #303, opened 2020-02-10
build: add a CMake based build for the tensorflow-swift-models

compnerd:
This adds a CMake based build system for the models, enabling building
the models on platforms where swift-package-manager does not currently
function properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,18 @@
+cmake_minimum_required(VERSION 3.16)
+project(Models
+  LANGUAGES Swift)
+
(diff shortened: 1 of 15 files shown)


BradLarson:
Because many of the CMake build files reference specific files, rather than directories of files, I can see people accidentally breaking these as they add or remove files to models. In the same way that we have a CI test for using SwiftPM to build all targets, do we have plans for enhancing our presubmit CI to also build these via CMake to make sure we don't break the CMake builds as well?

Also, will we need to add CMake build instructions in parallel to our Readme files?

compnerd:
I think that getting CI for the CMake based build would be great.  I know for the swift-apis we do actually have CI to ensure that it continues to work.

I'm happy to add instructions on how to build with the CMake based build, especially now that I have it building :)

PR #306, opened 2020-02-10
docs: document CMake support

compnerd:
Add some documentation on how to use the CMake based build and some of
the limitations of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,50 @@ swift build
 swift test
 ```
 
+### *Experimental* CMake Support
+
+There is experimental support for building with CMake.  This is required to build the models on Windows, and can also be used to cross-compile the models and the demo programs.
+
+**NOTE**: tests are currently not supported with the CMake based build.
+
+It is highly recommended that you use CMake 3.16 or newer to ensure that `-B` and parallel builds function properly in the example commands below.
+
+macOS:
+
+```
+cmake                                                              \
+  -B /BinaryCache/tensorflow-swift-models                          \
+  -D CMAKE_BUILD_TYPE=Release                                      \
+  -D CMAKE_Swift_COMPILER=$(TOOLCHAINS=tensorflow xcrun -f swiftc) \
+  -G Ninja                                                         \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,6 +26,40 @@ swift build
...
+
+**NOTE**: tests are currently not supported with the CMake based build.
+
+It is highly recommended that you use CMake 3.16 or newer to ensure that `-B` and parallel builds function properly in the example commands below.
+
+macOS:
+
+```
+cmake                                                              \
+  -B /BinaryCache/tensorflow-swift-models                          \
```
saeta: Building into a folder on the root partition is probably a bad idea?

README.md
```diff
@@ -26,6 +26,40 @@ swift build
...
 
+### *Experimental* CMake Support
+
+There is experimental support for building with CMake.  This is required to build the models on Windows, and can also be used to cross-compile the models and the demo programs.
+
+**NOTE**: tests are currently not supported with the CMake based build.
+
+It is highly recommended that you use CMake 3.16 or newer to ensure that `-B` and parallel builds function properly in the example commands below.
+
+macOS:
```
saeta: Does this also work on Linux?

README.md
```diff
@@ -26,6 +26,40 @@ swift build
 swift test
 ```
 
+### *Experimental* CMake Support
+
+There is experimental support for building with CMake.  This is required to build the models on Windows, and can also be used to cross-compile the models and the demo programs.
+
+**NOTE**: tests are currently not supported with the CMake based build.
```
saeta: Note: this is definitely not ideal. :-(

README.md
compnerd: Yes, added Linux instructions

README.md
compnerd: The docker build does that actually.  It should be obvious which paths to adjust though.  I think having that for the example command is reasonable.

README.md
compnerd: Agreed, I hope that we can get that added soon.

PR #314, opened 2020-02-13
Update README.md

compnerd:
Add instructions covering `regex` dependency

### DIFF
--- a/Transformer/README.md
+++ b/Transformer/README.md
@@ -39,4 +39,10 @@ To get a toolchain, you can:
 
 It also currently requires Python 3.x.
 
+The `regex` module is also required for Python.  If this is not available, you can install it via:
+
+```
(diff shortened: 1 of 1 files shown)


Transformer/README.md
```diff
@@ -39,4 +39,11 @@ To get a toolchain, you can:
 
 It also currently requires Python 3.x.
 
+The `regex` and `numpy` module are also required for Python.  If this is not available, you can install it via:
```
saeta: Nit: line wrapping.

compnerd: It seems that the rest of the file doesn't wrap at 80-cols either.  It relies on Markdown to handle that for us when rendering.  I suspect that this is to aid editing the file using the GitHub web editor.

PR #316, opened 2020-02-14
Add Batcher and example

sgugger:
This PR adds Batcher to swift-models and show one example of use in Imagenette.


Package.swift
```diff
@@ -32,7 +33,9 @@ let package = Package(
         .package(url: "https://github.com/kylef/Commander.git", from: "0.9.1"),
     ],
     targets: [
-        .target(name: "Datasets", dependencies: ["ModelSupport"], path: "Datasets"),
+        .target(name: "Batcher", path: "Batcher"),
+        .target(name: "ImageClassificationModels", path: "Models/ImageClassification"),
```
brettkoonce: dupe line?

brettkoonce:
cool stuff! 🤩

Package.swift
sgugger: Oh yes, thanks for flagging!

8bitmp3:
sorry to jump in here - can I suggest changing the copyright year to 2020 or 2019-2020. this is great stuff @sgugger

sgugger:
Ah yes, I just copied and pasted without thinking ;)

BradLarson:
It looks like CI is failing on missing CMake files for the new Batcher components with that new parallel build path. I've added those with other recent commits, if you need templates for what should be added. If you have any trouble with those, I can push the appropriate files into the PR.

Examples/MobileNet-Imagenette/main.swift
```diff
@@ -16,24 +16,24 @@ import Datasets
...
 import TensorFlow
 
-let batchSize = 16
-
-let dataset = Imagenette(inputSize: .resized320, outputSize: 224)
-let testBatches = dataset.testDataset.batched(batchSize)
+let batchers = ImagenetteBatchers(
+    inputSize: .resized320, 
+    outputSize: 224, 
+    batchSize: 128,
```
brettkoonce: what gpu do you have 🤯

sgugger: Just a TITAN RTX in this case ;). Will lower a bit, but if S4TF wasn't so greedy it could fit on most cards.

brettkoonce:
ps setting the batchSIze to 16 with numWorkers = 8 produces this error, not sure if you want to chase it:

```
Starting training...                                                            
Fatal error: Can't form Range with upperBound < lowerBound: file /swift-base/swi
ft/stdlib/public/core/Range.swift, line 709                                     
Fatal error: Can't form Range with upperBound < lowerBound: file /swift-base/swi
ft/stdlib/public/core/Range.swift, line 709                                     
Fatal error: Can't form Range with upperBound < lowerBound: file /swift-base/swi
ft/stdlib/public/core/Range.swift, line 709   
```

sgugger:
That's weird, I didn't get any error on my side. Let me try again.

sgugger:
Found the reason: when training at batch size 16, the last batch in the validation set was of size 4 < numWorkers, which was then crashing concurrentMap. Tested training runs at batchSize 16 now (though lr needs to be adjusted).

Batcher/Batcher.swift
```diff
@@ -101,11 +101,12 @@ public struct BatchIterator<C: Collection>: IteratorProtocol, Sequence where C.I
     public mutating func next() -> C.Element? {
         guard pos < samplesCount else { return nil }
         let end = Swift.min(pos + b.batchSize, samplesCount)
+        print(pos,end)
```
brettkoonce: might nuke this fwiw

sgugger: Oh, good catch. It was for debugging the problem. Will remove

Batcher/Backend.swift
```diff
@@ -0,0 +1,49 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import TensorFlow
+
+// ThreadSafe and concurrentMap based on https://talk.objc.io/episodes/S01E90-concurrent-map
+// TODO: build a proper separate module that does the parallel processing
```
dabrahams: Missing a doc comment.  Every declaration [should have one](https://swift.org/documentation/api-design-guidelines/#write-doc-comment).  We'll be discussing this Friday, to get context.

[example1](https://gist.github.com/dabrahams/9f81ed0799b36046370c4bd0d7561c41), [example2](https://gist.github.com/dabrahams/1fa37dd518cac97209db5fedbfc09ce0)

Batcher/Backend.swift
```diff
@@ -0,0 +1,49 @@
...
+            let min = i*cs
+            let max = min+cs>count ? count : min+cs
+            for idx in (min..<max) {
+                let element = self[idx]
+                let transformed = transform(element)
+                result.atomically { $0[idx] = transformed }
+            }
+        }
+        return result.value.map { $0! }
+    }
```
dabrahams: That is a truly horrible implementation; it synchronizes as every element is processed!  For many workloads, it will be a slowdown.  You can get it to speed up *some* mapping operations, but the transform has to be *really* slow to overcome the inherent overhead.

It's not clear from looking at the code that the authors really knew what they were doing; by default the inner `for` loop handles just one element.

Please see [this gist](https://gist.github.com/dabrahams/ea5495b4cccc2970cd56e8cfc72ca761) which demonstrates.  The first line shows how to run it.

Batcher/Batcher.swift
```diff
@@ -0,0 +1,138 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import TensorFlow
+
+// Return the input with no operation done
```
dabrahams: So this should be a [documentation comment](https://swift.org/documentation/api-design-guidelines/#write-doc-comment), starting with three slashes.  It's pretty close to acceptable if you add a slash, but “the input” is unnecessarily vague, and it has more words than necessary.  

Tiny nits: `T` is sort of the standard type parameter name when it's really fully generic, and you can leave out `return`.I would write:

```swift
/// Returns `x`.
public func identity<T>(_ x: T) -> T { x }
```

Batcher/Batcher.swift
```diff
@@ -0,0 +1,138 @@
...
+public func identity<C>(_ x: C) -> C { return x }
```
dabrahams: Tiny nits: `T` is sort of the standard type parameter name when it's really fully generic, and you can leave out “`return`”.  I would write:

```swift
public func identity<T>(_ x: T) -> T { x }
```

Batcher/Batcher.swift
```diff
@@ -0,0 +1,138 @@
...
+
+// Default sample on a collection
```
dabrahams: I don't know what this comment means.  The line after it would almost be an ideal doc comment, though.

Batcher/Batcher.swift
```diff
@@ -0,0 +1,138 @@
...
+// Returns indices from 0 to dataset.count, potentially shuffled
```
dabrahams:
```swift
// Returns the elements of `0..<dataset.count`, in that order if `shuffled == false` and randomly shuffled otherwise.
```

This summary was hard to write and I was (strongly!) tempted to describe it by writing out its implementation.  That is often a sign that the thing being documented doesn't represent a valuable/usable abstraction.  What do you think?

Batcher/Batcher.swift
```diff
@@ -0,0 +1,138 @@
...
+public func defaultSample<C: Collection>(on dataset: inout C, shuffled: Bool) -> [Int] {
```
dabrahams: I don't understand the name for the same reason I didn't understand the first line of the comment above: I don't know what a “default sample on a dataset” means.  

It's very unusual that `dataset` is passed `inout`, since you never mutate it.  This leads me to suspect you never tested this method because you'd have had to been operating on a mutable dataset and the compiler would have made you put a `&` before the argument.  Either that, or there's some very subtle good reason for the `inout` and that should be documented here.

sgugger: This the default implementation of a method used in Batcher. While this one does not mutate the dataset, some others might (for instance, the sample method used for a `LanguageModelDataset` will call `dataset.shuffle()`). 

Unless you can pass a method with `dataset` not inout for something that has the signature `(inout C, Bool) -> [Int]`, the inout is necessary here.

Batcher/Backend.swift
sgugger: Thanks for the gist!

PR #319, opened 2020-02-14
move bert demo to examples namespace

brettkoonce:
fixes import warnings, eg "'TensorFlowModels' /home/skoonce/swift-models: warning: ignoring duplicate product 'TextModels'"


Package.swift
```diff
@@ -36,6 +36,7 @@ let package = Package(
         .target(name: "ModelSupport", dependencies: ["SwiftProtobuf"], path: "Support"),
         .target(name: "ImageClassificationModels", path: "Models/ImageClassification"),
         .target(name: "TextModels", dependencies: ["Datasets"], path: "Models/Text"),
+        .target(name: "Bert-CoLA", dependencies: ["TextModels"], path: "Examples/Bert-CoLA"),
```
saeta: I think you also might need to add the target above in the products section.

Also, nit: Could we call this: `BERT-CoLA`?

brettkoonce: @saeta let me know!

BradLarson:
I think we may have collided here, because I did the same with PR #321 that I just submitted. I had been working on a larger reorganization of the model so that the dataset and an example like the above could be extracted. I made a few more elaborate changes to the overall structure as a first step so that we can remove some duplicate code and reuse some of this in the Transformer example.

I liked your phrasing better in your Readme, so I used that for the text in the BERT-CoLA Readme I submitted as part of my pull request. Thanks for writing that up.

Sorry for not making it clear that I was working on this. If you'd prefer to have your PR landed first, I think I can rework mine to be a follow-on to yours.

brettkoonce:
nbd, carry on!

BradLarson:
Thanks, again sorry about that. Appreciate the cleanup work.

PR #325, opened 2020-02-15
BERT: remove hardcoded path assumption

compnerd:
`/tmp` is not the universal temporary directory.  Construct the path
in a more portable manner by using Foundation.  This also allows the
user to override the temporary path.

### DIFF
--- a/Examples/BERT-CoLA/main.swift
+++ b/Examples/BERT-CoLA/main.swift
@@ -18,7 +18,9 @@ import TensorFlow
 import TextModels
 
 let bertPretrained = BERT.PreTrainedModel.bertBase(cased: false, multilingual: false)
-let workspaceURL = URL(fileURLWithPath: "/tmp/bert_models", isDirectory: true)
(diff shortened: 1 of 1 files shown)


Examples/BERT-CoLA/main.swift
```diff
@@ -18,7 +18,9 @@ import TensorFlow
 import TextModels
 
 let bertPretrained = BERT.PreTrainedModel.bertBase(cased: false, multilingual: false)
-let workspaceURL = URL(fileURLWithPath: "/tmp/bert_models", isDirectory: true)
+let workspaceURL = URL(fileURLWithPath: "bert_models", isDirectory: true,
+                       relativeTo: URL(fileURLWithPath: NSTemporaryDirectory(),
+                                       isDirectory: true))
```
saeta: At some point, I'd like to have some syntactic sugar for the filesystem. :-( I find this quite cumbersome.

eaplatanios: Using `.appendingPathComponent` looks a bit less verbose and nested, but I agree that some syntactic sugar would be nice. :)

PR #327, opened 2020-02-18
Make `Image` more flexible

rahulbhalley:
Following changes have been proposed on `Image` structure:

1. Exposed `format` and `quality` argument labels to `saveImage(_:size:format:directory:name:quality:)` public function with default values of `format: .rgb` (because there's high probability the image will be RGB), `quality: 95`.
2. `size` is now used to resize the `tensor` instead of reshaping it. (There doesn't seem to be a requirement of reshaping image tensor.) But if it does seem important to community I think of renaming `size` to `newShape` for `reshaped(to:)` and `newSize` for resizing `tensor`.
3. `size` argument label is now `(Int?, Int?)` with default values `(nil, nil)`. If new argument value is provided during call then image tensor is resized otherwise original size is maintained.
4. `resized(to:)` instance method now uses `rankLifted()` and `squeezingShape(at:)` for more legibility. 

Suggestion(s) for further improvement:

1. Add `_Raw.rGBToGrayscale()` (akin to Pythonic [`tf.image.rgb_to_grayscale`](https://www.tensorflow.org/api_docs/python/tf/image/rgb_to_grayscale?version=nightly)) method inside `save(to:format:quality)` at beginning. Currently this method is not available under `_Raw` enumeration. Using this method will automatically convert RGB to grayscale using `format` value from `saveImage(_:size:format:directory:name:quality:)` to save image.
2. In similar context to previous point add `_Raw.GrayscaleToRGB()` (akin to [`tf.image.grayscale_to_rgb`](https://www.tensorflow.org/api_docs/python/tf/image/grayscale_to_rgb)).

### DIFF
--- a/DCGAN/main.swift
+++ b/DCGAN/main.swift
@@ -153,8 +153,8 @@ for epoch in 0 ... epochs {
     // Render images.
     let generatedImage = generator(noise)
     try saveImage(
-        generatedImage, size: (28, 28), directory: outputFolder,
-        name: "\(epoch).jpg")
+        generatedImage, shape: (28, 28), format: .grayscale, directory: outputFolder,
+        name: "\(epoch)")
 
     // Print loss.
     let generatorLoss_ = generatorLoss(fakeLabels: generatedImage)
@@ -165,5 +165,5 @@ for epoch in 0 ... epochs {
 let noise1 = Tensor<Float>(randomNormal: TensorShape(1, 100))
 let generatedImage = generator(noise1)
 try saveImage(
-        generatedImage, size: (28, 28), directory: outputFolder,
-        name: "final.jpg")
+    generatedImage, shape: (28, 28), format: .grayscale, directory: outputFolder,
+    name: "final")
--- a/Support/Image.swift
+++ b/Support/Image.swift
@@ -20,29 +20,29 @@ public struct Image {
         case bgr
         case rgb
     }
-    
+
     enum ImageTensor {
         case float(data: Tensor<Float>)
         case uint8(data: Tensor<UInt8>)
     }
-    
+
     let imageData: ImageTensor
-    
+
     public var tensor: Tensor<Float> {
         switch self.imageData {
         case let .float(data): return data
         case let .uint8(data): return Tensor<Float>(data)
         }
     }
-    
+
     public init(tensor: Tensor<UInt8>) {
         self.imageData = .uint8(data: tensor)
     }
-    
+
     public init(tensor: Tensor<Float>) {
         self.imageData = .float(data: tensor)
     }
-    
+
     public init(jpeg url: URL, byteOrdering: ByteOrdering = .rgb) {
         let loadedFile = _Raw.readFile(filename: StringTensor(url.absoluteString))
         let loadedJpeg = _Raw.decodeJpeg(contents: loadedFile, channels: 3, dctMethod: "")
@@ -53,7 +53,7 @@ public struct Image {
             self.imageData = .uint8(data: loadedJpeg)
         }
     }
-    
+
     public func save(to url: URL, format: _Raw.Format = .rgb, quality: Int64 = 95) {
         let outputImageData: Tensor<UInt8>
         switch format {
@@ -77,12 +77,12 @@ public struct Image {
             print("Image saving isn't supported for the format \(format).")
             exit(-1)
         }
-        
+
         let encodedJpeg = _Raw.encodeJpeg(
             image: outputImageData, format: format, quality: quality, xmpMetadata: "")
         _Raw.writeFile(filename: StringTensor(url.absoluteString), contents: encodedJpeg)
     }
-    
+
     public func resized(to size: (Int, Int)) -> Image {
         switch self.imageData {
         case let .uint8(data):
--- a/Support/Image.swift
+++ b/Support/Image.swift
@@ -100,11 +100,20 @@ public struct Image {
     }
 }
 
-public func saveImage(_ tensor: Tensor<Float>, size: (Int, Int)? = nil,
+public func saveImage(_ tensor: Tensor<Float>, shape: (Int, Int), size: (Int, Int)? = nil,
                       format: _Raw.Format = .rgb, directory: String, name: String,
                       quality: Int64 = 95) throws {
     try createDirectoryIfMissing(at: directory)
-    let image = Image(tensor: tensor)
+    let channels: Int
+    switch format {
+    case .rgb: channels = 3
+    case .grayscale: channels = 1
+    default:
+        print("\(format) is not supported yet.")
+        exit(-1)
+    }
+    let reshapedTensor = tensor.reshaped(to: [shape.0, shape.1, channels])
+    let image = Image(tensor: reshapedTensor)
     let resizedImage = size != nil ? image.resized(to: (size!.0, size!.1)) : image
     let outputURL = URL(fileURLWithPath: "\(directory)\(name).jpg")
     resizedImage.save(to: outputURL, format: format, quality: quality)
--- a/Support/Image.swift
+++ b/Support/Image.swift
@@ -44,8 +44,8 @@ public struct Image {
     }
     
     public init(jpeg url: URL, byteOrdering: ByteOrdering = .rgb) {
-        let loadedData = _Raw.readFile(filename: StringTensor(url.absoluteString))
-        let loadedJpeg = _Raw.decodeJpeg(contents: loadedData, channels: 3, dctMethod: "")
+        let loadedFile = _Raw.readFile(filename: StringTensor(url.absoluteString))
+        let loadedJpeg = _Raw.decodeJpeg(contents: loadedFile, channels: 3, dctMethod: "")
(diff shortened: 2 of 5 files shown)


Support/Image.swift
```diff
@@ -20,29 +20,29 @@ public struct Image {
         case bgr
         case rgb
     }
-
+
```
BradLarson: There are a lot of odd whitespace edits in this file. Were those inserted by another tool? I don't think we need all these spaces on empty lines, so these could probably be removed.

Support/Image.swift
```diff
@@ -53,8 +53,8 @@ public struct Image {
             self.imageData = .uint8(data: loadedJpeg)
         }
     }
-
-    public func save(to url: URL, format: _Raw.Format = .grayscale, quality: Int64 = 95) {
+    
+    public func save(to url: URL, format: _Raw.Format = .rgb, quality: Int64 = 95) {
```
BradLarson: Switching the default to .rgb here will potentially cause problems with some of the examples, because they were written with the assumption of grayscale as the default parameter. If the default is switched here, you may need to also change all examples that use this to reflect the new default.

Support/Image.swift
```diff
@@ -77,35 +77,35 @@ public struct Image {
...
+                    size: Tensor<Int32>([Int32(size.0), Int32(size.1)])).squeezingShape(at: 0))
         }
 
     }
 }
 
-public func saveImage(_ tensor: Tensor<Float>, size: (Int, Int), directory: String, name: String)
-    throws
-{
+public func saveImage(_ tensor: Tensor<Float>, size: (Int?, Int?) = (nil, nil),
```
BradLarson: The way that the resizing logic is currently set up here is that resizing is only performed if both width and height are specified. In that case, it might be cleaner to use (Int, Int)? instead of separate optionals.

Also, the formatting of these lines doesn't seem to line up with the Google Style Guide. Would you be able to re-run swift-format against this using the configuration file in the swift-models repository?

rahulbhalley: Apologies but I'm not sure how to run swift-format. Building it doesn't provide me with a cmd tool for usage. 

I cloned it on ~/Desktop and ran `swift build` inside `swift-format` directory & it build successfully. Then if I run `swift-format` from any location in cmd it produces an error: `zsh: command not found: swift-format`.

It'd be really helpful if you'd direct me to an example installation and usage. 

Btw I've updated `size` argument type.

Support/Image.swift
rahulbhalley: Please note that all the tests were passed with this new default behavior. So does this still mean that some examples might experience image loading failure? Let me know so that I can prioritize on looking into it.

rahulbhalley:
Btw can somebody please explain me if `_Raw` enumeration is meant for exposing all TensorFlow C++ APIs then why some APIs are not available like `_Raw.rGBToGrayscale()` or `_Raw.GrayscaleToRGB()`? 

More specifically how can I access these functions?

@dan-zheng

Support/Image.swift
BradLarson: The GAN, DCGAN, and Autoencoder examples use this saving behavior at runtime, not during unit tests, so they need to be manually run to observe their behavior. I'd make sure all three of these still run properly with this change.

dan-zheng:
> Btw can somebody please explain me if `_Raw` enumeration is meant for exposing all TensorFlow C++ APIs then why some APIs are not available like `_Raw.rGBToGrayscale()` or `_Raw.GrayscaleToRGB()`?
> 
> More specifically how can I access these functions?

It actually is not the case that all TensorFlow C++ ops are imported into the `TensorFlow._Raw` enum.

[Swift TensorFlow bindings](https://github.com/tensorflow/swift-apis/tree/master/Sources/TensorFlow/Bindings) (the `_Raw` enum) are generated by [the `generate_wrappers.py` script](https://github.com/tensorflow/swift-apis/blob/f78b8808e1e84e552326502de30579b08c528aef/Sources/TensorFlow/Bindings/generate_wrappers.py). The script has [mappings for supported argument/attribute types](https://github.com/tensorflow/swift-apis/blob/f78b8808e1e84e552326502de30579b08c528aef/Sources/TensorFlow/Bindings/generate_wrappers.py#L107-L132) - if a TensorFlow op has arguments or attributes with unsupported types, no Swift binding is created for it.

That said, I didn't find TensorFlow C++ ops for rgb <-> grayscale conversion. I did find [`tf.image.rgb_to_grayscale`](https://www.tensorflow.org/api_docs/python/tf/image/rgb_to_grayscale), which appears to be a non-primitive function [implemented in Python](https://github.com/tensorflow/tensorflow/blob/v2.1.0/tensorflow/python/ops/image_ops_impl.py#L1859-L1886).

rahulbhalley:
Sorry for dirty commit message. 🙄

Introduced `shape` argument to reshape image tensor, `size` argument spatially resizes the image, and `channels` property for passing appropriate number of channels based on `format` (RGB: 3, grayscale: 1) to `reshaped(to:)` method.

rahulbhalley:
Sorry for dirty commit message. 😅

Introduced `shape` argument to reshape image tensor, `size` argument spatially resizes the image, and `channels` property for passing appropriate number of channels based on `format` (RGB: 3, grayscale: 1) to `reshaped(to:)` method.

Support/Image.swift
rahulbhalley: Done. ✅

DCGAN/main.swift
```diff
@@ -165,5 +165,5 @@ for epoch in 0 ... epochs {
 let noise1 = Tensor<Float>(randomNormal: TensorShape(1, 100))
 let generatedImage = generator(noise1)
 try saveImage(
-        generatedImage, size: (28, 28), directory: outputFolder,
-        name: "final.jpg")
+    generatedImage, shape: (28, 28), format: .grayscale, directory: outputFolder,
+    name: "final")
```
rahulbhalley: Removed .jpg because `saveImage(_:shape:)` already appends it.

Support/Image.swift
```diff
@@ -88,24 +88,33 @@ public struct Image {
         case let .uint8(data):
             return Image(
                 tensor: _Raw.resizeBilinear(
-                    images: Tensor<UInt8>([data]),
-                    size: Tensor<Int32>([Int32(size.0), Int32(size.1)])))
+                    images: Tensor<UInt8>(data.rankLifted()),
+                    size: Tensor<Int32>([Int32(size.0), Int32(size.1)])).squeezingShape(at: 0))
```
BradLarson: Talking with Sylvain, he believes that `squeezingShape(at:)` will lead to incorrect behavior here. You state that this was added for legibility, but it adds code here over what was present before. If there isn't another reason for its presence, I think we'd prefer to not have this change.

Likewise for `rankLifted()`, which I don't believe is as concise as the previous `[data]`. If clarity is the desire, the previous argument probably could remain.

rahulbhalley: Ok. I've removed the suggested operations.

rahulbhalley:
> That said, I didn't find TensorFlow C++ ops for rgb <-> grayscale conversion. I did find [`tf.image.rgb_to_grayscale`](https://www.tensorflow.org/api_docs/python/tf/image/rgb_to_grayscale), which appears to be a non-primitive function [implemented in Python](https://github.com/tensorflow/tensorflow/blob/v2.1.0/tensorflow/python/ops/image_ops_impl.py#L1859-L1886). We could port the implementation to Swift.

@dan-zheng I believe importing such functions' implementation would be helpful in situation when user wants to load or save an image with certain image format, for instance saving an RGB image in grayscale format. 

This holds true for loading image datasets also. Maybe someone would want to load RGB dataset in grayscale format if the aim is to train an image colorization neural network. But I'm not sure about any use case of loading grayscale dataset with RGB channels. I think it'll make dataset/image loading/saving look more flexible. 

I would like to know others opinion on grayscale to RGB. But surely I'd love RGB to grayscale implementation.

rahulbhalley:
@BradLarson can we proceed?

dan-zheng:
> > That said, I didn't find TensorFlow C++ ops for rgb <-> grayscale conversion. I did find [`tf.image.rgb_to_grayscale`](https://www.tensorflow.org/api_docs/python/tf/image/rgb_to_grayscale), which appears to be a non-primitive function [implemented in Python](https://github.com/tensorflow/tensorflow/blob/v2.1.0/tensorflow/python/ops/image_ops_impl.py#L1859-L1886). We could port the implementation to Swift.
> 
> @dan-zheng I believe importing such functions' implementation would be helpful in situation when user wants to load or save an image with certain image format, for instance saving an RGB image in grayscale format.

There's a question whether such domain-specific (color conversion) functionality should live in [tensorflow/swift-apis](https://github.com/tensorflow/swift-apis) or elsewhere, like [tensorflow/swift-models](https://github.com/tensorflow/swift-models).

I think we should be wary of adding too much to the core `Tensor` type [tensorflow/swift-apis](https://github.com/tensorflow/swift-apis) to keep the API surface area small (core math-y and shape-y and indexing-y things only). Domain-specific extensions can always live elsewhere.

PR #328, opened 2020-02-19
Minor rework of BERT downloading code to use existing download(from:to:) function

BradLarson:
This replaces some of the bespoke downloading code in the BERT model with the existing `download(from:to:)` function. This centralizes some of the code, and aids with current issues around URLSession and Windows.

Because the BERT downloading code doesn't assume that the destination file name will match the source, `download(from:to:)` needed to be extended to detect whether the destination was a directory or a specific file. If the former, place the downloaded files into the directory with that name (creating it if necessary), with the downloaded files maintaining their original name. If a file is provided as the target, rename the downloaded files to that (creating the parent directory if needed).


brettkoonce:
i really like the pattern of bringing code in then refactoring it in place, fwiw.

PR #324, opened 2020-02-15
build: add `Transfomer` model to CMake based build

compnerd:
This enables building the Transformer model with CMake, enabling it to
be built on platforms where swift-package-manager is not yet viable.

PR #329, opened 2020-02-19
[AutoDiff] Fix Transformer CMake build linker error.

dan-zheng:
Fix Transformer linker error due to original/derivative declaration access level
mismatch: derivative is `@usableFromInline` but original is not.

Issue tracked at TF-1160.

Robust fix is TF-1099: require `@derivative` declarations to have same access
level as original declaration.

---

Unblocks https://github.com/tensorflow/swift-models/pull/324: add Transformer to CMake build.


dan-zheng:
Merging to unblock progress!

PR #330, opened 2020-02-19
tweak boston data getter to work on linux

brettkoonce:
this makes this work on my system, but is probably not the right way to solve this... 🤫

### DIFF
--- a/Datasets/BostonHousing/BostonHousing.swift
+++ b/Datasets/BostonHousing/BostonHousing.swift
@@ -40,7 +40,7 @@ public struct BostonHousing {
             remoteRoot: remoteURL, localStorageDirectory: localURL,
             extract: false)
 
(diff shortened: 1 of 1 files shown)


BradLarson:
I think the problem might be a little further up the file, in [line 37](https://github.com/tensorflow/swift-models/pull/330/files#diff-3e9044893e20378740b39bb14b9e6d6cR37). I recently changed the `download(from:to)` function to detect whether the target was a directory or file, and you need to manually annotate directories for them to be registered as such.

In line 37, I think we need to add `isDirectory: true`, like is done [here](https://github.com/tensorflow/swift-models/blob/master/Datasets/MNIST/MNIST.swift#L36).

Sorry I missed this during the various merges today.

brettkoonce:
@BradLarson that looks better, let me know!

PR #331, opened 2020-02-20
Transformer: replace Python with Swift objects

texasmichelle:
Remove python imports from Transformer.
Remove encoder.py file.
Add GPT-2 functions to BytePairEncoding.
Lint.


texasmichelle:
With this implementation, I'm seeing output like this, which looks like a problem with my byte mapping:
```
$ swift run -c release Transformer 0.5 "Introducing Swift for TensorFlow"
Introducing Swift for TensorFlow izf Ft Tf Ft Ft fcj izf Ft Ft fcj izf fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj fcj f
```

texasmichelle:
Incorporated latest refactoring from master and output looks better:
```
$ swift run -c release TransformerDemo 0.5 "Introducing Swift for TensorFlow"
Introducing Swift for TensorFlow

The following are the main features of Swift for!

A built-in Swift compiler for macOS and iOS

A Swift compiler for Swift programming languages
```

Models/Text/BytePairEncoding.swift
```diff
@@ -82,13 +133,26 @@ extension BytePairEncoder {
...
     }()
 
+    /// Regular expression matching the OpenAI GPT-2 implementation.
+    internal static let gpt2Glossary: [String] = [
+        "'s", "'t", "'re", "'ve", "'m", "'ll", "'d", " ?\\p{L}+", " ?\\p{N}+",
+        " ?[^\\s\\p{L}\\p{N}]+", "\\s+(?!\\S)", "\\s+",
+    ]
+
+    internal static let gpt2GlossaryRegex: NSRegularExpression = {
+        let escapedGlossary = gpt2Glossary.map { $0 }.joined(separator: "|")
```
saeta:
```suggestion
        let escapedGlossary = gpt2Glossary.joined(separator: "|")
```

Do you mean this?

Models/Text/BytePairEncoding.swift
```diff
@@ -82,13 +133,26 @@ extension BytePairEncoder {
...
+        "e.g", "i.e", "&amp;", "&#124;", "&lt;", "&gt;", "&apos;", "&quot;", "&#91;", "&#93;",
+    ]
 
     internal static let defaultGlossaryRegex: NSRegularExpression = {
         let escapedGlossary = defaultGlossary.map { "\\Q\($0)\\E" }.joined(separator: "|")
         return try! NSRegularExpression(pattern: "(?:\(escapedGlossary))|(?!\(escapedGlossary))")
     }()
 
+    /// Regular expression matching the OpenAI GPT-2 implementation.
+    internal static let gpt2Glossary: [String] = [
```
saeta:
```suggestion
    internal static let gpt2Glossary = [
```

Does the type not get inferred automatically?

Models/Text/BytePairEncoding.swift
```diff
@@ -60,14 +59,66 @@ public struct BytePairEncoder {
...
+    ///   - token: Token to encode.
+    /// - Returns: Array containing the BPE-coded tokens.
+    public func encodeGpt2(token: String) -> [String] {
+        // Split into parts before encoding.
+        var parts = BytePairEncoder.splitWithDelimitersGpt2(
+            token: token,
+            glossaryRegex: BytePairEncoder.gpt2GlossaryRegex)
+        if parts.count < 2 {
+            // Encode the full token and return.
+            return parts.map({ BytePairEncoder.encodedToken($0) })
```
saeta:
```suggestion
            return parts.map { BytePairEncoder.encodedToken($0) }
```
Does the suggestion compile? If so, is that preferable?

texasmichelle:
Those a great style pointers, thanks! They all made sense.

compnerd:
Didn't look at the diff, but, this definitely resolved my issues with the decoding on Windows.

Models/Text/BytePairEncoding.swift
```diff
@@ -60,14 +59,66 @@ public struct BytePairEncoder {
...
         // if useCache { cache[token] = encoded }
         return encoded
     }
+
+    /// Encodes the provided token to a sequence of BPE-coded tokens.
+    ///
+    /// - Parameters:
+    ///   - token: Token to encode.
+    /// - Returns: Array containing the BPE-coded tokens.
+    public func encodeGpt2(token: String) -> [String] {
```
dan-zheng: From the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#name-according-to-side-effects):
> #### Name functions and methods according to their side-effects.
> * Those without side-effects should read as noun phrases, e.g. `x.distance(to: y)`, `i.successor()`.
> * Those with side-effects should read as imperative verb phrases, e.g., `print(x)`, `x.sort()`, `x.append(y)`.

`encodeGpt2` is not a mutating function, so the name should be a noun phrase.
How about `func encoded`?

```swift
    /// Returns the byte pair encoding of `word`.
    /// - Parameters:
    ///   - word: The English word to encode.
    /// - Returns: The byte pair encoded tokens of `word`.
    public func encoded(_ word: String) -> [String] { ... }
```

I'm not sure what the parameter represents (an English word, or a token of some other encoding?), so I'm not the parameter documentation is accurate.

The alternative name `func bytePairEncoded` unnecessarily repeats type information (from `BytePairEncoder`), so I think `func encoded` is better.

---

Also from the Swift API Design Guidelines: ["clarity at the point of use is your most important goal"](https://swift.org/documentation/api-design-guidelines/#clarity-at-the-point-of-use).

Let's look at an example usage:

```swift
let bpeTokens = bytePairEncoder.encoded(word)
```
I think this is quite clear!

Models/Text/BytePairEncoding.swift
```diff
@@ -60,14 +59,66 @@ public struct BytePairEncoder {
...
+
+        return encodedTokens
+    }
+
+    /// Decodes the provided BPE-coded token to a sequence of tokens.
+    ///
+    /// - Parameters:
+    ///   - token: BPE-coded token to decode.
+    /// - Returns: Array containing the decoded tokens.
+    public func decode(token: String) -> String {
```
dan-zheng: `decode(token:)` is not a mutating function, so the name should be a noun phrase.

How about `func decoded`?
```swift
    /// Returns the decoded word for the given byte pair encoded token.
    /// - Parameters:
    ///   - word: The byte pair encoded token.
    /// - Returns: The decoded word.
    public func decoded(_ encodedToken: String) -> [String] { ... }
```

Models/Text/BytePairEncoding.swift
dan-zheng: I noticed that `func encode` already exists, in addition to `func encodeGpt2`.
Do these methods use two different encoding strategies?

If so, I think an idiomatic design is to define a `Encoding` enum in `BytePairEncoder`:
```swift
extension BytePairEncoder {
    enum Encoding {
        case gpt2
        case default // I'm not sure what this case represents. Perhaps there's a more descriptive name.
    }
}
```

And to add an optional `Encoding` argument to `func encoded`:
```swift
    public func encoded(_ word: String, encoding: Encoding? = .default) -> [String] { ... }
```

Models/Text/BytePairEncoding.swift
```diff
@@ -42,9 +43,7 @@ public struct BytePairEncoder {
     public func encode(token: String) -> [String] {
```
dan-zheng: I wonder why `encode` returns `[String]` instead of `String`?
Why does `encode` has type `(String) -> [String]` while `decode` has type `(String) -> String`?

Models/Text/BytePairEncoding.swift
```diff
@@ -144,6 +211,29 @@ extension BytePairEncoder {
         return parts
     }
 
+    /// Uses the given regex to split a token into individual glossary terms.
+    ///
+    /// - Parameters:
+    ///   - token: Full text.
+    ///   - glossaryRegex: Regular expression for segmenting the given token.
+    /// - Returns: Array of substrings that match the given regex.
+    internal static func splitWithDelimitersGpt2(
```
dan-zheng: Two suggestions:
- The method name should form a noun phrase, since the method is non-side-effectful.
- Abbreviations like `GPT2` or [`JSON`](https://developer.apple.com/documentation/foundation/jsonencoder) should be fully capitalized, unless they appear as the first word in a camelcase word (e.g. `gpt2XXX` or `jsonEncoder`).
```suggestion
    internal static func splittingWithDelimitersGPT2(
```

Models/Text/BytePairEncoding.swift
dan-zheng: Interestingly, I noticed that `JSONEncoder` defines [`func encode<T>(_: T) throws -> Data`](https://developer.apple.com/documentation/foundation/jsonencoder/2895034-encode). The method uses an imperative verb name even though it's semantically functional.

[`JSONEncoder`](https://developer.apple.com/documentation/foundation/jsonencoder) is defined in Foundation though, not the standard library. So perhaps it doesn't adhere as tightly to the API Design Guidelines.

dan-zheng: Update: I'm re-evaluating this naming suggestion. I think there's more room for improvement - I'll reply with an update after I look at the code more closely.

dan-zheng: Here's the updated feedback. Feel free to ignore other comments!
* Does `BytePairEncoder` contain any state (e.g. instance variables), or does it only use static mapping variables? If it's the latter (no state), it seems to me that `BytePairEncoder` may be most naturally represented using simple top-level functions: `func bytePairEncoded(_:)` and `func bytePairDecoded(_:)`. There's no reason to create a `BytePairEncoder` struct.
  * This feedback is similar to https://github.com/tensorflow/swift-models/pull/230#pullrequestreview-322493790: replace `PlainTextReporter` and `JsonReporter` types with a `func print(using style: PrintingStyle)` top-level function.
* If you'd like to make a less drastic change than deleting `BytePairEncoder`, I think the best method names are just `func encode(_:)` and `func decode(_:)`, not `func encoded(_:)` and `func decoded(_:)`. `encode(_:)` and `decode(_:)` have precedent in Swift, e.g. [`JSONEncoder.encode(_:)`](https://developer.apple.com/documentation/foundation/jsonencoder/2895034-encode) and [`UTF16.encode(_:)`](https://developer.apple.com/documentation/swift/unicode/utf16/2905288-encode).

Let me know if this makes sense! Sorry for the review thrash.

texasmichelle: `BytePairEncoder` contains state, so leaving it as a struct makes sense. The `decode` function does not, so I made it static.
The enum option is a good idea, so I added a Variant enum and combined 2 sets of functions.
Thanks for the detailed links and reasoning behind your suggestions! Because of precedent, I left `encode` and `decode` as-is.

Models/Text/BytePairEncoding.swift
texasmichelle: That's good clarification about acronym capitalization, thanks!! Since I combined this function with an enum, I renamed it to `splittingWithDelimiters`.

Models/Text/BytePairEncoding.swift
texasmichelle: That's a good catch. I'll think about whether it makes sense for decode to return `[String]`.

texasmichelle: The purpose of `encode` is for tokenization, which turns a single string into an array of tokens. `decode` is used to turn an encoded token into an unencoded string. Each of the encoded tokens is captured individually as a logits slice. This happens in a loop and after decoding, they are all concatenated together to form the result text, so I think it makes sense for decode to accept and return a singular string.

PR #334, opened 2020-02-20
Add support for pretrained VGG16

Shashi456:
The first of PRs for addition of Pretrained Image classification models. `extract` added because it gives me a linker error otherwise.
Is this way of overloading `VGG16` struct to add a `load` function an okay way to go forward with pretrained models?
I've tested the loading of weights and it does work. Build passes locally.

I can also add VGG19 pretrained in the same PR, but I have just added this initially for ease of review.


8bitmp3:
This is cool. Getting a pretrained net eg for transfer learning without Python/shell`wget` and not having to define a base VGG architecture would be welcome. Less lines of code 👍

8bitmp3:
@shashi456 @bradlarson and team. I suggest maybe you can make this PR about more than one pretrained net. There’s a model tensorflow zoo here that is still a gem 💎 here https://github.com/tensorflow/models/tree/master/research/slim#pre-trained-models (TF Slim FTW!)

brettkoonce:
🥳🎉🎁💪🚀!!!

Models/ImageClassification/VGG.swift
```diff
@@ -95,3 +97,74 @@ public struct VGG19: Layer {
         return backbone.sequenced(through: flatten, dense1, dense2, output)
     }
 }
+
+public func extract(tarGZippedFileAt source: URL, to destination: URL) throws {
```
BradLarson: Rather than having file extraction and manual downloading code here, a more robust solution would be to add compressed file extraction to the CheckpointReader, which handles downloading checkpoints for you. This is something I haven't gotten to yet, but I can work on adding this. The file extraction code [here](https://github.com/tensorflow/swift-models/blob/master/Datasets/DatasetUtilities.swift#L109) will be lifted out of that file and into a more general extraction function for use in multiple places.

I'm working on other areas now, but I'll see about extending CheckpointReader with this functionality, which can then remove the need for this downloading and extraction code.

Models/ImageClassification/VGG.swift
```diff
@@ -95,3 +97,74 @@ public struct VGG19: Layer {
...
+    let process = Process()
+    process.environment = ProcessInfo.processInfo.environment
+    process.executableURL = URL(fileURLWithPath: "/bin/bash")
+    process.arguments = ["-c", "tar -C \(destination.path) -xzf \(source.path)"]
+    try process.run()
+    process.waitUntilExit()
+}
+
+extension VGG16 {
+    public func load(from directory:URL) throws -> VGG16 {
```
BradLarson: We'll need to call out that this is a checkpoint trained against ImageNet, provide a reference to the source, and supply whatever information is relevant about training conditions. `load(from:)` is probably too generic of a name, because that implies that this could be used to load from other sources, when this is hardcoded for this one specific checkpoint. Also, the `from` parameter here really seems to be a destination on the local filesystem, not the source of the checkpoint, which could be a little confusing.

Shashi456: @BradLarson What about `pretrained` and like a print statement at the start saying `Loading vgg16 trained on Imagenet`?

Models/ImageClassification/VGG.swift
Shashi456: But this workaround works for now right?

BradLarson: I don't think that this `extract(tarGZippedFileAt:to:)` free function belongs in the model file for the general VGG model, so I'd like to see if we can remove that in some way. I also have concerns that it may not work on all platforms as-is, which is why the `extractArchive(for:)` function is set up the way it is. I'd prefer to use the latter as the basis for a more general approach that will extend to other models.

If there's no rush on this, I believe that I can add this capability to the checkpoint reader soon, and then this will become much simpler.

Shashi456:
@BradLarson so the generalized `extractArchive` is currently blocking this PR right?

BradLarson:
I had some time to work on this, so PR #362 adds the ability for CheckpointReader to download and extract compressed checkpoint files. Once that's in, you'll be able to safely remove all of the extraction and downloading code here and replace it with:

```swift
let remoteCheckpoint = URL(
    string: "https://storage.googleapis.com/s4tf-hosted-binaries/checkpoints/VGG16/vgg_16.tar.gz")!
let checkpointReader = try CheckpointReader(checkpointLocation: remoteCheckpoint, modelName: "VGG")
```
I'm using a checkpoint we're hosting ourselves, because the VGG-16 checkpoint you were using was a TensorFlow v1 checkpoint and our reader only supports v2 checkpoints. I converted the file and re-uploaded to our hosted bucket, which is what I link above. I didn't realize the old tf.slim hosted checkpoints from early 2017 and before were v1 checkpoints, so we'll have to convert those if we need others. VGG-16 isn't used much as a model with modern TensorFlow, so I couldn't find newer checkpoints to draw from.

I'm still not a huge fan of the `pretrained(from:)` naming, because this method performs an action and that isn't strongly expressed in the name. Perhaps `loadPretrainedModel(from:)`, `loadPretrainedWeights(from:)`?

Ideally, I'd like this to be specified as per PyTorch and others, where `pretrained: Bool = false` is an option on init, loading the pretrained weights if true. I'd like to come up with a more consistent design for initializing layers from loaded weights, so that layers aren't created with random weights first and then have those weights blown away by loaded ones. The Transformer model takes a reader as an initialization parameter, and I like the way that's handled.

Anyway, the renamed extension method is probably fine for now, once the downloading code is replaced, but it's something to think about.

BradLarson:
PR #362 has been merged in, so you should be safe to rework this to remove the downloading code in favor of the CheckpointReader's new remote archive support, using the code I describe above.

Shashi456:
Will start adding a slew of models in the next few days, thanks a lot for helping with the addition @bradlarson.

Shashi456:
@bradlarson, could I ask you to add more models to the bucket? And how do you want to test this?

Shashi456:
@BradLarson Done. Would like to ask if you are happy with the code and naming of the function?

Models/ImageClassification/VGG.swift
```diff
@@ -1,97 +1,149 @@
...
+
+    @differentiable
+    public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+        let backbone = input.sequenced(through: layer1, layer2, layer3, layer4, layer5)
+        return backbone.sequenced(through: flatten, dense1, dense2, output)
+    }
+}
+
+extension VGG16 {
+    public func pretrained(from directory:URL) throws -> VGG16 {
```
BradLarson: This is set up as an instance method, so you have to allocate a VGG instance, call this on it, and get a second VGG instance in return. That's undesirable , so this either should be a class method, or should modify `self` and not return a new VGG instance. Also, `directory` isn't used for anything anymore, so that parameter can probably be removed if it's no longer needed.

If this is done as a class method, I'd name it `pretrainedModel()`. If done as an instance method, the fact that it mutates the model should be explicit in the name, so maybe `loadPretrainedModel()`.

Models/ImageClassification/VGG.swift
```diff
@@ -1,97 +1,149 @@
-// Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
BradLarson: Something looks like it got messed up in the formatting. Is there a way just to have the changes from the section at the bottom be applied, rather than changing the whole file?

Models/ImageClassification/VGG.swift
```diff
@@ -1,97 +1,149 @@
...
+        vgg16.dense1.weight = Tensor(checkpointReader.loadTensor(named: "vgg_16/fc6/weights"))
+        vgg16.dense1.bias = Tensor(checkpointReader.loadTensor(named: "vgg_16/fc6/weights"))
+        vgg16.dense2.weight = Tensor(checkpointReader.loadTensor(named: "vgg_16/fc7/weights"))
+        vgg16.dense2.bias = Tensor(checkpointReader.loadTensor(named: "vgg_16/fc7/weights"))
+        vgg16.output.weight = Tensor(checkpointReader.loadTensor(named: "vgg_16/fc8/weights"))
+        vgg16.output.bias = Tensor(checkpointReader.loadTensor(named: "vgg_16/fc8/weights"))
+        
+        return vgg16
+    }
+}
```
BradLarson: Minor formatting nit: there should be a newline at the end of this.

dabrahams:
team to have a discussion about the API details

texasmichelle:
Reminder for @dabrahams and @BradLarson to make some decisions on API design.

Shashi456:
Is there any interest in taking this forward?

dabrahams:
@Shashi456 Thanks for your patience.  We're definitely interested in supporting pretrained models and hope to get back to you soon about this PR in detail!

BradLarson:
I have to again apologize for the delay, I'm working on a larger reorganization of model checkpoint loading and saving as part of a process of migrating it from swift-models to swift-apis. This is taking longer than anticipated, but I'm still trying to get a clean layered API that we can use for pretrained models across the repository.

I hope to have something to show soon on this, and should start getting some of the first restructuring PRs in presently.

Shashi456:
is there any interest in taking this forward?

BradLarson:
@Shashi456 - If it's all right with you, my plan was to add checkpoint-specific name mapping to PR #631 (in addition to the extension-based name mapping that's already there) and then I could reimplement the relevant portions from here in that PR. That way, we'll be able to leverage that higher-level loading API for the VGG model as a first example of ImageNet checkpoint loading.

I was going to leave this PR open until I did so (which I should get to shortly), so that this wasn't forgotten. Does that sounds like a sensible plan?

BradLarson:
Unfortunately, due to the timing of the wind-down of the project we never were able to complete the rework to support this that I'd wanted. I'm sorry this hung out there as long as it did. Maybe if someone continues a fork of this repository work can proceed on this feature.

PR #338, opened 2020-02-21
Add Batcher examples

sgugger:
This PR adds one utility function (collectURLs) and a notebook showing how to assemble a standard ImageNet preprocessing pipeline (with random resized crop and random flip) on Imagenette using Batcher.

My next steps are: adding an example of pipeline assembling the data for a language model on Wikitext-103, then to add training loop examples to both those notebooks. I don't know if you want to gather all of this in one PR or merge as I progress.


Support/FileManagement.swift
```diff
@@ -35,3 +35,28 @@ public func download(from source: URL, to destinationDirectory: URL) throws {
     let downloadedFile = try Data(contentsOf: source)
     try downloadedFile.write(to: URL(fileURLWithPath: destinationFile))
 }
+
+/// Collect all filemanes URL under a folder `url`, potentially recursing throu all subfolders.
```
BradLarson:
```suggestion
/// Collect all file URLs under a folder `url`, potentially recursing through all subfolders.
```

Support/FileManagement.swift
```diff
@@ -35,3 +35,28 @@ public func download(from source: URL, to destinationDirectory: URL) throws {
...
+/// Optionally filters some extension (only jpeg or txt files for instance).
+///
+/// - Parameters:
+///   - url: The folder to explore.
+///   - recurse: Will explore all subfolders if set to `true`.
+///   - extensions: Only keeps URLs with extensions in that array if it's provided
+public func collectURLs(under url: URL, recurse: Bool = false, filtering extensions: [String]? = nil) -> [URL] {
```
BradLarson: Generally, I think we avoid type names as parameters in Swift, so rather than `url` as a parameter name here, I'd suggest `directory` instead.

Also, the lines look like they might have gotten a little long in a couple places here, so if you can get swift-format operational on your end it might be able to reformat these in this file.

Support/FileManagement.swift
```diff
@@ -35,3 +35,28 @@ public func download(from source: URL, to destinationDirectory: URL) throws {
...
+    var res: [URL] = []
```
BradLarson: Instead of `res` as a name, would `files` be more descriptive as to the contents of this variable?

Support/FileManagement.swift
sgugger: A google search gives me quite a few swift formatters, which one are you talking about? (We may want to point it out in a CONTRIBUTING.md btw.)

sgugger:
Ok, managed to get the swift formatter on that changed file. One question I had for the notebook was: what is the best way to install swift-models in the first cell? From the repo directly?
Also not sure of what the story is on colab, is swift-models installed there by default?

dan-zheng:
> One question I had for the notebook was: what is the best way to install swift-models in the first cell? From the repo directly?
> Also not sure of what the story is on colab, is swift-models installed there by default?

`swift-models` is not currently installed on Colab by default, though installing it by default makes sense to me.

In the meantime, `swift-jupyter`'s [SwiftPM `%install` directives](https://github.com/google/swift-jupyter/tree/1182faaf90e6381110900af321590552f955926d#install-directives) is the best way to install `swift-models`.

It seems you're already using this in [`Imagenette.ipynb`](https://github.com/tensorflow/swift-models/blob/ebaa23f73dd51e8d81aac028310481420976bc25/Examples/Notebooks/Imagenette.ipynb):
```swift
%install-location $cwd/swift-install
%install '.package(path: "~/git/swift-models")' Batcher ModelSupport Datasets
```

But you should change the package location to a GitHub URL:

```swift
%install-location $cwd/swift-install
%install-swiftpm-flags -c release
%install '.package(url: "https://github.com/tensorflow/swift-models", .branch("master"))' Batcher ModelSupport Datasets
```

Check out the [I/O '19 MiniGo demo](https://github.com/tensorflow/swift/blob/io19/demos/demo-4-minigo.ipynb) as an example.

8bitmp3:
(I can confirm that in Colab running a variant of this one line in the first cell does the job e.g. as Dan mentioned
‘%install '.package(url: "https://github.com/tensorflow/swift-models", .branch("master"))' Batcher ModelSupport Datasets’ 

Thanks for letting us know @dan-zheng why swift-models isn’t in the toolchain by default,  I had a discussion with @saeta about this thing and was gonna ask on Swift@tensorflow.)

sgugger:
Thanks Dan, updated the installation cell.

sgugger:
Added a new notebook for the input pipeline to gather the data for a Language model dataset. This requires some additional functionality I put in a new swift file.

sgugger:
Test seems to fail because of the CMake build, but the file looks okay to me. Not sure what's wrong there.

Datasets/LabeledExample.swift
```diff
@@ -34,3 +34,13 @@ public struct LabeledExample: TensorGroup, KeyPathIterable, Collatable {
         data = Tensor<Float>(handle: TensorHandle<Float>(handle: _handles[dataIndex]))
     }
 }
+
+public struct TensorPair<S1: TensorFlowScalar, S2: TensorFlowScalar>: Collatable, KeyPathIterable {
```
dan-zheng: The name `TensorPair` sounds like a generic "tuple of two tensors" type.
It should define with generic property names `first` and `second`:
```swift
public struct TensorPair<T: TensorFlowFloatingPoint>: Differentiable {
    var first: Tensor<T>
    var second: Tensor<T>

    @differentiable
    public init(_ first: Tensor<T>, _ second: Tensor<T>) {
        self.first = first
        self.second = second
    }
}
```

However, based on the usage of `TensorPair`, it seems like you want a type with semantics "labels":
```swift
    public subscript(index: Int) -> TensorPair<Int32, Int32> {
        // What does the pair represent?
        ...
    }
```

So instead, I think you should use a more specific name than `TensorPair`: it's a tuple of an input tensor and a "target" tensor.

Maybe `DatasetExample`? or `DatasetElement` ? If it's specific to `LanguageModelDataset`, making it a nested `struct Element` also makes sense.

Datasets/LabeledExample.swift
```diff
@@ -34,3 +34,13 @@ public struct LabeledExample: TensorGroup, KeyPathIterable, Collatable {
...
+    public var input: Tensor<S1>
+    public var target: Tensor<S2>
```
dan-zheng: Does `target` represent an "expected output", i.e. a label? How about `label` instead?
```suggestion
    public var label: Tensor<S2>
```

Datasets/LabeledExample.swift
dan-zheng: If nothing more, please add a doc comment to `TensorPair`!

Otherwise, folks browsing https://www.tensorflow.org/swift/api_docs will be confused why a `TensorPair` type (generic name) has `input` and `target` properties (specific names). 🙂

sgugger: Initially I used `TensorPair` because the current `LabeledExample` is not flexible enough and constrains the type of input/target. In general though, a `DatasetExample`/`DatasetElement`/`LabeledExample` could have more than one tensors as inputs (tabular models use one tensor for categorical variables and another one for continuous for instance) and more than one tensor as targets/labels (in object detection for instance, there is often one tensor for the bounding boxes and one for the labels). So I think we should either stick to:
- `TensorPair` with first and second attributes as you suggested
or
- Highlight that this is a basic dataset example in `DatasetExample`/`DatasetElement`/`LabeledExample` by adding `Basic` in the name (or something else).

Then the user clearly knows that they have to implement a more complex struct if their problem requires it. Since the latter results in longer names, I'd personally go for the first but I welcome any other thoughts on this.

dan-zheng: Actually, a specialized `TensorPair` type sounds good for now. We can defer further design.

Datasets/LabeledExample.swift
dan-zheng: Actually, `second` please.

Datasets/LanguageModelDataset.swift
```diff
@@ -0,0 +1,135 @@
...
+  //The number of batches
+  private var batchCount: Int
+  //The sequence length of the last batch
+  private var lastLength: Int
+  //Indices used to iterate through the dataset
+  public var indices: [Int]
+  //Cumulative lengths
+  private var cumLengths: [Int]
+  //To conform to Collection
+  public var startIndex: Int { return 0 }
```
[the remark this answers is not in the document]
sgugger: Done, let me know what you think.

Datasets/LabeledExample.swift
sgugger: Done and updated the notebooks accordingly.

Datasets/LanguageModelDataset.swift
```diff
@@ -0,0 +1,135 @@
...
+  //The length of a contiguous chunk of text
+  private var batchLength: Int
+  //The number of batches
+  private var batchCount: Int
+  //The sequence length of the last batch
+  private var lastLength: Int
+  //Indices used to iterate through the dataset
+  public var indices: [Int]
+  //Cumulative lengths
+  private var cumLengths: [Int]
```
texasmichelle: Please use the full word in this variable name or choose a different word.

sgugger: Done (and then some)

Datasets/LanguageModelDataset.swift
```diff
@@ -0,0 +1,135 @@
+import TensorFlow
+
+//Build a dataset suitable for language modeling from an array of texts
+public struct LanguageModelDataset<Item> {
```
texasmichelle: This naming is a bit confusing, since I expect it to be an analog to ImageClassificationDataset, a protocol.

texasmichelle: I also think this is problematic since it does not implement TensorGroup, which would make it usable with Dataset.

dan-zheng: I think `Batcher` is an alternative to `Dataset`. `TensorGroup` is a nasty thing that's necessary just for `Dataset`. So I understand `Batcher` as an alternative to `Dataset` and a step towards a full input pipeline story - when the story is done, the poor `TensorGroup` protocol and derived conformances can be removed.

texasmichelle: If `TensorPair` implements `TensorGroup`, this should work in the existing structure. I still think the naming of `LanguageModelDataset` is confusing, but it's probably ok to merge for now. Perhaps @BradLarson has an opinion since I think he created `ImageClassificationDataset`.

Datasets/LanguageModelDataset.swift
```diff
@@ -0,0 +1,135 @@
...
+  public typealias Index = Int
+  public typealias Element = TensorPair<Int32, Int32>
+
+  //A function that reads Item to get an array of Int
+  public let openItem: (Item) -> [Int]
```
dan-zheng: I think the name `openItem` has room for improvement.
I can't tell that it's the name of a function, and I don't know what an item is. What does the returned `[Int]` represent?

Datasets/LanguageModelDataset.swift
sgugger: IMHO `ImageClassificationDataset` is the thing that is badly named has it's just a wrapper around a training set and a validation set. It's not even a collection. It should be `Datasets` plural, as there is nothing specific to image classification in it.

Note that `Dataset` does not work with `LazyTensor` so can't be used on TPU. `TensorPair` purposely does not conform to `TensorGroup` as it's meant to be used with `Batcher` which is the proposed project to get rid of `Dataset` for now.

As for the name of `LanguageModelDataset`, I think it is accurate as: it is a collection, so `Dataset` is not misleading, and it prepares any array of texts for language modelling.

Yes, the whole dataset examples right now are a bit messy as there are two directions being explored in parallel. I think it's okay to have this exploration while we wait to settle on one design.

Datasets/LanguageModelDataset.swift
```diff
@@ -0,0 +1,135 @@
+import TensorFlow
+
+//Build a dataset suitable for language modeling from an array of texts
+public struct LanguageModelDataset<Item> {
+  public typealias Index = Int
+  public typealias Element = TensorPair<Int32, Int32>
```
dan-zheng: Please move `Index` and `Element` to the `Collection` conformance below, since they are `Collection` protocol requirements.

Datasets/LanguageModelDataset.swift
```diff
@@ -0,0 +1,135 @@
+import TensorFlow
+
+//Build a dataset suitable for language modeling from an array of texts
```
dan-zheng: Could you please use triple slashes for documentation comments?
Since types are nouns, the doc comment for a type should be a noun clause, not a verb clause (e.g. `Build`).

```suggestion
/// A language model dataset.
```

I found these [Swift API documentation comment guidelines](https://swift.org/documentation/api-design-guidelines/#write-doc-comment) to be very helpful. All declarations (especially public ones) can use a doc comment.

Datasets/LanguageModelDataset.swift
dan-zheng: `Item` as the generic parameter name is quite undescriptive: it's a general noun like "thing". Is there a more descriptive and accurate name that you could use?

Datasets/LanguageModelDataset.swift
sgugger: I had triple slashes but the swift-formatter removed them FYI. Will address.

Datasets/LanguageModelDataset.swift
sgugger: I was actually thinking of removing it entirely. It's to apply some function lazily on the items, and we already have `.lazy.map`. The only thing that will change is make the underlying items will be a `Collection` instead of an array (since `.lazy.map` does not return an array). 

The elements of that collection would be `[Int]` since we expect preprocessed texts (e.g. tokenized an numericalized) at this stage.

Datasets/LabeledExample.swift
```diff
@@ -34,3 +34,17 @@ public struct LabeledExample: TensorGroup, KeyPathIterable, Collatable {
         data = Tensor<Float>(handle: TensorHandle<Float>(handle: _handles[dataIndex]))
     }
 }
+
+// A simple tuple of Tensor, useful for many deep learning tasks (one tensor of inputs,
+// one tensor of labels). Note that if your task is more elaborate, you should write your
+// own struct of Tensors. Automatic conformance to Collatable can be derived the same way as
+// here.
+public struct TensorPair<S1: TensorFlowScalar, S2: TensorFlowScalar>: Collatable, KeyPathIterable {
```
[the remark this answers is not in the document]
sgugger: Sure, I can do this and declare the conformance of `LabeledExample` to `TensorGroup` in an extension. It will change all names used in swift-models though since `LabeledExample` uses the semantics `data` and `label` (not `first` and `second`), so I think a separate PR is best for this.

Datasets/LanguageModelDataset.swift
sgugger: This will disappear with the use of an underlying collection and the removal of `openItem`.

Datasets/LanguageModelDataset.swift
dan-zheng: Nice! I think code changes that lead to a simpler design (where declarations are easier to explain and document) are in the right direction!

Datasets/LanguageModelDataset.swift
dan-zheng: I see. To confirm: the `Item` generic parameter will be removed after design changes?

sgugger: Yes. But there will be a generic conforming to collection instead.

Datasets/LabeledExample.swift
```diff
@@ -34,3 +34,17 @@ public struct LabeledExample: TensorGroup, KeyPathIterable, Collatable {
         data = Tensor<Float>(handle: TensorHandle<Float>(handle: _handles[dataIndex]))
     }
 }
+
+// A simple tuple of Tensor, useful for many deep learning tasks (one tensor of inputs,
+// one tensor of labels). Note that if your task is more elaborate, you should write your
+// own struct of Tensors. Automatic conformance to Collatable can be derived the same way as
+// here.
```
dan-zheng: It is very important to use backticks around type names in documentation comments, so that the type names will be formatted correctly in [generated API documentation](https://www.tensorflow.org/swift/api_docs).

Please use `/// - Note:` for "note" documentation blurbs. This helps clearly separate main documentation (what is `TensorPair`?) from auxiliary documentation. They also show up nicely in generated API documentation.

```suggestion
/// A generic tuple of two tensors.
///
/// - Note: `TensorPair` has a generic name and provides little semantic information. Rather than using this type, consider creating a custom type with a more descriptive name.
```

sgugger: Will do, not sure there is a documentation for swift-models yet?

sgugger:
Ok, so merging this for now. Will make a follow-up PR tomorrow or Sunday to address all your comments (thanks a lot for taking the time to review thoroughly btw).

dan-zheng:
> Will make a follow-up PR tomorrow or Sunday to address all your comments (thanks a lot for taking the time to review thoroughly btw).

Sure thing! I think chatting about design "in-person" (via call) might be more productive, if we both have time!

Datasets/LabeledExample.swift
dan-zheng: Good point, `swift-models` doesn't have hosted API docs yet.

Datasets/LanguageModelDataset.swift
texasmichelle: OK, understanding them as two separate approaches helps clear up my misunderstanding. Agreed that settling on a design and documenting the direction will help with future efforts.

PR #340, opened 2020-02-21
Initial implementation of a TensorFlow v2 checkpoint writer

BradLarson:
This is an initial implementation of a TensorFlow v2 checkpoint writer. It has no dependencies on the TensorFlow runtime or libraries, and currently only depends on SwiftProtobuf (a dependency we can remove if needed by manually reading and writing the small metadata protobufs here).

I'm labeling this as an initial implementation, and can consider this a work-in-progress if others feel it is incomplete, because the checkpoints it generates currently cannot be read by Python TensorFlow. They can be read by our Swift-based checkpoint reader that is present alongside this in swift-models. The core deficiencies are a lack of correct CRC32C values in the metadata and an incomplete footer in the index header.

It has been validated on the Transformer model, and example tensor extraction code has been provided for that model in Transformer/CheckpointWriter.swift, alongside a commented-out sample of how to save and reload a checkpoint for Transformer in its main.swift file.


Support/Checkpoints/CheckpointWriter.swift
```diff
@@ -0,0 +1,80 @@
...
+            location: location, shard: shard, totalShards: numShards)
+
+        var outputBuffer = Data()
+        // TODO: Write this directly to disk, rather than accumulating it in memory.
+        for tensorName in tensorList {
+            guard let tensor = tensors[tensorName] else {
+                fatalError("Mismatch in sorted tensors at name: \(tensorName).")
+            }
+            let scalars = tensor.array.scalars
+            // scalars.withUnsafeBytes { pointer in
```
saeta: Tinkercruft?

Transformer/main.swift
```diff
@@ -67,3 +67,19 @@ for _ in 0..<100 {
...
+let writer = CheckpointWriter(tensors: tensors)
+try writer.write(to: temporaryDirectory, name: "model2.ckpt")
+
+let reader2 = try CheckpointReader(
+    checkpointLocation: temporaryDirectory.appendingPathComponent("model2.ckpt"),
+    modelName: "Transformer",
+    additionalFiles: [])
+
+let recreatedmodel = TransformerLM(reader: reader2, config: config, scope: "model")
+*/
```
saeta: Nit: trailing newline?

Support/Checkpoints/CheckpointWriter.swift
BradLarson: Oops, thought I got all those. Good catch.

Transformer/main.swift
BradLarson: Fixed, thanks.

BradLarson:
@shabalind - Good point, we should have tests for this. I've added a simple roundtrip test of writing tensors of various shapes to a checkpoint, then reading back the checkpoint, and finally verifying the tensors are as they should be. This uncovered an issue with how I was writing CRC bytes, so it's already paid off.

Are there additional tests you'd like to see for this initial implementation?

BradLarson:
With release branches about to be cut, I'll merge this in and continue work on further TF compatibility in a follow-on PR.

Issue #341, opened 2020-02-23
Imagenette dataset opening archive


8bitmp3:
Hey @BradLarson and @sgugger Thanks a lot for your efforts, really good stuff 👍.

I have a small issue with extracting the Imagenette archive, once it's been downloaded, before training MobileNetV1. I've tested this both in Xcode and Colab. See error messages snippets below to give you a better idea.

The downloaded dataset is likely stored in a folder - `Imagenette` - though in Colab it's not a folder but a file called `Imagenette` without a suffix under `/tmp/`. (We are looking for a file called `imagenette-320.tgz`.)

To give more background, in Colab, the libraries were imported by running this first (thanks @marcrasi 👏 for swift-jupyter and @saeta for suggesting to explore the repo): `%install '.package(url: "https://github.com/tensorflow/swift-models", .branch("master"))' ImageClassificationModels Datasets ModelSupport`.
(I also tried explicitly importing Batcher, just in case, both in Xcode and Colab).

In `Imagenette.swift` and this commit https://github.com/tensorflow/swift-models/commit/a1f5d08d09f9e41903b15d58a46af87cd9b3b21f#diff-b0b36780b8a7904e46299ea540e5a3d9, there're a few lines which may be worth looking at but I'm still learning Swift:
```swift
    let _ = DatasetUtilities.downloadResource(
         filename: "imagenette\(size.suffix)", fileExtension: "tgz",
         remoteRoot: location.deletingLastPathComponent(), localStorageDirectory: directory)
```

Colab's error message:
```
Fetching URL: https://s3.amazonaws.com/fast-ai-imageclas/imagenette-320.tgz...
Archive saved to: /tmp/Imagenette
Extracting archive...
tar (child): /tmp/Imagenette/imagenette-320.tgz: Cannot open: Not a directory
tar (child): Error is not recoverable: exiting now

gzip: stdin: unexpected end of file
/bin/tar: Child returned status 2
/bin/tar: Error is not recoverable: exiting now
Fatal error: Could not load Imagenette dataset: Error Domain=NSCocoaErrorDomain Code=260 "The file doesn’t exist.": file /tmp/tmpkvzfm0q9/swift-install/package/.build/checkouts/swift-models/Datasets/Imagenette/Imagenette.swift, line 217
```

Xcode's error message (very similar):
```
Loading resource: imagenette-320
File does not exist locally at expected path: /var/folders/......./Imagenette/imagenette-320 and must be fetched
[...]

PR #342, opened 2020-02-23
Minor correction to Imagenette Batcher downloading to indicate directory

BradLarson:
PR #328 changed the core downloading function to automatically detect if a download destination was a file or directory. In order for a directory to be detected, it must be labeled as such when constructed, otherwise it will be treated as a file.

This fixes the Batcher pathway in Imagenette to download to a directory named Imagenette, not a file by that name. This should address issue #341.


BradLarson:
(Verified fix in https://github.com/tensorflow/swift-models/issues/341#issuecomment-590097488 , so I merged this in.)

sgugger:
Yes, I spent an our debugging the same bug while building the notebook in #338, I should have thought of adding the fix in Datasets as well. Thanks for the fix!

PR #344, opened 2020-02-24
Implement the DLRM model

jordannad:


jordannad:
@googlebot I signed it!

Models/Recommendation/DLRM.swift
```diff
@@ -0,0 +1,110 @@
...
+
+    /// sparse represents the categorical inputs to the mini-batch.
+    ///
+    /// The array should be of length `numCategoricalInputs`.
+    /// Each tensor within the array should be a vector of length `batchSize`.
+    public let sparse: [Tensor<Int32>]
+}
+
+// Work-around for lack of inout support
+@differentiable(wrt: latentFactors, vjp: computeEmbeddingsVJP)
```
dan-zheng: Rather than using `@differentiable(vjp:)` for derivative registration, please use the `@derivative(of:)` attribute instead! I think `@derivative(of:)` is easier to understand: it marks a function as the _derivative of_ another one.
```swift
// The original function.
fileprivate func computeEmbeddings(
    sparseInputs: [Tensor<Int32>],
    latentFactors: [Embedding<Float>]
) -> [Tensor<Float>] { ... }

// The derivative function.
@derivative(of: computeEmbeddings)
fileprivate func computeEmbeddingsVJP(
    sparseInput: [Tensor<Int32>],
    latentFactors: [Embedding<Float>]
) -> (
    value: [Tensor<Float>],
    pullback: (Array<Tensor<Float>>.TangentVector) -> Array<Embedding<Float>>.TangentVector
) { ... }
```

`@derivative(of:)` is more flexible because it enables derivative registration without annotating the original function (which may live in a separate file). For example, `@derivative(of:)` allows us to register derivatives for imported C functions (like `sinf`), which is pretty neat.

`@differentiable(vjp:)` is deprecated and has already been removed in latest toolchains.

Tests/RecommendationModelTests/InferenceTests.swift
```diff
@@ -0,0 +1,94 @@
...
+            lnBot: bottomMLPSize,
+            lnTop: topMLPSize)
+
+        let result = model(Tensor(ones: [batchSize, nDense]),
+                           sparseInput: [Tensor([7, 3, 1, 3, 1, 6, 7, 8, 9, 2]),
+                                          Tensor([17, 13, 19, 0, 1, 6, 7, 8, 9, 10])])
+        XCTAssertEqual([batchSize], result.shape)
+    }
+
+    func testDLRMTraining() {
```
xihui-wu: Why have the testDLRMTraining in InferenceTests ? Better to rename the class or have them separate?

xihui-wu:
Thanks for contributing this @jordannad. Two general questions regarding the model as I looked into PyTorch implementations:
1. For MLP they have activations between the dense layers, any reason we omit it?
2. Looks like there are two other OPs besides concat, should we support ?

Models/Recommendation/DLRM.swift
```diff
@@ -0,0 +1,110 @@
...
+        // TODO: Dot interactions
+    }
+
+    @differentiable
+    public func callAsFunction(_ input: DLRMInput) -> Tensor<Float> {
+        callAsFunction(input.dense, sparseInput: input.sparse)
+    }
+
+    @differentiable(wrt: self)
+    public func callAsFunction(_ denseInput: Tensor<Float>, sparseInput: [Tensor<Int32>]) -> Tensor<Float> {
```
xihui-wu: Can you not omitting the label for denseInput?

rahulbhalley:
@jordannad can you please give a little description about what DLRM means?

jordannad:
Please take a look? I believe I've addressed the comments. Thank you!

Models/Recommendation/DLRM.swift
```diff
@@ -0,0 +1,197 @@
...
+        callAsFunction(denseInput: input.dense, sparseInput: input.sparse)
+    }
+
+    @differentiable(wrt: self)
+    public func callAsFunction(
+        denseInput: Tensor<Float>,
+        sparseInput: [Tensor<Int32>]
+    ) -> Tensor<Float> {
+        precondition(denseInput.shape.last! == nDense)
+        assert(sparseInput.count == latentFactors.count)
```
xihui-wu: use precondition as well ?

PR #345, opened 2020-02-25
Support: add a basic bi-directional mapping utility

compnerd:
This allows us to map from key or value to their respective inverses.
As this is a common operation, place this into the Support module.

### DIFF
--- a/Support/BijectiveDictionary.swift
+++ b/Support/BijectiveDictionary.swift
@@ -1,0 +1,94 @@
+// Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// A bijective mapping behaving similar to dictionary with the additional
+/// ability to lookup the key from the value
+public struct BijectiveDictionary<Key: Hashable, Value: Hashable> {
+  private var valueForKey: [Key:Value] = [:]
+  // TODO: specialize where value is Int32 such that keyForValue is actually an
+  // array?
+  private var keyForValue: [Value:Key] = [:]
+
+  /// The number of of key-value pairs in the dictionary
+  public var count: Int { valueForKey.count }
+
+  /// Initializes the dictionary from a sequence of tuples of key and value.
+  public init<T: Sequence>(_ mapping: T) where T.Element == (Key, Value) {
+    valueForKey = .init(uniqueKeysWithValues: mapping)
+    keyForValue = .init(uniqueKeysWithValues: mapping.lazy.map { ($1, $0) })
+  }
+
+  /// Initializes the dictionary from a dictionary
+  public init(_ dictionary: [Key: Value]) {
+    valueForKey = dictionary
+    keyForValue = .init(uniqueKeysWithValues: dictionary.lazy.map { ($1, $0) })
+  }
+
+  /// Reserves enough space to store the specified number of key-value pairs.
+  ///
+  /// If you are adding a known number of key-value pairs to a bijective
+  /// dictionary, use this method to avoid multiple reallocations.  This method
+  /// ensures that the dictionary has unique, mutable, contiguous storage, for
+  /// space allocated for at least the requested number of key-value pairs.
+  ///
+  /// - Parameter minimumCapacity: the requested number of key-value pairs to
+  ///   store.
+  public mutating func reserveCapacity(_ minimumCapacity: Int) {
+    valueForKey.reserveCapacity(minimumCapacity)
+    keyForValue.reserveCapacity(minimumCapacity)
+  }
+
+  /// Access the value associated with the given key for reading and writing.
+  ///
+  /// The subscript returns the value for the given key if the key is found in
+  /// the dictionary, or `nil` if the key is not found.
+  ///
+  /// The addition of the key-value pair requires that the mapping hold the
+  /// conditions of a bijection: the mapping must remain one-to-one and onto.
+  ///
+  /// - Parameter key: the key to find in the dictionary.
+  /// - Returns: the value associated with `key` if `key` is in the dictionary,
+  ///   or `nil` otherwise.
+  public subscript(key: Key) -> Value? {
+    get { return valueForKey[key] }
+    set(value) {
+      if let value = value {
+        if let k_0 = self.keyForValue.updateValue(key, forKey: value), k_0 != key {
+          self.valueForKey[k_0] = nil
+        }
+        if let v_0 = self.valueForKey.updateValue(value, forKey: key), v_0 != value {
+          self.keyForValue[v_0] = nil
+        }
+      } else {
+        if let oldValue = self.valueForKey[key] {
+          self.keyForValue[oldValue] = nil
+          self.valueForKey[key] = nil
+        }
+      }
+
+      assert(valueForKey.count == keyForValue.count,
+             "forward count: \(valueForKey.count), backward count: \(keyForValue.count), \(self)")
+    }
+  }
+
+  /// Retrieves the key associated with the given value if found in the
+  /// dictionary, or `nil` if the value is not found.
+  ///
+  /// - Parameter value: the value to find in the dictionary.
+  /// - Returns: the key associated with `value` if `value` is in the
+  ///   dictionary, or `nil` otherwise.
+  public func key(_ value: Value) -> Key? {
+    return keyForValue[value]
+  }
+}
--- a/Package.swift
+++ b/Package.swift
@@ -92,5 +92,6 @@ let package = Package(
         .testTarget(name: "CheckpointTests", dependencies: ["ModelSupport"]),
         .target(
             name: "BERT-CoLA", dependencies: ["TextModels", "Datasets"], path: "Examples/BERT-CoLA"),
+        .testTarget(name: "SupportTests", dependencies: ["ModelSupport"]),
     ]
 )
--- a/Support/CMakeLists.txt
+++ b/Support/CMakeLists.txt
@@ -1,4 +1,5 @@
 add_library(ModelSupport
+  BijectiveDictionary.swift
   Checkpoints/CheckpointIndexReader.swift
   Checkpoints/CheckpointReader.swift
   Checkpoints/Protobufs/tensor_bundle.pb.swift
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -3,6 +3,7 @@ add_subdirectory(DatasetsTests)
 add_subdirectory(FastStyleTransferTests)
 add_subdirectory(ImageClassificationTests)
 add_subdirectory(MiniGoTests)
+add_subdirectory(SupportTests)
 
 add_executable(ModelTests
   LinuxMain.swift)
@@ -12,6 +13,7 @@ target_link_libraries(ModelTests PRIVATE
   FastStyleTransferTests
   ImageClassificationTests
   MiniGoTests
+  SupportTests
   XCTest)
 
 add_test(NAME ModelTests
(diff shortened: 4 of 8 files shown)


Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
+
+public struct BiMap<Key: Hashable, Value: Hashable> {
```
dabrahams: - Should have a doc comment describing what this thing is and what its parameters signify.
- We'd have to meet an exceedingly high bar to justify the use of an abbrev. in this name, like “this is a term of art and people won't recognize it if we spelled it out.”
- Given that this mapping is symmetrical, I wonder if `Key` and `Value` are really appropriate names.

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+  public typealias Key = Key
```
dabrahams: Each member declaration (typealiases, properties, etc.), public or otherwise, should have a doc comment.  Only a sentence fragment is required; see/expand https://swift.org/documentation/api-design-guidelines/#write-doc-comment

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+  public typealias Value = Value
+
+  var keyed: [Key:Value] = [:]
+  // TODO: specialize where value is Int32 such that valued is actually an
+  // array?
+  var valued: [Value:Key] = [:]
+
+  public init(mapping: [Key:Value]) {
```
dabrahams: This initializer is what I'd call a [value preserving type conversion](https://swift.org/documentation/api-design-guidelines/#type-conversion) and therefore should not have an argument label.

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
+
+public struct BiMap<Key: Hashable, Value: Hashable> {
+  public typealias Key = Key
+  public typealias Value = Value
+
+  var keyed: [Key:Value] = [:]
+  // TODO: specialize where value is Int32 such that valued is actually an
+  // array?
+  var valued: [Value:Key] = [:]
```
dabrahams: These names are… not particularly Scottish ;-)

Guidelines say they should be noun phrases (e.g. `mapFromKeysToValues`), but it's a bit unwieldy in this case.  I might think of them as functions (e.g. `valueForKey` and `keyForValue`).

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+
+  public init(mapping: [Key:Value]) {
+    for (key, value) in mapping {
+      keyed[key] = value
+      valued[value] = key
+    }
```
dabrahams: Even in this naive implementation, I'd think you just want `keyed = mapping`.

Also, 

```swift
valued = .init(uniqueKeysWithValues: mapping.lazy.map { ($1, $0) })
```

is a more efficient way to build the reverse mapping and establishes the right invariant to boot.  

Per Sean Parent [directly](https://www.youtube.com/watch?v=W2tWOdzgXHA), or [via me](https://developer.apple.com/videos/play/wwdc2018/223/), "no raw loops".

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public mutating func reserveCapacity(_ minimumCapacity: Int) {
+    keyed.reserveCapacity(minimumCapacity)
+    valued.reserveCapacity(minimumCapacity)
+  }
+
+  public var count: Int {
+    assert(keyed.count == valued.count,
```
dabrahams: This doesn't seem like the most useful place for this assertion.  I would do it after every mutation.

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+           "keyed count: \(keyed.count), valued count: \(valued.count), \(self)")
+    return keyed.count
+  }
+
+  public subscript(key: Key) -> Value? {
```
dabrahams: These two subscripts will be ambiguous when `Key == Value`.  It seems to me at least one of them should have an explicit label.

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+    get { return keyed[key] }
+    set(value) {
+      guard let value = value else { return }
```
dabrahams: This line seems problematic.  In a regular dictionary `d[k] = nil` erases `k` from `d`.  Especially if you're going with no comments (for now), you'd better offer the analogous behavior!

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+      keyed[key] = value
+      valued[value] = key
+    }
+  }
+
+  public subscript(value: Value) -> Key? {
```
dabrahams: See comments for the previous subscript.

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+    get { return valued[value] }
+    set(key) {
+      guard let key = key else { return }
+      keyed[key] = value
+      valued[value] = key
+    }
+  }
+}
+
+extension BiMap: ExpressibleByDictionaryLiteral {
```
dabrahams: Unless you anticipate real code is going to derive serious benefit from this, and will suffer badly if people write `Bimap([…])` instead of just `[…]`, I would leave this out.  Literal convertibility is [generally overused](https://forums.swift.org/t/se-0270-review-2-add-collection-operations-on-noncontiguous-elements/31653/43) and literals are a huge source of type checker slowness.

Support/BiMap.swift
dabrahams: I would make it generic over any sequence of `(Key, Value)`.

Support/BiMap.swift
```diff
@@ -0,0 +1,55 @@
...
+  public init(dictionaryLiteral elements: (Self.Key, Self.Value)...) {
```
dabrahams: If the other intializer had been made generic and unlabeled you could have just written `self = .init(elements)`.

Support/BiMap.swift
compnerd: Personally, I really dislike `Key` and `Value` and used them because the originating idea was a dictionary.  Any suggestions on renaming them?

compnerd: As to the abbreviation: C++: `Boost.Bimap`, Java:`Guava.BiMap`, Rust: `bimap` as examples of prior art that justify the naming IMO.

dan-zheng: chiming in: I prefer `BidirectionalMap`. Or `Bimap`, rather than `BiMap`.

Support/BiMap.swift
compnerd: I don't think thats particularly terrible.  Removed.

Support/BiMap.swift
compnerd: The reverse subscript was removed in favour of `key(_:)`

Support/BiMap.swift
compnerd: Moved into the subscript operator.

Support/BiMap.swift
compnerd: Replaced in the new constructors.

Support/BiMap.swift
compnerd: Made it a generic constructor instead.

Support/BiMap.swift
compnerd: The typealiases are not necessary and have been removed.  Thanks @dan-zheng for pointing that out!

Support/BiMap.swift
compnerd: `Bimap` it is

dabrahams: Precedents from other languages are generally useful, especially when there's nothing else to guide us, but in this case they all conflict with the precedent we have in Swift.  In Swift, these things are not called maps, but dictionaries.  Yes, `BidirectionalDictionary` is long and slightly unwieldy, but it has no issues with comprehensibility (scratch that, I thought of one, see below) and is consistent with the design principle Saleem and I agreed on, which is “it's like Dictionary, but bidirectional,” with the eventual possibility of replicating all of Dictionary's API so it's a drop-in replacement.  The benefits of having a perfectly comprehensible name almost always outweigh the cost of a few more characters.  The execeptions are when the name itself is going to appear extremely often.  For type names, that's rarely the case, and I don't *think* it's the case here.

As for the comprehensibility issue, `Dictionary` is already bidirectional—it's a `BidirectionalCollection` (or should be).  This bidirectionality is something else.  I'd like to find a less technical word, but `BijectiveDictionary` seems like a good way to avoid that problem.

Support/Bimap.swift
```diff
@@ -0,0 +1,63 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// A bijective mapping behaving similar to dictionary with the additional
+/// ability to lookup the key from the value
+public struct Bimap<Key: Hashable, Value: Hashable> {
```
dabrahams: @compnerd please see https://github.com/tensorflow/swift-models/pull/345#discussion_r384814160 in case it's being hidden due to the "outdated" status.

Support/BiMap.swift
dabrahams: The comment applied to many lines in the file, so marking it resolved based just on the removal of the typealiases is a bit suspect.

Support/BiMap.swift
dabrahams: Once we settled on the "it's a bidirectional dictionary" story, other possibilities came to mind, e.g. `base` and `inverse`.  Not saying to change them again, but I thought it might be valuable to know about these ideas.

Support/Bimap.swift
compnerd: @dabrahams renamed to `BijectiveDictionary`.

Support/BiMap.swift
dan-zheng: `BidirectionalDictionary` is great!

Support/BijectiveDictionary.swift
```diff
@@ -0,0 +1,63 @@
...
+  public mutating func reserveCapacity(_ minimumCapacity: Int) {
+    valueForKey.reserveCapacity(minimumCapacity)
+    keyForValue.reserveCapacity(minimumCapacity)
+  }
+
+  public subscript(key: Key) -> Value? {
+    get { return valueForKey[key] }
+    set(value) {
+      if let value = value {
+        precondition(self.key(value) == nil,
```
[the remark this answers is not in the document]
dabrahams: Specifically, I'd have expected something like this:

```swift
public subscript(k: Key) -> Value? { // WARNING: UNTESTED!
  get { return valueForKey[k] }
  set {
    if let v0 = valueForKey.updateValue(forKey: k, newValue), v0 != newValue {
      keyForValue[v0] = nil
    }
    if let k0 = keyForValue.updateValue(forKey: newValue, k), k0 != k {
      valueForKey[k0] = nil
    }
   }
}
```
Incidentally, writing this code out makes me dislike the names we've currently got for the dictionary properties a lot.  Whatever names we choose should contribute to the readability of code like this.

Support/BijectiveDictionary.swift
```diff
@@ -0,0 +1,63 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// A bijective mapping behaving similar to dictionary with the additional
+/// ability to lookup the key from the value
+public struct BijectiveDictionary<Key: Hashable, Value: Hashable> {
+  var valueForKey: [Key:Value] = [:]
```
dabrahams: These `var`s should be private.

Support/BiMap.swift
dabrahams: @dan-zheng did you read my last paragraph?

Support/BijectiveDictionary.swift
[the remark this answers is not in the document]
compnerd: Sure, so, the reasoning for the behaviour (perhaps I mis-implemented it?) is that a bijection is both one-to-one and onto.  If the value being set for the key already exists, the bijection no longer holds as it would not be one-to-one.  If key already exists and we have broken the onto relationship, the post-check should break.  I suppose I can make that more clear with the precondition.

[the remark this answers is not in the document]
compnerd: Okay, thinking about this further, the conditions do actually hold with the original way that we were talking about, and I think I tripped myself up because of the test code that I am using.

texasmichelle:
Merging since this implementation is documented as one-to-one and onto, with test cases to verify. Thanks @compnerd !! 🎉

PR #346, opened 2020-02-25
Transformer: add a simple UI wrapper for GPT-2

compnerd:
This adds a simple UI wrapper for the transformer demo.

### DIFF
--- a/Transformer/TransformerLoop.swift
+++ b/Transformer/TransformerLoop.swift
@@ -1,0 +1,116 @@
+// Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import ModelSupport
+import TextModels
+import TensorFlow
+import Transformer
+
+internal class GPT2 {
+  private let checkpoint: URL =
+      URL(string: "https://storage.googleapis.com/gpt-2/models/117M/model.ckpt")!
+  private let auxiliary: [String] = [
+    "checkpoint",
+    "encoder.json",
+    "hparams.json",
+    "model.ckpt.meta",
+    "vocab.bpe",
+  ]
+
+  private let FS: FileManager = FileManager.default
+
+  enum GPT2Error: Error {
+  case invalidEncoding(id: Int32)
+  }
+
+  let storage: URL
+  let configuration: (file: URL, data: Data)
+  let encoder: (file: URL, data: Data)
+
+  let reader: CheckpointReader
+  let parameters: TransformerLMConfig
+  let model: TransformerLM
+  let bpe: BytePairEncoder
+  let mapping: BijectiveDictionary<String, Int32>
+
+  public var seed: Tensor<Int32>
+  public var temperature: Float = 1.0
+
+  var states: [AttentionContext]
+
+  public init() throws {
+    storage = FS.temporaryDirectory.appendingPathComponent("Transformer")
+
+    reader = try CheckpointReader(checkpointLocation: checkpoint,
+                                  modelName: "Transformer",
+                                  additionalFiles: auxiliary)
+
+    // Load model from configuration
+    let hparams_json: URL = storage.appendingPathComponent("hparams.json")
+    configuration = try (hparams_json, Data(contentsOf: hparams_json))
+
+    parameters = try JSONDecoder().decode(TransformerLMConfig.self,
+                                          from: configuration.data)
+    model = TransformerLM(reader: reader, config: parameters, scope: "model")
+
+    // Load existing token mappings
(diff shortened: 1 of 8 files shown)


Transformer/TransformerDemoUI.swift
```diff
@@ -0,0 +1,241 @@
+import WinSDK
```
texasmichelle: I'm getting a compiler error running from the repo root on ubuntu:

```
$ swift run -c release TransformerDemo 0.5 "Introducing Swift for TensorFlow"
.../compnerd/swift-models/Transformer/TransformerDemoUI.swift:1:8: error: no such module 'WinSDK'
import WinSDK
       ^
```

Is this missing from Package.swift or something I need to install in my environment?

Transformer/TransformerDemoUI.swift
```diff
@@ -0,0 +1,241 @@
...
+      self.gpt = GPT2()
+      self.buttonDelegate.gpt = self.gpt
+      self.label.text = "Transformer ready!"
+    }
+
+    return true
+  }
+}
+
+ApplicationMain(CommandLine.argc, CommandLine.unsafeArgv, nil, SwiftApplicationDelegate())
```
texasmichelle: I'm getting a compiler error running from the repo root on ubuntu:

```
$ swift run -c release TransformerDemo 0.5 "Introducing Swift for TensorFlow"
.../compnerd/swift-models/Transformer/TransformerDemoUI.swift:241:1: error: expressions are not allowed at the top level
ApplicationMain(CommandLine.argc, CommandLine.unsafeArgv, nil, SwiftApplicationDelegate())
^
```

Transformer/TransformerDemoUI.swift
```diff
@@ -0,0 +1,241 @@
...
+import SwiftWin32
+
+import Dispatch
+import Foundation
+import ModelSupport
+import TextModels
+import TensorFlow
+import Transformer
+
+internal extension BytePairEncoder {
```
texasmichelle: I think this is broadly applicable enough to be added to Models/Text/BytePairEncoding.swift. It could simplify some of the usage in Models/Text/BERT.swift.

Transformer/TransformerDemoUI.swift
compnerd: The dependency for that is something which is dependent on Windows itself.  I think we need to figure out how to exclude this from Package.swift by default.

Transformer/TransformerDemoUI.swift
compnerd: Ah, right, because it must be named `main.swift`.  Adjusted along with namespacing.

Transformer/TransformerDemoUI.swift
dan-zheng: > I think we need to figure out how to exclude this from Package.swift by default.

I wonder if this requires adding imperative code to `Package.swift`?

compnerd: I *think* that `.windows` was added as a platform, so it should be possible to at least mark the file to not be built.

Shashi456:
@compnerd could you put a screenshot in, excited to look at it xD.

Transformer/main.swift
```diff
@@ -12,97 +12,20 @@
...
-let start_token = tokenToID["<|endoftext|>"]!
-var tokens = Tensor(shape: [1, 1], scalars: [start_token])
-var temperature = Float(1.0)
+let gpt: GPT2 = GPT2()!
 
 // Set temperature.
-if CommandLine.arguments.count >= 2 {
-    temperature = Float(CommandLine.arguments[1])!
-}
+gpt.temperature = CommandLine.arguments.count > 2
```
texasmichelle: Should this be `>= 2`? It's possible to include a temperature without seed.

compnerd: Ugh ... clearly, I copy paste good :-(

Transformer/TransformerDemoUI.swift
compnerd: Moved.

Transformer/TransformerLoop.swift
```diff
@@ -0,0 +1,132 @@
...
+    seed = Tensor(shape: [1, 1], scalars: [mapping["<|endoftext|>"]!])
+    let empty: Tensor<Float> =
+        Tensor<Float>(zeros: [parameters.headCount, 0,
+                              parameters.embeddingSize / parameters.headCount])
+    states = (0 ..< parameters.layerCount).map { _ in
+      AttentionContext(key: empty, value: empty)
+    }
+  }
+
+  func tensor(for string: String) -> Tensor<Int32> {
```
compnerd: I think that this should be called `embedding`

Transformer/TransformerLoop.swift
```diff
@@ -0,0 +1,132 @@
...
+        result.slice(lowerBounds: [0, timesteps - 1, 0],
+                     upperBounds: [batchSize, timesteps, vocabularySize]) / temperature
+    seed = Tensor(randomCategorialLogits: logits.squeezingShape(at: 1),
+                  sampleCount: 1)
+
+    let ID: Int32 = Int32(seed[0][0])!
+    if let token: String = mapping.key(ID) {
+      return BytePairEncoder.decode(token: token)
+    }
+    print("ID \(ID) not found!")
```
compnerd: Since this is actually not supposed to happen, perhaps it makes sense to have this function be a `throws` function and return a `String` instead?

texasmichelle: SGTM

Transformer/TransformerLoop.swift
texasmichelle: yes, good call!

PR #347, opened 2020-02-25
models: support building against PythonKit

compnerd:
Enable use of PythonKit and Python, preferring PythonKit over Python.


brettkoonce:
@compnerd getting 
```
> swift run Gym-FrozenLake
/home/skoonce/swift/swift-models/Gym/FrozenLake/main.swift:18:12: error: no such module 'Python'
    import Python
           ^
```
now on linux with the latest s4tf master, should i be doing something differently?

> which swift
/home/skoonce/swift/swift-source/swift/swift-nightly-install/usr/bin/swift

saeta:
@brettkoonce : I think you probably need to switch to `import PythonKit`. If you're flipping back and forth between the last stable release, and a recent nightly, you can use the following idiom:

```swift
#if canImport(PythonKit)
  import PythonKit
#else
  import Python
#endif
```
Example: [PythonConversion.swift](https://github.com/tensorflow/swift-apis/blob/526cefd892ce183cba1d98aac2ba9e447e2156af/Sources/TensorFlow/Core/PythonConversion.swift#L17-L21)

brettkoonce:
@saeta a later checkout started working fine, just a one-off thing

PR #348, opened 2020-02-25
Add zero init for last BN and don't use bias

sgugger:
As @pschuh mentioned to me, ResNet V2 did not have the zero init for the last BatchNorm layers, so this PR addresses that.
Also, now that https://github.com/tensorflow/swift-apis/pull/674 has landed in the nightlies, we can set the flag to not use bias in all the convolutional layers of both ResNet.


sgugger:
If you mean between ResNet and ResNetV2 (and probably others) it is definitely possible to refactor a lot, and this is something I intend to work on when higher priority stuff is done.

Issue #285, opened 2020-01-23
Add option to not normalize CIFAR10 images


rickwierenga:
I noticed that https://github.com/tensorflow/swift-models/blob/master/Datasets/CIFAR10/CIFAR10.swift always normalizes images. I would like to contribute a flag where the user can set `normalizing: false`.

The new function declaration would be:

```swift
loadCIFARFile(named name: String, in directory: URL, normalizing: Bool = false)
```

We should use a default value for `normalizing` so it is compatible with previous versions.

sainikhit2k:
Hi ,  is this issue currently open ? If it is I would like to contribute to this issue

rickwierenga:
@sainikhit2k, I think @manideep2510 is working on this in https://github.com/tensorflow/swift-models/pull/349

BradLarson:
Fixed by PR #349.

PR #349, opened 2020-02-25
Added an optional argument to the function loadCIFARFile making normalization of CIFAR images optional

manideep2510:
Addressing the request raised in #285 for adding an option to not normalize CIFAR10 images.
Added an optional argument to the function loadCIFARFile making normalization of CIFAR images optional. The default is to normalize the images so that the function is compatible with the previous versions of the code.
I haven't seen a PR for this even though the issue was more than a month older, so I did it myself.
Please review and let me know.

### DIFF
--- a/Datasets/CIFAR10/CIFAR10.swift
+++ b/Datasets/CIFAR10/CIFAR10.swift
@@ -106,9 +106,9 @@ func loadCIFARFile(named name: String, in directory: URL, normalize: Bool = true
         
 }
 
-func loadCIFARTrainingFiles(localStorageDirectory: URL) -> LabeledExample {
+func loadCIFARTrainingFiles(localStorageDirectory: URL, normalize: Bool = true) -> LabeledExample {
     let data = (1..<6).map {
-        loadCIFARFile(named: "data_batch_\($0).bin", in: localStorageDirectory)
+        loadCIFARFile(named: "data_batch_\($0).bin", in: localStorageDirectory, normalize: normalize)
     }
     return LabeledExample(
         label: Tensor(concatenating: data.map { $0.label }, alongAxis: 0),
@@ -116,6 +116,6 @@ func loadCIFARTrainingFiles(localStorageDirectory: URL) -> LabeledExample {
     )
 }
 
-func loadCIFARTestFile(localStorageDirectory: URL) -> LabeledExample {
-    return loadCIFARFile(named: "test_batch.bin", in: localStorageDirectory)
+func loadCIFARTestFile(localStorageDirectory: URL, normalize: Bool = true) -> LabeledExample {
+    return loadCIFARFile(named: "test_batch.bin", in: localStorageDirectory, normalize: normalize)
 }
--- a/Datasets/CIFAR10/CIFAR10.swift
+++ b/Datasets/CIFAR10/CIFAR10.swift
@@ -88,17 +88,15 @@ func loadCIFARFile(named name: String, in directory: URL, normalizing: Bool = tr
     let images = Tensor<UInt8>(shape: [imageCount, 3, 32, 32], scalars: bytes)
 
     // Transpose from the CIFAR-provided N(CHW) to TF's default NHWC.
-    let imageTensor = Tensor<Float>(images.transposed(permutation: [0, 2, 3, 1]))
+    var imageTensor = Tensor<Float>(images.transposed(permutation: [0, 2, 3, 1]))
 
     if normalizing {
         let mean = Tensor<Float>([0.485, 0.456, 0.406])
         let std = Tensor<Float>([0.229, 0.224, 0.225])
-        let imagesNormalized = ((imageTensor / 255.0) - mean) / std
-        return LabeledExample(label: Tensor<Int32>(labelTensor), data: imagesNormalized)
+        imageTensor = ((imageTensor / 255.0) - mean) / std
     }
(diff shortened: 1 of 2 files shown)


manideep2510:
Resolves #285

rohanreddych:
Hey @manideep2510 I think it would be better if we  change the function call (2x) in the functions below this one.

manideep2510:
Hi @rohanreddych,
Changed all the functions involving the function ```loadCIFARFile```. 
Thanks for the suggestion. I overlooked it by mistake.

Datasets/CIFAR10/CIFAR10.swift
```diff
@@ -94,21 +94,28 @@ func loadCIFARFile(named name: String, in directory: URL) -> LabeledExample {
 
     let mean = Tensor<Float>([0.485, 0.456, 0.406])
     let std = Tensor<Float>([0.229, 0.224, 0.225])
-    let imagesNormalized = ((imageTensor / 255.0) - mean) / std
 
-    return LabeledExample(label: Tensor<Int32>(labelTensor), data: imagesNormalized)
+    // Normalize the images according to the input boolean flag passed (Default to 'true')
+    if normalize {
```
xihui-wu: I would do in this way:

var imageTensor = Tensor<Float>(images.transposed(permutation: [0, 2, 3, 1]))
let mean = Tensor<Float>([0.485, 0.456, 0.406])
let std = Tensor<Float>([0.229, 0.224, 0.225])
imageTensor = normalize ? ((imageTensor / 255.0) - mean) / std : imageTensor

return LabeledExample(label: Tensor<Int32>(labelTensor), data: imageTensor)

manideep2510: I was actually following the optional normalization written in [swift-models/Datasets/MNIST/MNIST.swift](Datasets/MNIST/MNIST.swift), So I have used the '''if''' condition. Your suggestion does seem compact. Will do it.

manideep2510:
@xihui-wu Thank you for pointing out, I will update the init method. 

> Also, can you add Test?

By "Test" you mean building the code on my local machine to look for errors or something else?
And I am kind of new to this, it would be great if you could point me to some examples of what a Test is and how to do this.
Thank you.

xihui-wu:
> @xihui-wu Thank you for pointing out, I will update the init method.
> 
> > Also, can you add Test?
> 
> By "Test" you mean building the code on my local machine to look for errors or something else?
> And I am kind of new to this, it would be great if you could point me to some examples of what a Test is and how to do this.
> Thank you.

I mean unit test your change by adding a test case to https://github.com/tensorflow/swift-models/blob/master/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift.

Datasets/CIFAR10/CIFAR10.swift
```diff
@@ -94,21 +94,28 @@ func loadCIFARFile(named name: String, in directory: URL) -> LabeledExample {
 
     let mean = Tensor<Float>([0.485, 0.456, 0.406])
     let std = Tensor<Float>([0.229, 0.224, 0.225])
-    let imagesNormalized = ((imageTensor / 255.0) - mean) / std
 
-    return LabeledExample(label: Tensor<Int32>(labelTensor), data: imagesNormalized)
+    // Normalize the images according to the input boolean flag passed (Default to 'true')
```
BradLarson: Comments should express information that is unclear from examining the code. In this case, I think it's pretty easy to follow that if `normalize` is true, the images will be normalized, so I think you're safe to remove this comment.

Datasets/CIFAR10/CIFAR10.swift
```diff
@@ -94,21 +94,28 @@ func loadCIFARFile(named name: String, in directory: URL) -> LabeledExample {
 
     let mean = Tensor<Float>([0.485, 0.456, 0.406])
     let std = Tensor<Float>([0.229, 0.224, 0.225])
```
BradLarson: Because the mean and standard deviation constants are only used when normalizing the images, my recommendation would be to move them inside of the if statement to be right before the normalization occurs.

Datasets/CIFAR10/CIFAR10.swift
manideep2510: Done

Datasets/CIFAR10/CIFAR10.swift
manideep2510: Done

Datasets/CIFAR10/CIFAR10.swift
manideep2510: I think using ```if/else``` statement is (infinitesimally) better because ```mean``` and ```std``` tensors won't be declared when we don't want to normalize the images.

manideep2510:
> Also, can you add Test?

Added a normalization argument to the current test case in https://github.com/tensorflow/swift-models/blob/master/Tests/DatasetsTests/CIFAR10/CIFAR10Tests.swift

Let me know if that's not enough.

Another idea for a test case, returning both Normalized and Non-Normalized images and compare them to know whether the image is correctly normalized or not. But, I think that's not required and might be overkill.

manideep2510:
@BradLarson Corrected the errors (Not initialized the arguments properly in init) because of which Kokoro failed. I am able to run the code on my local computer without error now.

PR #352, opened 2020-02-26
build: add testing support to the CMake build

compnerd:
Add support for tests to CMake to enable running the test suite on
Windows.


compnerd:
`cmake --build ... --target test` or `ninja -C ... test`.  I'll be adding this to the documentation in a follow up commit.

PR #353, opened 2020-02-26
docs: update documentation on how to test with CMake

compnerd:
Add example command lines to demonstrate how to run the test suite via
CMake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,37 +37,54 @@ It is highly recommended that you use CMake 3.16 or newer to ensure that `-B` an
 macOS:
 
 ```
+# Configure
 cmake                                                              \
   -B /BinaryCache/tensorflow-swift-models                          \
+  -D BUILD_TESTING=YES                                             \
   -D CMAKE_BUILD_TYPE=Release                                      \
   -D CMAKE_Swift_COMPILER=$(TOOLCHAINS=tensorflow xcrun -f swiftc) \
   -G Ninja                                                         \
   -S /SourceCache/tensorflow-swift-models
+# Build
+cmake --build /BinaryCache/tensorflow-swift-models
+# Test
+cmake --build /BinaryCache/tensorflow-swift-models --target test
 ```
 
 Linux:
 
 ```
+# Configure
 cmake                                     \
   -B /BinaryCache/tensorflow-swift-models \
+  -D BUILD_TESTING=NO                     \
   -D CMAKE_BUILD_TYPE=Release             \
   -D CMAKE_Swift_COMPILER=$(which swiftc) \
   -G Ninja                                \
   -S /SourceCache/tensorflow-swift-models
+# Build
+cmake --build /BinaryCache/tensorflow-swift-models
 ```
 
 Windows:
 
 ```
 set SDKROOT=C:/Library/Developer/Platforms/Windows.platform/Developer/SDKs/Windows.sdk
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,37 +37,54 @@ It is highly recommended that you use CMake 3.16 or newer to ensure that `-B` an
...
 cmake                                     \
   -B /BinaryCache/tensorflow-swift-models \
+  -D BUILD_TESTING=NO                     \
   -D CMAKE_BUILD_TYPE=Release             \
   -D CMAKE_Swift_COMPILER=$(which swiftc) \
   -G Ninja                                \
   -S /SourceCache/tensorflow-swift-models
+# Build
+cmake --build /BinaryCache/tensorflow-swift-models
 ```
```
saeta: Testing example?

README.md
```diff
@@ -37,37 +37,54 @@ It is highly recommended that you use CMake 3.16 or newer to ensure that `-B` an
...
-  -D CMAKE_Swift_FLAGS="-sdk %SDKROOT% -I %SDKROOT%/usr/lib/swift -L %SDKROOT%/usr/lib/swift/windows"                ^
-  -G Ninja                                                                                                           ^
+set DEVELOPER_LIBRARY_DIR=C:/Library/Developer/Platforms/Windows.platform/Developer/Library
+: Configure
+"%ProgramFiles%\CMake\bin\cmake.exe"                                                                                                                                                                                                                  ^
+  -B C:/BinaryCache/tensorflow-swift-models                                                                                                                                                                                                           ^
+  -D BUILD_SHARED_LIBS=YES                                                                                                                                                                                                                            ^
```
dan-zheng: Woah, different end lines characters.

README.md
```diff
@@ -37,37 +37,54 @@ It is highly recommended that you use CMake 3.16 or newer to ensure that `-B` an
 macOS:
 
 ```
+# Configure
 cmake                                                              \
   -B /BinaryCache/tensorflow-swift-models                          \
+  -D BUILD_TESTING=YES                                             \
```
saeta: Do you need this if you're not building tests? On the flip side, why shouldn't this be the default?

compnerd: Okay, so, you caught me.  This is a slight .... unnecessary bit of typing.  This *is* the default.  I just want to explicitly have the user realize that they are building testing which actually changes the built products!  Your library is now built with `-enable-testing`.  This means that this library should ideally not be shipped in this form.  You need to explicitly pass `-D BUILD_TESTING=NO` to *disable* testing support.  This is similar to how the default build type is Debug, which you really don't want to ship.  I'm okay with removing it if you feel that its verbose without much value.

README.md
compnerd: Need to test it on Linux which is why I haven't put it there - the entire thing needs the answer to: is XCTest available?

README.md
compnerd: Yes, the line continuation character is different in CMD.  The comment leader is different too!

README.md
dan-zheng: Great explanation of the context! I think keeping `-D BUILD_TESTING=YES` makes sense.

README.md
dan-zheng: Aha, it's a line continuation character like `\`. Not an end line character (`LF` vs `CRLF`).

PR #355, opened 2020-03-02
Follow-up on comments of #338

sgugger:
Addresses @dan-zheng comments on #338 
I also changed the name of `items` to `numericalizedTexts` as it documents better what the expected input is.


Datasets/LanguageModelDataset.swift
```diff
@@ -1,65 +1,56 @@
...
 
-//Build a dataset suitable for language modeling from an array of texts
-public struct LanguageModelDataset<Item> {
-  public typealias Index = Int
-  public typealias Element = TensorPair<Int32, Int32>
-
-  //A function that reads Item to get an array of Int
-  public let openItem: (Item) -> [Int]
-  //The size of a batch
+/// A dataset suitable for language modeling
```
dan-zheng: Please end all doc comments with a period.

Datasets/LabeledExample.swift
```diff
@@ -35,9 +35,9 @@ public struct LabeledExample: TensorGroup, KeyPathIterable, Collatable {
     }
 }
 
-// A simple tuple of Tensor, useful for many deep learning tasks (one tensor of inputs,
+// A simple tuple of `Tensor`, useful for many deep learning tasks (one tensor of inputs,
```
dan-zheng: Would you mind applying the doc comment that I suggested in https://github.com/tensorflow/swift-models/pull/338/files#r385994099?

Please note the comments I left there as well!

Datasets/LanguageModelDataset.swift
```diff
@@ -105,28 +98,7 @@ extension LanguageModelDataset: Collection {
...
-      openItem: { $0 },
-      batchSize: batchSize,
-      sequenceLength: sequenceLength,
-      items: items,
-      lengths: items.map { $0.count })
-  }
-}
-
-//sampleIndices function to use in conjunction with a LanguageModelDataset
+/// sampleIndices function to use in conjunction with a `LanguageModelDataset`
```
dan-zheng: This doc comment should read like a sentence. The meaning right now is unclear to me:
* What is the returned `[Int]`?
* Why is the dataset `inout`?

Please add `/// - Parameters:` and `/// Returns:` doc comments.

Datasets/LanguageModelDataset.swift
```diff
@@ -1,65 +1,56 @@
...
-//Build a dataset suitable for language modeling from an array of texts
-public struct LanguageModelDataset<Item> {
-  public typealias Index = Int
-  public typealias Element = TensorPair<Int32, Int32>
-
-  //A function that reads Item to get an array of Int
-  public let openItem: (Item) -> [Int]
-  //The size of a batch
+/// A dataset suitable for language modeling 
+public struct LanguageModelDataset<C> where C: Collection, C.Index==Int, C.Element==[Int] {
```
dan-zheng: Could you find a more descriptive name than `C`? Generic parameter names are important: `Tensor<Scalar>` is much clearer than `Tensor<T>`.

Datasets/LanguageModelDataset.swift
sgugger: Done.

Datasets/LanguageModelDataset.swift
sgugger: Added more prose, let me know if it's still unclear.

Datasets/LanguageModelDataset.swift
sgugger: Used `Texts`  since it's the type of the collection of the underlying texts. Let me know if you have a better idea.

Datasets/LabeledExample.swift
sgugger: Sorry I had missed that one. I changed the text a little bit because it's weird to have a documentation on something say: do not use this.

What other comments were you referring to>

dan-zheng: Awesome edits! That's all I wanted 🙂

Datasets/LanguageModelDataset.swift
```diff
@@ -1,65 +1,57 @@
...
+where Texts: Collection, Texts.Index==Int, Texts.Element==[Int] {
+  /// The size of a batch.
   public var batchSize: Int
-  //The length of a sequence
+  /// The length of a sequence.
   public var sequenceLength: Int
-  //The array of raw items to use
-  public let items: [Item]
-  //The length of each processed item
+  /// The array of raw items to use.
```
dan-zheng: I think this doc comment could use improvement.
I think this is an improvement:
```suggestion
  /// The collection of numericalized texts.
```
But still not great. What does "numericalized" means?

Datasets/LanguageModelDataset.swift
```diff
@@ -105,28 +99,12 @@ extension LanguageModelDataset: Collection {
...
-      items: items,
-      lengths: items.map { $0.count })
-  }
-}
-
-//sampleIndices function to use in conjunction with a LanguageModelDataset
+/// The sampleIndices function to use in conjunction with a `LanguageModelDataset` in a `Batcher`.
+/// Will shuffle the dataset in place instead of the indices (like the default function does).
+/// - Parameters:
+///   - dataset: The underlying `LanguageModelDataset`
```
dan-zheng: Please consistently use periods.
```suggestion
///   - dataset: The underlying `LanguageModelDataset`.
```

Datasets/LanguageModelDataset.swift
sgugger: numericalized is a standard NLP text. Preprocessing raw texts has two steps:
- tokenization, which is splitting a text in tokens (either characters, words, or part of words depending on the tokenizer)
- numericalization, which is converting those tokens to numbers using a vocab.

Datasets/LanguageModelDataset.swift
```diff
@@ -1,65 +1,57 @@
...
-public struct LanguageModelDataset<Item> {
-  public typealias Index = Int
-  public typealias Element = TensorPair<Int32, Int32>
-
-  //A function that reads Item to get an array of Int
-  public let openItem: (Item) -> [Int]
-  //The size of a batch
+/// A dataset suitable for language modeling.
+public struct LanguageModelDataset<Texts> 
+where Texts: Collection, Texts.Index==Int, Texts.Element==[Int] {
```
dan-zheng: It may be surprising that `Texts` is a `Collection` where `Index == Int` and `Element == [Int]`. One might expect `Texts` to be a collection of `String`.

Could you please add doc comments that help explain the `Texts` generic parameter? Maybe you can clarify that it's numericalized.

sgugger: Did that and added some explanation on what those terms mean. `NumericalizedTexts` is kind of long for the generic parameters, which is why I stuck to `Texts`. Let me know if there is anything else I can do to make this clearer.

Datasets/LanguageModelDataset.swift
```diff
@@ -1,65 +1,63 @@
...
-
-  //A function that reads Item to get an array of Int
-  public let openItem: (Item) -> [Int]
-  //The size of a batch
+/// A dataset suitable for language modeling.
+///
+/// - Note: This struct does not handle the preprocessing required in NLP
+/// and expects you have already tokenized and numericalized your raw texts
+/// (that is split them in tokens, then mapped those tokens to their ids in your
+/// vocabulary). Therefore the generic type Texts refers to a collection of
```
dan-zheng: Please take care to add backticks! It can really speed up review if you proofread your writing or write backticks the first time :)
```suggestion
/// vocabulary). Therefore the generic type `Texts` refers to a collection of
```

Datasets/LanguageModelDataset.swift
```diff
@@ -105,28 +105,12 @@ extension LanguageModelDataset: Collection {
...
-      lengths: items.map { $0.count })
-  }
-}
-
-//sampleIndices function to use in conjunction with a LanguageModelDataset
+/// The sampleIndices function to use in conjunction with a `LanguageModelDataset` in a `Batcher`.
+/// Will shuffle the dataset in place instead of the indices (like the default function does).
+/// - Parameters:
+///   - dataset: The underlying `LanguageModelDataset`.
+///   - shuffled: Shuffle or not the dataset.
```
dan-zheng: Should `shuffled` be named `shuffle`? The boolean flag sounds like it should be used as an imperative verb.
```suggestion
///   - shuffled: Shuffles the data iff `true`.
```

sgugger: That would be a broader change in `Batcher`. I have no strong opinion on either syntax.

Datasets/LanguageModelDataset.swift
sgugger: I actually wasn't sure on this one since it's not a type but a generic parameter. Sorry about that!

Datasets/LabeledExample.swift
```diff
@@ -35,10 +35,12 @@ public struct LabeledExample: TensorGroup, KeyPathIterable, Collatable {
     }
 }
 
-// A simple tuple of Tensor, useful for many deep learning tasks (one tensor of inputs,
-// one tensor of labels). Note that if your task is more elaborate, you should write your
-// own struct of Tensors. Automatic conformance to Collatable can be derived the same way as
-// here.
+/// A generic tuple of two tensors `Tensor`.
```
dabrahams: “two tensors `Tensor`” doesn't quite parse for me.

How is this a “tuple” and not a “pair?”  The former usually implies it has contains an arbitrary number of elements.

I would say “A pair of `Tensor`s having heterogeneous scalar types”

Datasets/LabeledExample.swift
```diff
@@ -35,10 +35,12 @@ public struct LabeledExample: TensorGroup, KeyPathIterable, Collatable {
...
+/// 
+/// - Note: `TensorPair` has a generic name and provides little semantic information, to conform to
+/// `Collatable`. You can use it for most basic datasets with one tensor of inputs and one tensor of
```
dabrahams: Nit: Markdown bullets should left-align all the text, so there should be two spaces before all but the first line of the bullet.

I don't know about `Collatable` yet, but it's hard to imagine how conforming to a protocol could impose naming requirements, or requirements on the non-provision of semantic information.  I think you should try again for this comment, thinking about what you really mean to say.  Hint: there's no need to make excuses for providing a very general-purpose component and naming it accordingly ;-)

Datasets/LanguageModelDataset.swift
```diff
@@ -1,45 +1,45 @@
...
-  //The size of a batch
+/// A dataset suitable for language modeling.
+///
+/// - Note: This struct does not handle the preprocessing required in NLP
+/// and expects you have already tokenized and numericalized your raw texts
+/// (that is split them in tokens, then mapped those tokens to their ids in your
+/// vocabulary). Therefore the generic type `Texts` refers to a collection of
+/// numericalized texts.
+public struct LanguageModelDataset<Texts> 
+where Texts: Collection, Texts.Index==Int, Texts.Element==[Int] {
```
dabrahams: `Index==Int` is almost invariably an over-constraint.  What you probably mean is `Texts: RandomAccessCollection`; you can always get to the nth element via `t[t.index(t.startIndex, offsetBy: n)]` which is admittedly horrible, but can be encapsulated so you don't have to repeat it.

Datasets/LanguageModelDataset.swift
dabrahams: How about `textIDs`?  Is there anything more to “numericalizing” than associating each token with an integer ID?

Datasets/LanguageModelDataset.swift
```diff
@@ -1,45 +1,45 @@
...
   //Drop the last batch if its length is less than sequenceLength
   public let dropLast: Bool
   //The length of a contiguous chunk of text
   private var batchLength: Int
-  //The number of batches
+  /// The number of batches.
   private var batchCount: Int
-  //The sequence length of the last batch
+  /// The sequence length of the last batch.
   private var lastLength: Int
```
dabrahams: I realize it's not part of this PR, but this should be `lastBatchLength`.

Datasets/LanguageModelDataset.swift
```diff
@@ -114,30 +114,12 @@ extension LanguageModelDataset: Collection {
...
-      sequenceLength: sequenceLength,
-      items: items,
-      lengths: items.map { $0.count },
-      dropLast: dropLast)
-  }
-}
-
-//sampleIndices function to use in conjunction with a LanguageModelDataset
+/// The sampleIndices function to use in conjunction with a `LanguageModelDataset` in a `Batcher`.
+/// Will shuffle the dataset in place instead of the indices (like the default function does).
```
dabrahams: Need a blank line after the summary.

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -57,8 +57,8 @@ public struct TextUnsupervised {
         var fileExtension = "tgz"
     }
 
-    public let trainingDataset: LanguageModelDataset<[Int]>
-    public let validationDataset: LanguageModelDataset<[Int]>
+    public let trainingDataset: LanguageModelDataset<[[Int]]>
```
dabrahams: What is this `[[Int]]` I keep seeing?  Would a `typealias` for it make the code more descriptive?

PR #357, opened 2020-03-04
Using ArgumentParser to parse CLI options for Benchmarks target.

### LINKED ISSUE #354: Evaluate replacing Commander with ArgumentParser in Benchmarks
Apple recently unveiled the [ArgumentParser](https://swift.org/blog/argument-parser) library for easing the parsing of command-line options. The [Benchmarks](https://github.com/tensorflow/swift-models/tree/master/Benchmarks) target currently uses Kyle Fuller's [Commander](https://github.com/kylef/Commander) to provide a clean interface for command-line options.

[ArgumentParser](https://github.com/apple/swift-argument-parser) provides an intriguing alternative, and seems to have growing support behind it, so we would like to examine if it is practical and desirable to switch from Commander to ArgumentParser within the Benchmarks target. Ideally, we would like to maintain the same command-line options and interface as exists now, so that Benchmarks won't change its interaction with internal test runners.

If Benchmarks could be converted to use ArgumentParser instead, while maintaining the same interface, and the code for doing so looked reasonable, I think we'd welcome a PR that made this change.

vguerra:
This PR maintains the same interface that was defined with Commander,
so it is a NFC in a sense.

Resolves #354.

### DIFF
--- a/Benchmarks/main.swift
+++ b/Benchmarks/main.swift
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import Commander
+import ArgumentParser
 import ModelSupport
 
 /// Collect benchmark results and print them to stdout using a given printing style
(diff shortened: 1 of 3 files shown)


vguerra:
@googlebot I signed it!

Benchmarks/main.swift
```diff
@@ -132,5 +90,101 @@ let main =
...
+        @Flag(help: "Run training benchmark.")
+        var training: Bool
+
+        @Flag(help: "Run inference benchmark.")
+        var inference: Bool
+
+        @Option(default: "", help: "Name of the benchmark to run.")
+        var benchmark: String
+
+        @Option(default: -1, help: "Number of batches.")
```
dan-zheng: What happens when these `Int` options have default value `-1`?
- `batches`
- `batchSize`
- `iterations `
- `epochs`

It seems they didn't have validation before. Do they need validation?

vguerra: No, they did not have validation before and since the idea was to not change behaviour I preferred not to add validation.

But having a closer look I see that those options are being ignored in https://github.com/tensorflow/swift-models/blob/master/Benchmarks/main.swift#L19. as each `BenchMarkModel` is supposed to define it's own default `BenchmarkSettings`.

So as a follow-up, we can either remove those CLI options or adapt `runBenchmark` to make use of them ( cc: @BradLarson ), wdyt?

Wdel06:
#357

vguerra:
> It looks like this PR changes `batchSize` flag to `batch-size`. Can you please update the PR to make it compatible with the previous command-line syntax?

hi @shabalind, thank you for the review. Indeed, ArgumentParser seems to translate camel-case option's name to it's dash-separated version. Setting a custom name is needed in this case,  https://github.com/tensorflow/swift-models/pull/357/commits/2fde48e1ab7be8aebc8970e505585ca5dd99389e fixes this.

PR #358, opened 2020-03-04
Move BytePairEncoder to ModelSupport

texasmichelle:
Move out of Model/Text to avoid circular dependencies caused by usage in datasets
Rename file to match struct name


texasmichelle:
I think this is necessary in order to use LanguageModelDataset with GPT2, since the dataset stores byte pair encoded indexes.

texasmichelle:
Is there a way to surface CMake errors before running kokoro? I don't get any failures locally with `swift test`.

BradLarson:
For surfacing CMake issues, Saleem just added some sample commands to [the Readme](https://github.com/tensorflow/swift-models) and I used those once to test things. However, I'm still forgetting things that CI surfaces, so I'm still not great at it.

texasmichelle:
Ah, that's right - thanks for pointing me to the CMake test instructions.

PR #361, opened 2020-03-05
Transformer: merge pair fixes

texasmichelle:
- Import merge pairs from the correct file
- Fix merge pair functionality to operate on characters instead of tokens
- Use vocabulary inside BytePairEncoder rather than a local variable
- Minor cleanup in TransformerLoop
- Lint

### DIFF
--- a/Support/Text/BytePairEncoder.swift
+++ b/Support/Text/BytePairEncoder.swift
@@ -24,7 +24,8 @@ public struct BytePairEncoder {
     /// A cache used to store encoded tokens and thus speed up encoding.
     //  private var cache: [String: [String]]
 
-    public init(vocabularyFile: URL, mergesFile: URL,
+    public init(
+        vocabularyFile: URL, mergesFile: URL,
         encoding: String.Encoding = .utf8
     ) throws {
         let vocabulary: Vocabulary = try Vocabulary(fromJSONFile: vocabularyFile)
--- a/Support/Text/BytePairEncoder.swift
+++ b/Support/Text/BytePairEncoder.swift
@@ -72,7 +72,7 @@ public struct BytePairEncoder {
         switch variant {
         case .gpt2:
             // Split into parts before encoding.
-            let unencodedTokens: [String] = BytePairEncoder.splittingWithDelimiters(
+            let unencodedTokens = BytePairEncoder.splittingWithDelimiters(
                 token: token,
                 glossaryRegex: BytePairEncoder.gpt2GlossaryRegex,
                 variant: .gpt2)
@@ -107,7 +107,7 @@ public struct BytePairEncoder {
         }
 
         // Check if the new word parts are in the vocabulary, and backtrack if necessary.
-        let encoded: [String] = parts.flatMap { part -> [String] in
+        let encoded = parts.flatMap { part -> [String] in
             if vocabulary.contains(part) { return [part] }
             return splitRecursively(part)
         }
--- a/Support/Text/BytePairEncoder.swift
+++ b/Support/Text/BytePairEncoder.swift
@@ -24,21 +24,25 @@ public struct BytePairEncoder {
     /// A cache used to store encoded tokens and thus speed up encoding.
     //  private var cache: [String: [String]]
 
-    public init(fromVocabularyFileURL vocabularyFile: URL, fromMergesFileURL mergesFile: URL,
-        encoding: String.Encoding = .utf8) throws {
+    public init(
+        fromVocabularyFileURL vocabularyFile: URL, fromMergesFileURL mergesFile: URL,
+        encoding: String.Encoding = .utf8
+    ) throws {
         let vocabulary: Vocabulary = try Vocabulary(fromJSONFile: vocabularyFile)
 
         let lines: ArraySlice<String> =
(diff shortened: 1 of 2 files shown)


Support/Text/BytePairEncoder.swift
```diff
@@ -62,21 +67,25 @@ public struct BytePairEncoder {
...
-        var parts: [String]
-        var encodedParts: [String]
-        var encoded: [String]
+        var parts = [String]()
 
         switch variant {
         case .gpt2:
             // Split into parts before encoding.
-            parts = BytePairEncoder.splittingWithDelimiters(
+            let unencodedTokens: [String] = BytePairEncoder.splittingWithDelimiters(
```
BradLarson:
```suggestion
            let unencodedTokens = BytePairEncoder.splittingWithDelimiters(
```
Minor, but the type should be able to be inferred here.

Support/Text/BytePairEncoder.swift
```diff
@@ -97,18 +106,8 @@ public struct BytePairEncoder {
...
-            encodedParts = parts.map({ BytePairEncoder.encodedToken($0) })
-
-        case .roberta, .none:
-            // Encoding has already occurred.
-            encodedParts = parts
-
-        }
         // Check if the new word parts are in the vocabulary, and backtrack if necessary.
-        encoded = encodedParts.flatMap { part -> [String] in
+        let encoded: [String] = parts.flatMap { part -> [String] in
```
BradLarson:
```suggestion
        let encoded = parts.flatMap { part -> [String] in
```

Again minor, but same deal as above.

Transformer/TransformerLoop.swift
```diff
@@ -14,111 +14,113 @@
...
+        let merges: (file: URL, data: Data) = try (mergesFileURL, Data(contentsOf: mergesFileURL))
+
+        // Create bytepair encoder with loaded token mappings
+        bpe = try BytePairEncoder(
+            fromVocabularyFileURL: vocabulary.file, fromMergesFileURL: merges.file)
+
+        // ...
+        let seedId: Int32 = Int32(bpe.vocabulary.id(forToken: "<|endoftext|>")!)
+        seed = Tensor(shape: [1, 1], scalars: [seedId])
+        let empty: Tensor<Float> =
```
BradLarson: Similarly, the type should be able to be inferred here, so an explicit annotation isn't needed. That might lead to a larger reformatting of this line, so I won't try a suggestion here.

Transformer/TransformerLoop.swift
```diff
@@ -14,111 +14,113 @@
...
+            AttentionContext(key: empty, value: empty)
+        }
     }
 
-    throw GPT2Error.invalidEncoding(id: id)
-  }
+    func embedding(for string: String) -> Tensor<Int32> {
+        let tokens: [String] = bpe.encode(token: string, variant: .gpt2)
+        // TODO(michellecasbon): Decide how to prevent OOV or choose a better ID (probably not 0).
+        let ids: [Int32] = tokens.map { Int32(bpe.vocabulary.id(forToken: $0) ?? 0) }
```
BradLarson:
```suggestion
        let tokens = bpe.encode(token: string, variant: .gpt2)
        // TODO(michellecasbon): Decide how to prevent OOV or choose a better ID (probably not 0).
        let ids = tokens.map { Int32(bpe.vocabulary.id(forToken: $0) ?? 0) }
```

Transformer/TransformerLoop.swift
```diff
@@ -14,111 +14,113 @@
...
+            (result.shape[0], result.shape[1], result.shape[2])
+        let logits =
+            result.slice(
+                lowerBounds: [0, timesteps - 1, 0],
+                upperBounds: [batchSize, timesteps, vocabularySize]) / temperature
+        seed = Tensor(
+            randomCategorialLogits: logits.squeezingShape(at: 1),
+            sampleCount: 1)
+
+        let id: Int32 = Int32(seed[0][0])!
```
BradLarson:
```suggestion
        let id = Int32(seed[0][0])!
```

Support/Text/BytePairEncoder.swift
```diff
@@ -24,20 +24,25 @@ public struct BytePairEncoder {
     /// A cache used to store encoded tokens and thus speed up encoding.
     //  private var cache: [String: [String]]
 
-    public init(fromFileURL file: URL, encoding: String.Encoding = .utf8) throws {
-        let vocabulary: Vocabulary = try Vocabulary(fromJSONFile: file)
+    public init(
+        fromVocabularyFileURL vocabularyFile: URL, fromMergesFileURL mergesFile: URL,
```
BradLarson: It's the convention in Swift to try to avoid having types as part of the parameter name. Perhaps we could remove the alternative parameter names and just use `vocabularyFile` and `mergesFile`? Dave or Dan might have additional naming  suggestions.

Transformer/TransformerLoop.swift
texasmichelle: Done, for the `empty` and `seedId` declarations.

Support/Text/BytePairEncoder.swift
texasmichelle: Done

PR #362, opened 2020-03-06
Adding ability to download compressed model checkpoints

BradLarson:
Extending CheckpointReader, this adds the ability to provide a URL to a .tar.gz or otherwise compressed model checkpoint, and have the reader download, extract, and read from the compressed checkpoint. It caches the uncompressed checkpoint in a temporary directory, and will not re-download it if present. If the archived checkpoint is on the local filesystem, it extracts directly from that file.

As a little cleanup, this also pulls out archive decompression into a standalone `extractArchive(at:to:fileExtension:deleteArchiveWhenDone)` function so that a common implementation of this can be used both by the datasets and the checkpoint reader.


Support/FileManagement.swift
```diff
@@ -63,23 +63,95 @@ public func download(from source: URL, to destination: URL) throws {
...
+    at archive: URL, to localStorageDirectory: URL, fileExtension: String? = nil,
+    deleteArchiveWhenDone: Bool = false
+) {
+    let archivePath = archive.path
+
+    #if os(macOS)
+        var binaryLocation = "/usr/bin/"
+    #else
+        var binaryLocation = "/bin/"
+    #endif
```
shabalind: Is this check here going to work on Windows?

BradLarson: It will not, anything dependent on running UNIX unarchiving tools at the command line isn't going to be portable to Windows at present. We'll need to find a more generalized approach (preferably unarchiving directly through a library or  some other means). This works on macOS and Linux currently and will build on Windows, just won't run correctly there.

BradLarson:
@texasmichelle - Good point on the tests. Assembled a local SavedModel and remote checkpoint test case that should exercise the various components. The tests uncovered a problem area that I hadn't thought of, so was able to fix that in the process.

Tests/CheckpointTests/CheckpointReaderTests.swift
```diff
@@ -0,0 +1,59 @@
...
+import XCTest
+
+@testable import ModelSupport
+
+final class CheckpointReaderTests: XCTestCase {
+    let resourceBaseLocation = URL(fileURLWithPath: #file).deletingLastPathComponent()
+        .appendingPathComponent("SavedModels")
+
+    func testLocalSavedModel() {
+        // The data shard for the SavedModel has been replaced with a 0-byte placeholder. This
```
texasmichelle: This is a great solution 🎉

PR #363, opened 2020-03-07
GPT-2 example

texasmichelle:
Adds a new example that fine-tunes GPT-2 on the WikiText2 dataset.

- [X] Adds the WikiText2 dataset from the [LanguageModel notebook](https://github.com/tensorflow/swift-models/blob/master/Examples/Notebooks/LanguageModel.ipynb)
- [X] Moves Transformer code into Models/Text/GPT2
- [X] Loads the OpenAI GPT-2 base model from a python checkpoint
- [X] Fine-tunes the loaded model
- [X] Lints
- [x] Includes tests
- [x] Reports validation accuracy properly
- [x] Supports similarly formatted datasets (WikiText2, WikiText103)


jekbradbury:
The WikiText2 dataset is formatted exactly the same way as Penn Treebank (the classic language modeling dataset from early deep learning papers) and WikiText103, so your loader is probably already more generic than you might realize!

texasmichelle:
Oooh it sounds like I need some better naming and parameterization :)

texasmichelle:
Removing the BERT code-sharing bullet item. Despite overlapping names, the objects AttentionInput, MultiHeadAttentionInput, and Embedding are used in very different ways in BERT and GPT2. I don't see a straightforward way of combining them, so for now a GPT2 suffix will avoid blowing up the scope of this already large PR.

texasmichelle:
I converted WikiText2 to a generic dataset `TextUnsupervised` that also supports WikiText103. I included tests, but commented them out for Kokoro since the dataset is > 500MB.

Penn TB has a licensing requirement so I did not include it, but the modifications to`TextUnsupervised` should be minimal for adding additional sources.

Models/Text/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ set_target_properties(TextModels PROPERTIES
   INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
 target_compile_options(TextModels PRIVATE
   $<$<BOOL:${BUILD_TESTING}>:-enable-testing>)
+target_link_libraries(TextModels PUBLIC
+  Datase)
```
brettkoonce: Database?

Models/Text/GPT2/Model.swift
```diff
@@ -275,14 +307,14 @@ public struct Embedding: Differentiable {
     }
 }
 
-public struct TransformerLM {
```
xihui-wu: How about naming this file to TransformerLM.swift ?

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -0,0 +1,183 @@
...
+    public init(
+        bpe: BytePairEncoder, variant: TextUnsupervisedVariant = TextUnsupervisedVariant.wikiText2
+    ) {
+        self.init(
+            localStorageDirectory: FileManager.default.temporaryDirectory.appendingPathComponent(
+                variant.rawValue, isDirectory: true), bpe: bpe, variant: variant
+        )
+    }
+
+    public init(
```
xihui-wu: Can we merge the two init(s) into one ?

Models/Text/GPT2/Model.swift
texasmichelle: Done

Datasets/TextUnsupervised/TextUnsupervised.swift
texasmichelle: I refactored to include defaults so that one can be called without parameters, similar to the [Imagenette dataset](https://github.com/tensorflow/swift-models/blob/master/Datasets/Imagenette/Imagenette.swift#L45-L47). I'm not sure how useful this will be in practice, but it simplifies some of our tests.

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -0,0 +1,188 @@
...
+
+        let documentsFull = try readCSV(in: path)
+        // TODO(michellecasbon): Process a larger number of documents.
+        let documents = Array(documentsFull[0..<1])
+
+        let embeddings = documents.map { embedding(for: $0, bpe: bpe) }
+        let lengths = embeddings.map { $0.count }
+
+        return LanguageModelDataset(
+            batchSize: 64,
```
sgugger: This should be the same as the `batchSize` used in the training loop (4). I'd suggest either hard-code 4 or make this a parameter we can pass.

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -0,0 +1,188 @@
...
+            sequenceLength: 72,
```
sgugger: This should be the same as the `contextSize` in the GPT2 model. So same as up there, either hard-code 1024 or make this a parameter.

sgugger:
Two minor changes suggested, but globally looking good to me!

Datasets/TextUnsupervised/TextUnsupervised.swift
texasmichelle: Hm. Using batchSize as a parameter to match the training loop and a sequence length of 1024 to match `contextSize`, I'm getting shape mismatch failures, e.g. 
```
Fatal error: Shapes of all inputs must match: values[0].shape = [119] != values[1].shape = [1024]
```
Does it make sense for Batcher to return a sequence of length 119? My understanding is that anything less than 1024 is dropped (or padded)? What values do you suggest, given these pre-trained values:
```
TransformerLMConfig(vocabSize: 50257, contextSize: 1024, embeddingSize: 768, headCount: 12, layerCount: 12)
```

Datasets/TextUnsupervised/TextUnsupervised.swift
texasmichelle: Thanks for this catch! Parameterizing gives us more flexibility.

Datasets/TextUnsupervised/TextUnsupervised.swift
sgugger: It's possible that the last batch has a sequence length shorter than 1024. If we drop it manually in the loop, the error should disappear (longterm fix would be to have `LanguageModelDataset` handle this optionally). I don't get why you didn't get the mismatch error with 72 before though

texasmichelle: OK, I parameterized both of those values and shapes match now. I think sequence length is being handled properly by `LanguageModelDataset`, without the need to drop manually.

texasmichelle:
This PR is already massive, so let's call it a start and make improvements in follow-up PRs.

PR #365, opened 2020-03-08
pix2pix 

s1ddok:
U-Net based generator model with Facades dataset support. Implementation is a bit rough, but I would love to adjust it in order to be merged into the repo, would love to hear a feedback!


BradLarson:
That's great, thanks for working on this. As a first step, would you be able to add license headers to all files (see the other models for examples on how this is done) and format these according to the Google Style Guide? For the latter, we use [swift-format](https://github.com/apple/swift-format) and have our formatting style codified in the .swift-format file here.

Beyond that, we'd probably want to integrate this into the project's main Package.swift, rather than have it exist as a separate sub-package. At some point, we might even be able to generalize the U-Net layers into something that can be used across examples, but I think this will be fine for now.

I have some smaller comments that I can post inline with the code, but those are the larger formatting issues that come to mind.

s1ddok:
Thanks for feedback, I will add the headers and comply with the swift-format

pix2pix/Sources/pix2pix/Layers.swift
```diff
@@ -0,0 +1,303 @@
...
+        return relu(input)
+    }
+}
+
+struct ELU: ParameterlessLayer {
+    @differentiable
+    func callAsFunction(_ input: Tensorf) -> Tensorf {
+        return elu(input)
+    }
+}
```
BradLarson: I'm curious, why did you need to encapsulate the various activation functions in layers? We've typically left them as functions and then [called them in callAsFunction()](https://github.com/tensorflow/swift-models/blob/master/Models/ImageClassification/ResNet.swift#L102). Doing so saves you from having to define the layers and properties, cutting down on a bit of the code needed.

pix2pix/Sources/pix2pix/Layers.swift
```diff
@@ -0,0 +1,303 @@
...
+                            filterInitializer: { Tensorf(randomNormal: $0, standardDeviation: Tensorf(0.02)) })
+    }
+    
+    @differentiable
+    public func callAsFunction(_ input: Tensorf) -> Tensorf {
+        var x = self.downRELU(input)
+        x = self.downConv(x)
+        x = self.upRELU(x)
+        x = self.upConv(x)
+        x = self.upNorm(x)
```
BradLarson: Unless you need the intermediate values here, you might be able to simplify this using `sequenced(through:)` ([an example](https://github.com/tensorflow/swift-models/blob/master/Models/ImageClassification/ResNet.swift#L163)).

dabrahams:
@s1ddok gentle ping?

s1ddok:
@dabrahams I’m somewhat purposefully stalling here, I want to get CycleGAN PR to be merged first (that one appeared to be a more mature implementation) and then reuse some code between the two.

pix2pix/Sources/pix2pix/Layers.swift
```diff
@@ -0,0 +1,303 @@
...
+        // Calculate mean & variance along H,W axes.
+        let mean = input.mean(alongAxes: [1, 2])
+        let variance = input.variance(alongAxes: [1, 2])
+        let norm = (input - mean) * rsqrt(variance + epsilon)
+        return norm * scale + offset
+    }
+}
+
+/// A 2-D layer applying padding with zeros over a mini-batch.
+public struct ZeroPad2D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
```
Shashi456: ZeroPadding is available in swift-apis, so this redeclaration might not be necessary.

s1ddok:
@BradLarson @dabrahams please take a look at updates. I've tried to combine comments from both this and #400. I think the implementations are very close. Please let me know if I can do anything else in order for this to be merged.

s1ddok:
@BradLarson @dabrahams sorry for bumping this too often, just don't want this to get outdated too much, I'm all yours for fixes while I'm still in context of this stuff, so please take a look if you have a moment, thanks!

pix2pix/Utils.swift
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import TensorFlow
+import Foundation
+
+public typealias Tensorf = Tensor<Float>
```
BradLarson, line 18: I know that it's a bit of a pain to type out Tensor<Float> as an explicit type throughout the models, but we don't use an abbreviated typealias in the other models (CycleGAN as an exception, now that I noticed this here). We might want to avoid a custom typealias for this type in only specific models in order to be consistent across the repository. Typing of Tensors is part of a larger discussion, but we may want to be verbose with these to be consistent right now.

CycleGAN/main.swift
```diff
@@ -185,6 +185,8 @@ for epoch in 0 ..< epochs {
         // MARK: Inference
 
         if step % options.sampleLogPeriod == 0 {
+             Context.local.learningPhase = .inference
```
BradLarson, line 188: As a minor formatting issue, this looks like it got shifted one space too far to the right.

s1ddok:
@BradLarson fixed!

pix2pix/README.md
```diff
@@ -0,0 +1,21 @@
+# pix2pix
+
+This repositery tries to closely resemble original pix2pix paper with the same architecture and hyper-parameters, but using Swift for TensorFlow. Results seem to be close to what original authors claimed to achieve, this is what a generator outputs after 7 hours of training using single GTX 1080 card.
```
BradLarson, line 3:
```suggestion
This example tries to closely resemble [the original pix2pix paper](https://arxiv.org/abs/1611.07004) with the same architecture and hyper-parameters, but using Swift for TensorFlow. Results seem to be close to what the original authors claimed to achieve, this is what a generator outputs after 7 hours of training using single GTX 1080 card:
```

pix2pix/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+![](https://i.imgur.com/Od9dfe8.jpg)
+![](https://i.imgur.com/zQxPCAd.jpg)
+
+In order to run the project you have to just call 
+
+```bash
+swift run -c release
```
BradLarson, line 11:
```suggestion
swift run -c release pix2pix
```

pix2pix/README.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+You will need a tensorboard installation in order to run the project.
```
BradLarson, line 14: With the removal of TensorBoardX, this sentence can probably be removed for now. Again, we'll revisit better TensorBoard support at some point in the future.

pix2pix/README.md
```diff
@@ -0,0 +1,21 @@
...
+```bash
+swift run -c release
+```
+
+You will need a tensorboard installation in order to run the project. 
+
+
+## Notes
+
+- Trainer will emit a lot of intermediate results in a calling directory, you can easily comment out that part. It is a workaround because for some reason tensorboard crashes when provided with result images, I'm investigating on the problem.
```
BradLarson, line 19: Does this note still apply, without the TensorBoard support?

pix2pix/README.md
```diff
@@ -0,0 +1,21 @@
...
+- You need to specify different URL for `dataset` or we download facades dataset for you
+- This implementation only provides a data loader for facades dataset, but the model is capable of learning on any pix2pix compatible dataset.
```
BradLarson, line 21:
```suggestion
- You need to specify a different URL for `dataset` or this will download the facades dataset for you.
- This implementation only provides a data loader for the facades dataset, but the model is capable of learning on any pix2pix compatible dataset.
```

s1ddok:
@BradLarson Sorry I completely forgot about outdated README, I've accepted all your grammar fixes and removed all of TensorBoard mentions.

PR #367, opened 2020-03-10
TransformerUI: add initial macOS UI

compnerd:
This is modelled after the Windows UI, and looks very similar.  This
actually helps demonstrate the point that the usage of the model is
extremely similar.  This sets an initial starting point for
improvements.

The biggest issue with the implementation currently is the `NSTextView`
does not update frequently enough resulting in the text appearing at the
end.

The differences from the Windows implementation include the fact that
`\n` does not need to be replaced with `\r\n`, the scrollview for the
textview needs to be constructed manually, and the coordinate system
uses a different origin.

### DIFF
--- a/Transformer/UI/macOS/main.swift
+++ b/Transformer/UI/macOS/main.swift
@@ -75,7 +75,7 @@ internal extension NSSlider {
 }
 
 func onMain(_ body: @escaping () -> ()) {
-  if NSThread.isMainThread {
+  if Thread.isMainThread {
     body()
   } else {
     DispatchQueue.main.async { body() }
@@ -171,9 +171,9 @@ extension SwiftApplicationDelegate {
     onBackground {
       for _ in 0 ..< 256 {
         do {
-          let word: String = gpt.generate()
+          let word: String = try gpt.generate()
           onMain {
-            try self.output.text = self.output.text + word
+            self.output.text = self.output.text + word
             let range: NSRange =
                 NSRange(location: self.output.text.count - 1, length: 1)
             self.output.scrollRangeToVisible(range)
--- a/Transformer/UI/macOS/main.swift
+++ b/Transformer/UI/macOS/main.swift
@@ -1,0 +1,191 @@
+// Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import AppKit
+
+import Dispatch
+import Foundation
+
+import TextModels
+
+internal class NSLabel: NSTextField {
+  public override init(frame: NSRect) {
+    super.init(frame: frame)
+
+    self.drawsBackground = false
+    self.isBezeled = false
+    self.isEditable = false
+    self.isSelectable = false
+  }
+
+  public required init?(coder: NSCoder) {
+    fatalError("init?(coder:) not implemented")
+  }
+}
+
+internal extension NSLabel {
+  convenience init(frame: NSRect, title: String) {
+    self.init(frame: frame)
+    self.stringValue = title
+  }
+}
+
+internal extension NSButton {
+  convenience init(frame: NSRect, title: String) {
+    self.init(frame: frame)
+    self.title = title
+  }
+
+  var text: String {
+    get { return self.stringValue }
+    set { self.stringValue = newValue }
+  }
+}
+
+internal extension NSTextField {
+  var text: String {
+    get { return self.stringValue }
+    set { self.stringValue = newValue }
+  }
+}
+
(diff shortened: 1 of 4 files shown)


compnerd:
CC: @texasmichelle @BradLarson

Transformer/UI/macOS/main.swift
```diff
@@ -0,0 +1,176 @@
...
+    self.output.font = ComicSansMS
+    self.output.isEditable = false
+
+    self.button.target = self
+    self.button.action = #selector(generate)
+
+    DispatchQueue.global(qos: .background).async {
+      do {
+        self.gpt = try GPT2()
+        self.label.text = "Transformer ready!"
```
BradLarson: This and other accesses to the label's text might need to be wrapped in an asynchronous dispatch on the main queue. The Main Thread Checker will probably warn you about these if run in Xcode.

Transformer/UI/macOS/main.swift
```diff
@@ -0,0 +1,176 @@
...
+    self.window.contentView?.addSubview(self.button)
+
+    let view: NSScrollView = NSScrollView(frame: self.output.frame)
+    view.hasVerticalScroller = true
+    view.hasHorizontalScroller = true
+    view.documentView = self.output
+    self.output.textContainer?.widthTracksTextView = false
+    self.window.contentView?.addSubview(view)
+
+    // self.window.contentView?.addSubview(self.output)
```
BradLarson: I'm guessing this is left over from testing?

compnerd: Thats left over from copy-pasting from Windows >.<  The need for the custom `NSScrollView` causes this.

Transformer/UI/macOS/main.swift
compnerd: Good catch.

BradLarson:
I think it would be really useful to have an Xcode project within the Transformer/UI/macOS directory, so that people could quickly open that and build the application. For the application bundle to work right, we might need to set up a Copy Files build phase to copy the appropriate libraries into the bundle and maybe a rewrite rule to get the paths set right.

If you'd like, I could create this and push that in.

Additionally, we'd probably want to update the Readme with both CMake and Xcode build instructions for the macOS UI application.

compnerd:
I think that setting it up with SPM and having that generate the Xcode project or just using cmake to generate the Xcode project would be nicer than adding all the Xcode project cruft.

I can update the instructions in a follow up change :)

Transformer/UI/macOS/main.swift
```diff
@@ -0,0 +1,189 @@
...
+
+internal extension NSSlider {
+  var value: Double {
+    get { return self.doubleValue }
+    set { self.doubleValue = newValue }
+  }
+}
+
+func onMain(_ body: @escaping () -> ()) {
+  if NSThread.isMainThread {
```
BradLarson:
```suggestion
  if Thread.isMainThread {
```
NSThread was deprecated in Swift 3, and this throws errors in an Xcode build.

Transformer/UI/macOS/main.swift
```diff
@@ -0,0 +1,189 @@
...
+    output.text = input.text
+    if !input.text.isEmpty {
+      gpt.seed = gpt.embedding(for: input.text)
+    }
+    gpt.temperature = Float(self.slider.value)
+
+    onBackground {
+      for _ in 0 ..< 256 {
+        do {
+          let word: String = gpt.generate()
```
BradLarson:
```suggestion
          let word: String = try gpt.generate()
```

Transformer/UI/macOS/main.swift
```diff
@@ -0,0 +1,189 @@
...
+          onMain {
+            try self.output.text = self.output.text + word
```
BradLarson:
```suggestion
            self.output.text = self.output.text + word
```
These two changes were required to align with how the GPT2 interface does error handling.

BradLarson:
Working with Tony, we were able to get SwiftPM to build and run a functional AppKit application from this. To do so, you can add to Package.swift a product of

```swift
        .executable(name: "TransformerMacOSUI", targets: ["TransformerMacOSUI"]),
```

and a target of

```swift
        .target(
            name: "TransformerMacOSUI", dependencies: ["Transformer"], path: "Transformer",
            exclude: ["UI/Windows/main.swift", "main.swift"],
            sources: ["UI/macOS/main.swift", "TransformerLoop.swift"],
            linkerSettings: [
                .unsafeFlags(
                    [
                        "-Xlinker", "-sectcreate", "-Xlinker", "__TEXT", "-Xlinker", "__info_plist",
                        "-Xlinker",
                        "./Transformer/UI/macOS/Info.plist",
                    ], .when(platforms: [.macOS]))
            ]),
```

Currently, there's a conflict in that TransformerLoop.swift is used both by the TransformerDemo and this  TransformerMacOSUI target. PR #363 should solve this when it is merged, and we might want to get that one in before this.

compnerd:
@BradLarson I think that the file being re-used is not a real issue - the file is already being used by two targets, this is creating a third.  None of them actually change the behaviour or contents of the file, merely add an additional reference to it.  I think that it should be safe to merge this without impacting the work that @texasmichelle is working on.

BradLarson:
@compnerd - Oh, by a conflict I mean that the same file is referred to as being part of two targets in SwiftPM, which the package manager throws errors about. If you add the above code in to insert this new target, you'll hit that error. Again, won't be a problem once that file is part of a shared library.

compnerd:
Ah, right, s-p-m cannot handle that case, CMake handles that just fine

PR #377, opened 2020-03-12
Adding Swift Package Manager target for macOS Transformer UI demo, update Readme

BradLarson:
Added a Swift Package Manager target for the new macOS Transformer UI demo. I also updated the Readme to show how to run this new demo, and made some changes to the text there to better reflect the current nature of the overall demo application.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -91,7 +91,7 @@ let package = Package(
                     [
                         "-Xlinker", "-sectcreate", "-Xlinker", "__TEXT", "-Xlinker", "__info_plist",
                         "-Xlinker",
-                        "./Transformer/UI/macOS/Info.plist",
+                        "Transformer/UI/macOS/Info.plist",
                     ], .when(platforms: [.macOS]))
             ]),
         .target(
--- a/Package.swift
+++ b/Package.swift
@@ -29,6 +29,7 @@ let package = Package(
         .library(name: "MiniGo", targets: ["MiniGo"]),
         .executable(name: "MiniGoDemo", targets: ["MiniGoDemo"]),
         .executable(name: "TransformerDemo", targets: ["TransformerDemo"]),
+        .executable(name: "TransformerUI", targets: ["TransformerUI"]),
         .executable(name: "GPT2-WikiText2", targets: ["GPT2-WikiText2"]),
     ],
     dependencies: [
@@ -81,6 +82,18 @@ let package = Package(
             path: "Transformer",
             exclude: ["UI/Windows/main.swift", "UI/macOS/main.swift"],
             sources: ["main.swift"]),
+        .target(
+            name: "TransformerUI", dependencies: ["TextModels"], path: "Transformer",
+            exclude: ["UI/Windows/main.swift", "main.swift"],
+            sources: ["UI/macOS/main.swift"],
+            linkerSettings: [
+                .unsafeFlags(
+                    [
+                        "-Xlinker", "-sectcreate", "-Xlinker", "__TEXT", "-Xlinker", "__info_plist",
+                        "-Xlinker",
+                        "./Transformer/UI/macOS/Info.plist",
+                    ], .when(platforms: [.macOS]))
+            ]),
         .target(
             name: "GPT2-WikiText2",
             dependencies: ["Batcher", "Datasets", "TextModels"],
--- a/Transformer/README.md
+++ b/Transformer/README.md
@@ -1,9 +1,10 @@
-# Transformer
+# GPT-2 Text Generation
 
-This is an implementation of [OpenAI's GPT-2 Transformer language model](https://github.com/openai/gpt-2) using [Swift for TensorFlow](https://github.com/tensorflow/swift).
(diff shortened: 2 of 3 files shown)


Package.swift
```diff
@@ -29,6 +29,7 @@ let package = Package(
         .library(name: "MiniGo", targets: ["MiniGo"]),
         .executable(name: "MiniGoDemo", targets: ["MiniGoDemo"]),
         .executable(name: "TransformerDemo", targets: ["TransformerDemo"]),
+        .executable(name: "TransformerUI", targets: ["TransformerUI"]),
```
compnerd: This needs to be restricted to macOS platforms, it will not be possible to build that target on Linux or Windows.

Package.swift
```diff
@@ -81,6 +82,18 @@ let package = Package(
...
+        .target(
+            name: "TransformerUI", dependencies: ["TextModels"], path: "Transformer",
+            exclude: ["UI/Windows/main.swift", "main.swift"],
+            sources: ["UI/macOS/main.swift"],
+            linkerSettings: [
+                .unsafeFlags(
+                    [
+                        "-Xlinker", "-sectcreate", "-Xlinker", "__TEXT", "-Xlinker", "__info_plist",
+                        "-Xlinker",
+                        "./Transformer/UI/macOS/Info.plist",
```
compnerd: is the `./` required?  Id really rather not have that, and have the path be `Transformer/UI/macOS/Info.plist` or better yet, the absolute path.  Making the command easy to replicate without `cd` is convenient for debugging

Transformer/UI/macOS/main.swift
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if canImport(AppKit)
```
compnerd: We don't do this in the Windows demo - alternatively, if you want to add both this *and* the Windows UI demo and build both somehow with s-p-m Im okay with that.  (That is, we should bump the s-p-m requirement and use the new platform conditions).

Package.swift
BradLarson: I've converted this to `Transformer/UI/macOS/Info.plist`. This works with `--package-path` when used outside of swift-models, so that seems like it should be good enough for general use.

Package.swift
texasmichelle: Does the `.when(platforms: [.macOS])` below refer to flags only?

Transformer/README.md
```diff
@@ -1,9 +1,10 @@
-# Transformer
+# GPT-2 Text Generation
 
-This is an implementation of [OpenAI's GPT-2 Transformer language model](https://github.com/openai/gpt-2) using [Swift for TensorFlow](https://github.com/tensorflow/swift).
+This is a demonstration of text generation using [OpenAI's GPT-2 Transformer language model](https://github.com/openai/gpt-2)
```
texasmichelle: s/using/with/ to avoid double "using" in one sentence.
Split across 2 lines to conform to 80-char limit.

Transformer/README.md
```diff
@@ -48,3 +49,12 @@ library search paths added (`-I` and `-L` with their respective values).
 
 Assuming that the generated DLLs are in the `PATH`, it should be possible to
 then execute the demo as `.\Transformer\TransformerUI.exe`.
+
+# macOS UI
+
+The macOS UI demo must be built and run from the root of the swift-models directory. You can run this example
```
texasmichelle: Potentially include minimum requirements/prereq's or at least state that this only works on macOS.

Package.swift
BradLarson: Yeah, I have that so the project will build cleanly on all platforms, only applying the AppKit-specific Info.plist inlining on macOS. Otherwise, the target itself still can be built on Linux and Windows, it just won't run there. We're checking to see if there's a way to selectively build on just macOS that doesn't involve rearranging the Package.swift to support the `#if os(macOS)` conditionals required.

Transformer/UI/macOS/main.swift
```diff
@@ -189,3 +190,7 @@ extension SwiftApplicationDelegate {
 let delegate: SwiftApplicationDelegate = SwiftApplicationDelegate()
 NSApplication.shared.delegate = delegate
 _ = NSApplicationMain(CommandLine.argc, CommandLine.unsafeArgv)
+
+#else
+fatalError("This demo requires AppKit on macOS to run.")
```
saeta:
```suggestion
#error("This demo requires AppKit on macOS to run.")
```

Just a silly idea. :-)

BradLarson: I'd intentionally wrapped this code to prevent a compile-time error on non-macOS platforms so that someone doing a `swift build` won't have it fail when it hits this file and tries to import AppKit. For example, CI will fail on presubmit if there's a compile error on non-macOS platforms here. Using a run-time error when you attempt to use this project on a non-macOS platform isn't great, but it was the best approach I could think of at the time.

saeta: Ah yes. That makes sense. Please ignore me. :-)

BradLarson:
I'm going to close this and we can do a standalone package for the macOS demo post-move from PR #413.

PR #378, opened 2020-03-12
add mobilenet v2

brettkoonce:


saeta:
Thank you @brettkoonce !! This is great.

Models/ImageClassification/MobileNetV2.swift
```diff
@@ -0,0 +1,181 @@
...
+    
+    public var invertedResidualBlock13 = InvertedBottleneckBlock(filters: (96, 160),
+        depthMultiplier: 6, strides: (2, 2))
+    public var invertedResidualBlock14 = InvertedBottleneckBlock(filters: (160, 160),
+        depthMultiplier: 6)
+    public var invertedResidualBlock15 = InvertedBottleneckBlock(filters: (160, 160),
+        depthMultiplier: 6)
+    
+    public var invertedResidualBlock16 = InvertedBottleneckBlock(filters: (160, 320),
+        depthMultiplier: 6, strides: (1, 1))
```
saeta: Out of curiosity, can we make these all collapse into an array of `InvertedBottleneckBlock`s? It might simplify both the model definition itself, as well as the `callAsFunction` implementation.

brettkoonce: yeah, i can probably reduce it a bit more, will have some free time in a bit!

texasmichelle:
This doesn't build on the latest nightly, but there's an upstream patch if you merge with master.

Models/ImageClassification/MobileNetV2.swift
```diff
@@ -0,0 +1,157 @@
...
+        } else {
+            return pwLinear
+        }
+    }
+}
+
+public struct InvertedResidualBlockStack: Layer {
+    var blocks: [InvertedResidualBlock] = []
+
+    public init(filters: (Int, Int), initialStrides: (Int, Int) = (2, 2), blockCount: Int) {
```
texasmichelle: Consider moving initialStrides to the end since it has a default value.

brettkoonce: tweaked!

brettkoonce:
@texasmichelle rebased + squashed atop master!

PR #379, opened 2020-03-12
Adds the ability to drop the last batch

### LINKED ISSUE #374: crash with batchSize = 2 and GPT2-WikiText2 demo
set batchSize to 2 in the demo, add following to loop over dataset:

```
import Batcher
import Datasets
import TensorFlow
import TextModels

var gpt = try GPT2()

let sequenceLength = gpt.contextSize
let batchSize = 2
let numWorkers = 8
// Use default WikiText2 dataset.
let dataset = TextUnsupervised(bpe: gpt.bpe, variant: .wikiText2, batchSize: batchSize,
    sequenceLength: sequenceLength)
let trainingBatcher = Batcher(
    on: dataset.trainingDataset, batchSize: batchSize, numWorkers: numWorkers, shuffle: true)
let validationBatcher = Batcher(
    on: dataset.validationDataset, batchSize: batchSize, numWorkers: numWorkers)

print("Dataset acquired.")

var optimizer = Adam(for: gpt.model, learningRate: 0.001)

for batch in trainingBatcher.sequenced() {
    let (documents, labels) = (batch.first, batch.second)
}

print("Starting training...")
```

run: `swift run GPT2-WikiText2`

Fatal error: Shapes of all inputs must match: values[0].shape = [1024] != values
[1].shape = [500]: file /home/skoonce/swift/swift-source/tensorflow-swift-apis/S
ources/TensorFlow/Bindings/EagerExecution.swift, line 300                       

works with bs = 1, something in https://github.com/tensorflow/swift-models/blob/b4c8b0e5b591bef13858cb89b4eec9d6de90bceb/Batcher/Batcher.swift#L108
and that the wikitext data isn't uniform is tripping this, looks like.

ping @sgugger

sgugger:
This PR adds a flag to optionally drop the last batch when it does not have `sequenceLength` elements. I think this will help fix #374 and in general be beneficial for training with fixed shapes on TPU.

### DIFF
--- a/Datasets/LanguageModelDataset.swift
+++ b/Datasets/LanguageModelDataset.swift
@@ -116,22 +116,24 @@ extension LanguageModelDataset: Collection {
 
 //Extension when Item is [Int] and openItem is not needed
 extension LanguageModelDataset where Item == [Int] {
-  public init(batchSize: Int, sequenceLength: Int, items: [Item], lengths: [Int]) {
+  public init(batchSize: Int, sequenceLength: Int, items: [Item], lengths: [Int], dropLast: Bool = false) {
     self.init(
       openItem: { $0 },
       batchSize: batchSize,
       sequenceLength: sequenceLength,
       items: items,
-      lengths: lengths)
+      lengths: lengths,
+      dropLast: dropLast)
   }
 
-  public init(batchSize: Int, sequenceLength: Int, items: [Item]) {
+  public init(batchSize: Int, sequenceLength: Int, items: [Item], dropLast: Bool = false) {
     self.init(
       openItem: { $0 },
       batchSize: batchSize,
       sequenceLength: sequenceLength,
       items: items,
-      lengths: items.map { $0.count })
+      lengths: items.map { $0.count },
+      dropLast: dropLast)
   }
 }
 
--- a/Datasets/LanguageModelDataset.swift
+++ b/Datasets/LanguageModelDataset.swift
@@ -15,6 +15,8 @@ public struct LanguageModelDataset<Item> {
   public let items: [Item]
   //The length of each processed item
   public let lengths: [Int]
+  //Drop the last batch if its length is less than sequenceLength
+  public let dropLast: Bool
   //The length of a contiguous chunk of text
   private var batchLength: Int
   //The number of batches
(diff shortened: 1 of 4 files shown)


Datasets/LanguageModelDataset.swift
```diff
@@ -31,15 +33,20 @@ public struct LanguageModelDataset<Item> {
...
   ) {
     self.openItem = openItem
     self.batchSize = batchSize
     self.sequenceLength = sequenceLength
     self.items = items
     self.lengths = lengths
+    self.dropLast = dropLast
     cumulativeLengths = lengths.reduce(into: []) { $0.append(($0.last ?? 0) + $1) }
     batchLength = (cumulativeLengths.last! - 1) / batchSize
+    if dropLast {
```
texasmichelle: This needs an additional check. If the last batch is the correct size, it shouldn't be dropped. This fix resolves the fatal error in the GPT-2 example where the last batch is unequal to the rest (batchSize: 2), but a batch size of 64 does not return any data.

Datasets/LanguageModelDataset.swift
```diff
@@ -31,15 +33,20 @@ public struct LanguageModelDataset<Item> {
     batchSize: Int,
```
texasmichelle: I find this variable name confusing. Based on its behavior, it seems to be the number of batches - is that understanding correct? Combined with the batchLength variable, this struct is difficult to understand.

Datasets/LanguageModelDataset.swift
```diff
@@ -31,15 +33,20 @@ public struct LanguageModelDataset<Item> {
...
     sequenceLength: Int,
     items: [Item],
-    lengths: [Int]
+    lengths: [Int],
+    dropLast: Bool = false
```
texasmichelle: There are two other inits at the bottom of this file, which I would like to see this added to.

Datasets/LanguageModelDataset.swift
sgugger: Yes, if the total length is a multiple of `sequenceLength`, the code below doesn't change anything. 

> but a batch size of 64 does not return any data

Could you expand on that? What do you mean?

Datasets/LanguageModelDataset.swift
sgugger: It is the size of the batch, e.g. your batches will be of size `batchSize x sequenceLength`

To build the batches, the struct internally concatenates all texts together (maybe with a shuffle) then makes a big array of size `batchSize` by `batchLength` (not for real to avoid memory issue, but it's a good mental model). The it goes through that array by spitting tensors of sizes  `batchSize` by `sequenceLength`. This fix makes sure  `batchLength` is a round mutliple of `sequenceLength` when `dropLast` is true.

Datasets/LanguageModelDataset.swift
sgugger: Thanks for reminding me they exist, they are removed by #355 but that PR is blocked for some reason.

Datasets/LanguageModelDataset.swift
texasmichelle: As discussed, this was a result of using a smaller validation set than training. Using independent batch sizes should resolve this.

Datasets/LanguageModelDataset.swift
texasmichelle: I see what you mean. I think this could be communicated a bit clearer, but works for now. It would be great to cover this in an api review, so I'll add a label.

PR #383, opened 2020-03-12
GPT2: Change conformance for TransformerLM and EmbeddingGPT from Differentiable to Module

texasmichelle:


saeta:
Note: I think this might mean we want to revert the (now redundant) explicit `KeyPathIterable` conformance from https://github.com/tensorflow/swift-models/pull/382 if this is a good and proper fix. CC @compnerd

texasmichelle:
I can do that as part of this PR

PR #390, opened 2020-03-13
Pad sequences to maxSequenceLength for BERT

xihui-wu:
This will avoid dynamic shape and enable training well on TPU.
Take Bert python implementation for reference: https://github.com/google-research/bert/blob/master/run_classifier.py#L449

### DIFF
--- a/Models/Text/BERT.swift
+++ b/Models/Text/BERT.swift
@@ -279,11 +279,18 @@ public struct BERT: Module, Regularizable {
                 tokenTypeIds.append(Int32(sequenceId))
             }
         }
-        let tokenIds = tokens.map { Int32(vocabulary.id(forToken: $0)!) }
+        var tokenIds = tokens.map { Int32(vocabulary.id(forToken: $0)!) }
 
         // The mask is set to `true` for real tokens and `false` for padding tokens. This is so
         // that only real tokens are attended to.
-        let mask = [Int32](repeating: 1, count: tokenIds.count)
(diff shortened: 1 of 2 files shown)


Models/Text/BERT.swift
```diff
@@ -279,11 +279,18 @@ public struct BERT: Module, Regularizable {
...
-        let tokenIds = tokens.map { Int32(vocabulary.id(forToken: $0)!) }
+        var tokenIds = tokens.map { Int32(vocabulary.id(forToken: $0)!) }
 
         // The mask is set to `true` for real tokens and `false` for padding tokens. This is so
         // that only real tokens are attended to.
-        let mask = [Int32](repeating: 1, count: tokenIds.count)
+        var mask = [Int32](repeating: 1, count: tokenIds.count)
+
+        // Zero-pad up to the max sequence length.
+        while tokenIds.count < maxSequenceLength {
```
BradLarson: Rather than a while loop here, would you be better off by appending an array of `[Int32](repeating: 0, count: maxSequenceLength 
 - tokenIds.count)` (created once, before `tokenIds` is modified) to each of the three arrays?

xihui-wu: Got it.

xihui-wu:
With further investigation, this fix is not needed, whereas a different fix is made here: https://github.com/tensorflow/swift-models/pull/399.

xihui-wu:
With further investigation, this fix is not needed, whereas a different fix is made here: https://github.com/tensorflow/swift-models/pull/399.

PR #394, opened 2020-03-14
add efficientnet

brettkoonce:


Models/ImageClassification/EfficientNet.swift
```diff
@@ -0,0 +1,196 @@
...
+    public init(
+        filters: (Int, Int),
+        initialStrides: (Int, Int) = (2, 2),
+        kernel: (Int, Int) = (3, 3),
+        blockCount: Int
+    ) {
+        self.blocks = [MBConvBlock(filters: (filters.0, filters.1),
+            strides: initialStrides, kernel: kernel)]
+        for _ in 1..<blockCount {
+            self.blocks += [MBConvBlock(filters: (filters.1, filters.1), kernel: kernel)]
```
saeta:
```suggestion
            self.blocks.append(MBConvBlock(filters: (filters.1, filters.1), kernel: kernel))
```

Models/ImageClassification/EfficientNet.swift
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+    @differentiable
+    public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+        return blocks.differentiableReduce(input) { $1($0) }
+    }
+}
+
+public struct EfficientNet: Layer {
```
saeta: IIUC, `EfficientNet` is actually a family of model architectures in the same way that `ResNet` is also a family of network architectures. Do you know what the specific variation of this EfficientNet is? And how hard might it be to parameterize this struct based on an enum specifying the variations?

Models/ImageClassification/EfficientNet.swift
```diff
@@ -0,0 +1,196 @@
...
+            filterShape: (3, 3, 3, 32),
+            strides: (2, 2),
+            padding: .same)
+        inputConvBatchNorm = BatchNorm(featureCount: 32)
+
+        finalConv = Conv2D<Float>(
+            filterShape: (1, 1, 320, 1280),
+            strides: (1, 1),
+            padding: .same)
+        avgPool = GlobalAvgPool2D()
```
saeta: Do you still need this initializer? (Looks like you initialize `dropout` on line `170` already.)

Models/ImageClassification/EfficientNet.swift
```diff
@@ -0,0 +1,196 @@
...
+    public var residualBlockStack1 = MBConvBlockStack(filters: (16, 24), blockCount: 2)
+    public var residualBlockStack2 = MBConvBlockStack(filters: (24, 40), kernel: (5, 5),
+        blockCount: 2)
+    public var residualBlockStack3 = MBConvBlockStack(filters: (40, 80), blockCount: 3)
+    public var residualBlockStack4 = MBConvBlockStack(filters: (80, 112), initialStrides: (1, 1),
+        kernel: (5, 5), blockCount: 3)
+    public var residualBlockStack5 = MBConvBlockStack(filters: (112, 192), kernel: (5, 5),
+        blockCount: 4)
+    public var residualBlockStack6 = MBConvBlockStack(filters: (192, 320), initialStrides: (1, 1),
+        blockCount: 1)
```
saeta: What about storing all the `MBConvBlockStack`'s inside an array (or ideally a `LayerArray` --- see below)? For example:

```suggestion
    public var backbone = LayerArray([
        MBConvBlockStack(filters: (16, 24), blockCount: 2),
        MBConvBlockStack(filters: (24, 40), kernel: (5, 5), blockCount: 2),
        MBConvBlockStack(filters: (40, 80), blockCount: 3),
        MBConvBlockStack(filters: (80, 112), initialStrides: (1, 1), kernel: (5, 5), blockCount: 3),
        MBConvBlockStack(filters: (112, 192), kernel: (5, 5), blockCount: 4),
        MBConvBlockStack(filters: (192, 320), initialStrides: (1, 1), blockCount: 1)
    ])
```

Note: I think we'd probably want to define a proper type for a LayerArray:

```swift
struct LayerArray<T: Layer>: Layer where T.Input == T.Output {
  var layers = [T]
  @differentiable
  func callAsFunction(_ input: T.Input) -> T.Output {
    layers.differentiableReduce(input) { $1($0) }
  }
}
```

Context: Based on some internal discussion related to https://github.com/tensorflow/swift-apis/issues/699 we thought that it might be better to have a wrapper type instead of conforming `Array` to `Layer` (i.e. `LayerArray` above). Better names wanted for `LayerArray`. CC @dabrahams

Models/ImageClassification/EfficientNet.swift
brettkoonce: no, removed!

Models/ImageClassification/EfficientNet.swift
brettkoonce: tweaked!

Models/ImageClassification/EfficientNet.swift
brettkoonce: you understand correctly, this is the b0 (baseline) variant.  the actual parameterization isn't [too bad](https://github.com/tensorflow/tpu/blob/d6f2ef3edfeb4b1c2039b81014dc5271a7753832/models/official/efficientnet/efficientnet_builder.py#L37), but the logic to get all the filter resizing/resolutions correct is a bit more tricky.

Models/ImageClassification/EfficientNet.swift
brettkoonce: like the syntax better. LayerArray would have to be added upstream first, right?

Models/ImageClassification/EfficientNet.swift
brettkoonce: @saeta see updated!

s1ddok:
Wow! I've been implemented this myself and now realized there is an implementation already! Do you have any plans on training side?

brettkoonce:
if you'd like to play with this, i've just been dropping into the imagenette demos, works fine there.  long term yes, would like to be able to train this from scratch.

s1ddok:
@brettkoonce 

I'm getting this kind of error during training:

```
Precondition failed: Count mismatch: 3 and 0: file /swift-base/tensorflow-swift-apis/Sources/TensorFlow/StdlibExtensions.swift, line 272
Current stack trace:
0    libswiftCore.so                    0x00007faf0d2fe910 swift_reportError + 50
1    libswiftCore.so                    0x00007faf0d36fb90 _swift_stdlib_reportFatalErrorInFile + 115
2    libswiftCore.so                    0x00007faf0d04fe31 <unavailable> + 1461809
3    libswiftCore.so                    0x00007faf0d04fa77 <unavailable> + 1460855
4    libswiftCore.so                    0x00007faf0d050012 <unavailable> + 1462290
5    libswiftCore.so                    0x00007faf0d04e460 _assertionFailure(_:_:file:line:flags:) + 517
6    libswiftTensorFlow.so              0x00007faf087144b0 static Array.DifferentiableView<A>..* infix(_:_:) + 977
Current stack trace:
	frame #10: 0x00007faea404f083 $__lldb_expr66`main at <Cell 8>:22:19
```

Have you seen anything like that?

brettkoonce:
var model = VGG16(classCount: 10)
let optimizer = AMSGrad(for: model)

produces a crash as well, so it's a bug with how differentiable layers are being passed through.

brettkoonce:
@s1ddok https://github.com/tensorflow/swift-apis/blob/3ebd1765480f6aeb15046b6766741009ac32f84a/Sources/TensorFlow/Optimizers/MomentumBased.swift#L414 would be my guess

brettkoonce:
https://github.com/tensorflow/swift-apis/blob/3ebd1765480f6aeb15046b6766741009ac32f84a/Sources/TensorFlow/Optimizers/MomentumBased.swift#L506 == the same
both adamax and adagrad fail

Models/ImageClassification/EfficientNet.swift
dabrahams: FYI, the name `LayerArray` implies the abstraction being created is an array, whereas `: Layer` indicates that it is a `Layer`.  Also please see https://github.com/tensorflow/swift-apis/pull/673#discussion_r394652982

dabrahams: See also tensorflow/swift-apis#708

brettkoonce: guess what i meant to say is that i would be happy to update this in a followup pr once the above is merged upstream.

brettkoonce:
still needs some filter resize logic tweaks!

brettkoonce:
working with lower learning rates here!

brettkoonce:
above is ready to go, in case it's not clear!

brettkoonce:
ready to go!

Models/ImageClassification/EfficientNet.swift
```diff
@@ -0,0 +1,328 @@
...
+    public var residualBlockStack3: MBConvBlockStack
+    public var residualBlockStack4: MBConvBlockStack
+    public var residualBlockStack5: MBConvBlockStack
+    public var residualBlockStack6: MBConvBlockStack
+
+    public var outputConv: Conv2D<Float>
+    public var outputConvBatchNorm: BatchNorm<Float>
+    public var avgPool = GlobalAvgPool2D<Float>()
+    public var dropoutProb: Dropout<Float>
+    public var outputClassifier: Dense<Float>
```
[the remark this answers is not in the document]
brettkoonce: see updated!  would this be the preferred style going forward?

Issue #389, opened 2020-03-13
Remove use of deprecated substring


texasmichelle:
The build produces a warning about the use of substring, which was merged as a workaround for the macOS build in #375 .

Replace this with a less-confusing, non-deprecated statement.

```
$ swift build
/home/michellecasbon/repos/tensorflow/texasmichelle/swift-models/Datasets/TextUnsupervised/TextUnsupervised.swift:134:34: warning: 'substring(to:)' is deprecated: Please use String slicing subscript with a 'partial range upto' operator.
            rows[rows.count - 1].substring(to: rows[rows.count - 1].index(rows[rows.count - 1].endIndex, offsetBy: -2)))
                                 ^
```

BradLarson:
Fixed by PR #395.

PR #395, opened 2020-03-14
replaced deprecated substring

SumanSudhir:
Replaced `substring(to:)` [here](https://github.com/tensorflow/swift-models/blob/eb3efbb7b3e831074ce6549ec2ba40c4d23c9690/Datasets/TextUnsupervised/TextUnsupervised.swift#L134) as it is showing warning: 'substring(to:)' is deprecated: Please use String slicing subscript with a 'partial range upto' operator.


SumanSudhir:
#389 @texasmichelle I saw this issue after making PR. Sorry if I have duplicated your work

texasmichelle:
The issue was there so that someone could fix it :) Thank you for this submission @SumanSudhir !!

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -129,9 +129,8 @@ public struct TextUnsupervised {
         var rows = rawText.components(separatedBy: "\"\n\"")
         // Removing the initial '"'.
         rows[0] = String(rows[0].dropFirst())
-        // Removing the last '"\n'.
-        rows[rows.count - 1] = String(
-            rows[rows.count - 1].substring(to: rows[rows.count - 1].index(rows[rows.count - 1].endIndex, offsetBy: -2)))
+        // Removing the last '"\n'
+        rows[rows.count - 1] = String(rows[rows.count - 1].suffix(2))
```
texasmichelle: This command does in fact remove the substring-related warning, but it removes everything *except* the last two characters. We're looking for a command that keeps everything except the trailing newline and quotation mark.

Examples/Notebooks/LanguageModel.ipynb
```diff
@@ -1,799 +1,1094 @@
 {
- "cells": [
```
texasmichelle: This diff makes it hard to see what has changed, which I believe should only be a single cell. I suggest doing a text replace, then running separately to ensure it's correct. @sgugger do you have a better process for making notebook changes?

sgugger: Ideally, just making the change in the cell is better to avoid large diffs. Otherwise, [reviewNB](https://www.reviewnb.com/) is great to see the diff on notebooks.

SumanSudhir:
@texasmichelle now I have made changes in only one line and have not run any cell still unable to get a small difference.

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -129,9 +129,8 @@ public struct TextUnsupervised {
         var rows = rawText.components(separatedBy: "\"\n\"")
         // Removing the initial '"'.
         rows[0] = String(rows[0].dropFirst())
-        // Removing the last '"\n'.
-        rows[rows.count - 1] = String(
-            rows[rows.count - 1].substring(to: rows[rows.count - 1].index(rows[rows.count - 1].endIndex, offsetBy: -2)))
+        // Removing the last '"\n'
+        rows[rows.count - 1] = String(rows[rows.count - 1].prefix(rows[rows.count - 1].count - 2))
```
dabrahams: Generally speaking, you don't want to access the `count` of a `String` because that's an O(N) operation, and there are lots of ways to avoid it here, with cleaner code. Here are couple of options:

```
rows[rows.indices.last!] = String(rows.last!.dropLast(2))
```

or

```
let lastIndex = rows.indices.last!
rows[lastIndex] = String(rows[lastIndex].dropLast(2))
```

PR #398, opened 2020-03-17
[DNM] build: copy the Protobuf dependency for TransformerUI

compnerd:
Automate the copying of `libswiftProtobuf.dylib` on macOS builds.  This
allows us to execute the binary from build tree without any additional
work.


compnerd:
This really should not be needed.  The `LC_RPATH` is setup for the build tree properly.  I think that something else is going on.

compnerd:
@BradLarson I believe that this ended up being unnecessary correct?  If so, I'd like to close off this PR.

PR #399, opened 2020-03-17
Drop last smaller batch in Cola-BERT

xihui-wu:
When training models on TPU, dynamic shape can cause recompilation overhead when input shape changes between steps. In Cola-BERT this can happen in Embedding layer where sequence length as well as batch size may vary if don't handle them carefully.

To prevent sequence length varying, we should pad sequences to maximum length. And we already did it by calling padAndBatch function [here](https://github.com/tensorflow/swift-models/blob/master/Datasets/CoLA/CoLA.swift#L168)

To prevent varying batch size we should drop the last batch where there aren't enough examples in there. And this PR is doing this.

Here is a brief explanation on how varying batch size happens: 
1. Our Cola data set has 8551 training examples and 1043 dev examples and we use batch 8 (1024/128). Apparently 8551=8x1068+7, 1043=8x130+3
2. Each step we process 8(a batch) training examples, then at 1069th step we only process 7 examples if we keep the last batch. Whereas each 10 steps we run an evaluation through ALL the dev dataset, thus at the end of the evaluation we are doomed to encounter the 3-examples batch. Hence, the recompilation happens every 10 steps aside of the very last step (1068th step)

### DIFF
--- a/Datasets/CoLA/DataUtilities.swift
+++ b/Datasets/CoLA/DataUtilities.swift
@@ -210,9 +210,7 @@ public struct GroupedIterator<Base: IteratorProtocol>: IteratorProtocol {
       }
     }
     guard let elementsToReduce = elements else {
-      if dropRemainder {
-        return nil
-      }
+      if dropRemainder { return nil }
       if currentGroup == nil { currentGroup = groups.values.startIndex }
       if currentGroup! >= groups.values.endIndex { return nil }
       while groups.values[currentGroup!].isEmpty {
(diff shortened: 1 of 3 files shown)


Datasets/CoLA/DataUtilities.swift
```diff
@@ -206,6 +210,9 @@ public struct GroupedIterator<Base: IteratorProtocol>: IteratorProtocol {
       }
     }
     guard let elementsToReduce = elements else {
+      if dropRemainder {
```
marcrasi: Could you put this all on one line, for consistency with the other short if statements in this function?

PR #400, opened 2020-03-17
CycleGan

s1ddok:
CycleGAN model reproduced in S4TF to achieve similar results with original paper


s1ddok:
I would love to hear feedbacks on how to optimize this thing. I've replicated hyper-parameters and model architectures from original repo, but it trains 10x slower than Python equivalent. I suspect I have some major implementation flaws here.

brettkoonce:
cool demo!  i wouldn't discount your efforts too much, this is still the early days with respect to optimization.

s1ddok:
@brettkoonce thanks! 

I've tried to abandon `TensorFlow.gradient(at:)` approach in favor of using manual back-propagation, but failed to combine multiple losses, hence I have quite a few superfluous model runs here compared to Python implementation. 

One thing that concerns me the most is that GPU utils never goes higher than 50% even though the dataset fits in RAM and augmentation is done on GPU as well. 

I also have a multi-gpu examples which trains almost 2x faster but unfortunately crashes eventually with weird errors. I suspect S4TF doesn't support concurrent model runs as of now.

brettkoonce:
my only suggestion would be to add a README with some pictures that explain how to run the model on a basic dataset to make it easier for people to understand what is going on.

s1ddok:
yup, this is on my todo-list. i've just finished the implementation and made sure it produces comparable results to what original authors had

s1ddok:
50 epochs are over, yay! Original authors trained for 200 epochs but that's too long for me. (50 epochs took around 2 days). I've got some very promising and close to original paper results:



Gonna finalize the readme part now. 

P.S. @BradLarson I'm finalizing this project for myself, so I would love to see your suggestions and hopefully get it merged :)

CycleGAN/main.swift
```diff
@@ -0,0 +1,224 @@
...
+        // MARK: Test
+
+        Context.local.learningPhase = .training
+        let realX = batch.first.image
+        let realY = batch.second.image
+
+        // we do it outside of GPU scope so that dataset shuffling happens on CPU side
+        let concatanatedImages = realX.concatenated(with: realY)
+
+        withDevice(.gpu, gpuIndex) {
```
BradLarson: The `withDevice()` API is incompatible with the newer x10 backend (that should provide significantly improved performance with this and other models), so this will probably need to be removed here and below. This isn't needed for GPU operation anyway, and we'll have a different means for explicit device placement in multi-GPU systems soon.

Package.swift
```diff
@@ -30,10 +30,14 @@ let package = Package(
         .executable(name: "MiniGoDemo", targets: ["MiniGoDemo"]),
         .executable(name: "TransformerDemo", targets: ["TransformerDemo"]),
         .executable(name: "GPT2-WikiText2", targets: ["GPT2-WikiText2"]),
+        .executable(name: "CycleGAN", targets: ["CycleGAN"])
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.7.0"),
+        .package(name: "SwiftProtobuf", url: "https://github.com/apple/swift-protobuf.git", from: "1.7.0"),
```
BradLarson: Is there a reason why this is assigned an explicit name?

CycleGAN/Models/Layers.swift
```diff
@@ -0,0 +1,200 @@
...
+    ///
+    /// - Parameter input: The input to the layer.
+    /// - Returns: The output.
+    @differentiable
+    public func callAsFunction(_ input: Input) -> Output {
+        return input.sequenced(through: pad, conv2d)
+    }
+}
+
+public struct ResnetBlock<NT: FeatureChannelInitializable>: Layer where NT.TangentVector.VectorSpaceScalar == Float, NT.Input == Tensorf, NT.Output == Tensorf {
```
BradLarson: Minor naming suggestion: ResNet always has the N capitalized in the usages I've seen. Maybe `ResNetBlock` and `ResNetGenerator`?

CycleGAN/main.swift
```diff
@@ -0,0 +1,224 @@
...
+            let scaledImages = _Raw.resizeNearestNeighbor(images: concatanatedImages, size: [286, 286])
```
BradLarson: We don't want to introduce new `_Raw` usage in models, so this can be replaced by the new `resize()` function: https://github.com/tensorflow/swift-apis/pull/706/files

CycleGAN/main.swift
```diff
@@ -0,0 +1,224 @@
...
+            var croppedImages = scaledImages.slice(lowerBounds: Tensor<Int32>([0, Int32(random() % 30), Int32(random() % 30), 0]),
+                                                   sizes: [2, 256, 256, 3])
+            if random() % 2 == 0 {
+                croppedImages = _Raw.reverse(croppedImages, dims: [false, false, true, false])
```
BradLarson: Again, we don't want to introduce new `_Raw` usage, so I believe this flipping can be performed without needing an explicit accelerated operation. If one is needed, I believe a negative scale along the X axis on a scale and translate operation can simulate this.

Package.swift
s1ddok: for some reason it just won't compile otherwise. I just followed the instruction from package manager error

CycleGAN/main.swift
s1ddok: fixed using https://github.com/tensorflow/swift-apis/commit/0738ff77ce4029b4b57394e5b7868b5fbffa85ba, now it requires a nightly toolchain in order to compile

CycleGAN/main.swift
s1ddok: fixed

CycleGAN/Models/Layers.swift
s1ddok: fixed

CycleGAN/main.swift
s1ddok: fixed

s1ddok:
@BradLarson I've tried to fix everything. Should I do anything else for this to be merged?

CycleGAN/Models/Generator.swift
```diff
@@ -0,0 +1,133 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import TensorFlow
+
+public struct ResNetGenerator<NT: FeatureChannelInitializable>: Layer where NT.TangentVector.VectorSpaceScalar == Float, NT.Input == Tensorf, NT.Output == Tensorf {
```
BradLarson: Does ResNetGenerator need to be generic about its normalization? I only see it being used with InstanceNorm2D here. Are there cases where that won't be true?

If it should remain generic, `NT` is difficult to parse as a name, maybe it could be changed to something indicating this is the normalization method.

CycleGAN/Models/Generator.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    var lastConv: Conv2D<Float>
+
+    public init(inputChannels: Int,
+                outputChannels: Int,
+                blocks: Int,
+                ngf: Int,
+                normalization: NT.Type,
+                useDropout: Bool = false) {
+        norm1 = .init(featureCount: ngf)
```
BradLarson: Generally, we don't call initializers in this manner. I'd recommend changing the `.init` calls to the type name, to be more consistent with the style we use.

CycleGAN/Models/Layers.swift
```diff
@@ -0,0 +1,200 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import TensorFlow
+
+public struct Identity: ParameterlessLayer {
```
BradLarson: Is this layer type used in this example?

CycleGAN/Models/Layers.swift
```diff
@@ -0,0 +1,200 @@
...
+    @differentiable
+    public func callAsFunction(_ input: Tensorf) -> Tensorf {
+        input
+    }
+}
+
+public struct LeakyRELU: ParameterlessLayer {
```
BradLarson: Rather than creating a new custom layer type, if you need to wrap `leakyRelu()` for use in Sequential, you can use it with [a Function Layer type](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Layers/Core.swift#L70).

CycleGAN/Models/Generator.swift
s1ddok: I think It is generally useful. This model can be reused with batch norm for example. NT stays for Normalization Type, haha. I will rename it!

CycleGAN/Models/Layers.swift
s1ddok: Oh yeah this is a leftover from pix2pix recursive U-Net, I will remove it

CycleGAN/Models/Layers.swift
s1ddok: cool, didn't know that one exists, thanks!

s1ddok:
@BradLarson fixed!

Package.swift
```diff
@@ -30,10 +30,13 @@ let package = Package(
         .executable(name: "MiniGoDemo", targets: ["MiniGoDemo"]),
         .executable(name: "GPT2-Inference", targets: ["GPT2-Inference"]),
         .executable(name: "GPT2-WikiText2", targets: ["GPT2-WikiText2"]),
+        .executable(name: "CycleGAN", targets: ["CycleGAN"])
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.7.0"),
+        .package(name: "SwiftProtobuf", url: "https://github.com/apple/swift-protobuf.git", from: "1.7.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.1")),
+        .package(url: "https://github.com/JohnSundell/Files", from: "4.0.0"),
```
dabrahams: We're concerned about introducing this dependency for all the models.  Can the functionality be implemented without it?

s1ddok: I think it makes it elegant in a few places, but I can drop it of course if that’s a problem

dabrahams: Yes, please!

s1ddok:
@BradLarson @dabrahams I dropped Files dependency, went Foundation-only. Code become a little bit more clumsy, but I think it's still okay. 

I've also used comments from @Shashi456 in #365 and dropped custom `ZeroPad` implementation in favor of `ZeroPadding2D` from `swift-api`s.

s1ddok:
@BradLarson @dabrahams friendly ping from my side now :) please tell If I can do anything about this PR

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.2
```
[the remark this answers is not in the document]
s1ddok: I couldn't generate proper xcode project without this one

Package.swift
```diff
@@ -30,10 +30,12 @@ let package = Package(
         .executable(name: "MiniGoDemo", targets: ["MiniGoDemo"]),
         .executable(name: "GPT2-Inference", targets: ["GPT2-Inference"]),
         .executable(name: "GPT2-WikiText2", targets: ["GPT2-WikiText2"]),
+        .executable(name: "CycleGAN", targets: ["CycleGAN"])
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.7.0"),
+        .package(name: "SwiftProtobuf", url: "https://github.com/apple/swift-protobuf.git", from: "1.7.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.1")),
+        .package(name: "TensorBoardX", url: "https://github.com/t-ae/tensorboardx-s4tf.git", from: "0.1.2"),
```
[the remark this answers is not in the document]
BradLarson: With other models examples, we've been printing out intermediate results to the console on every epoch: https://github.com/tensorflow/swift-models/blob/master/Examples/MobileNetV2-Imagenette/main.swift#L65 . That might be good enough for now, and then when we have better logging facilities and TensorBoard integration we can update all of these with that. We're also working towards having a common training loop, which could have logging as one of its callbacks so that we'd be able to update all of these quickly (and allow for easy multi-device training).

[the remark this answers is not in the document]
s1ddok: by intermediate results I mean model progress on a one particular image, not loss/metric values

Package.swift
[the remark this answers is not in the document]
BradLarson: Huh, weird. I've been using this with various Xcodes, up to 11.3.1, and haven't seen a problem with this package in those. Are you seeing this with 11.4?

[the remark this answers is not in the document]
s1ddok: Yup, I will try to downgrade it back, but it seems like having it on 5.2 is okay since S4TF toolchain is itself 5.2

Package.swift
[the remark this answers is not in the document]
BradLarson: Oh, I don't see anything wrong with saving out intermediate images to a directory of some sort. We do that with the autoencoder examples, providing a sample image per epoch to watch the evolution of the model.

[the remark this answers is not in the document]
s1ddok: okay, I will introduce it then, because it is now done through TensorBoard-X which is to be removed

s1ddok:
@BradLarson I think I fixed everything you've mentioned

PR #412, opened 2020-03-23
GPT-2: Change absolute paths to relative in Windows build instructions

texasmichelle:
Build instructions with absolute paths do not work for all installations.

Add a reminder to copy installation files after every VS update.

### DIFF
--- a/Transformer/README.md
+++ b/Transformer/README.md
@@ -34,41 +34,49 @@ Although all the models build and run, not all of them have been tested.  Partic
 
 #### Configure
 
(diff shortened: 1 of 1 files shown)


texasmichelle:
@compnerd I noticed absolute paths in the root [README.md](https://github.com/tensorflow/swift-models/blob/master/README.md#experimental-cmake-support), which makes me think the right fix was just to add a drive letter. I'll test it to make sure, then open a new PR so that we have consistency across READMEs.

PR #414, opened 2020-03-24
Refactored FastStyleTransfer ColabDemo

### LINKED ISSUE #410: Update FastStyleTransfer ColabDemo
May I update [ColabDemo](https://github.com/tensorflow/swift-models/blob/master/FastStyleTransfer/Demo/ColabDemo.ipynb) notebook with following changes?

- [ ]  'open in colab' link still points to  @vvmnnnkv's colab notebook which is not up-to-date with recent changes such as integrating `PythonKit` and new `CheckpointReader`; So I guess it should be changed to an updated one.

 ~The 2nd cell of notebook holds both conditional as well as normal swift code which is not possible to execute as a single cell, at least in colab and causes error~:
~Solution would be to split the code into 3 cells.~ 
**EDIT:** There's no problem keeping both swift code and conditionals in same cell

- [x] Minor refactoring such as bringing `let plt...` on newline, adjusting indentations, ~adding comment on the first line of conditional block cells as discussed [here](https://github.com/tensorflow/swift/issues/405#issuecomment-598013192)~ 
**EDIT:** Required only if conditionals are the first piece of code in a cell
- [ ] Optionally, we could clip `img` tensor accordingly to avoid warnings from matplotlib

Also, is it necessary to keep hacky layer mappings in the colab notebook? we could move the [Helpers.swift](https://github.com/tensorflow/swift-models/blob/master/FastStyleTransfer/Demo/Helpers.swift) in the Utility package of FastStyleTransfer and call the `importWeights` method (providing its source code link for additional reference)

I've done said changes in this colab [notebook](https://colab.research.google.com/gist/Kshitij09/90aadb408f8d7f4661fa90cc1a103312/fast_style_transfer_with_s4tf.ipynb)

Kshitij09:
As discussed in #410, Refactored the ColabDemo notebook with:
 - appropriate indentation for conditional block
 - moved `plt` declaration to newline
 - grouped matplotlib related code under single comment.

Fixes #410 

There are few more suggestions I made in the issue, awaiting for your feedback on them.

### DIFF
--- a/FastStyleTransfer/Demo/ColabDemo.ipynb
+++ b/FastStyleTransfer/Demo/ColabDemo.ipynb
@@ -74,13 +74,15 @@
     "import FastStyleTransfer\n",
     "import ModelSupport\n",
     "\n",
-    "// Imaging capabilities\n",
-    "%include \"EnableIPythonDisplay.swift\"\n",
+    "\n",
     "#if canImport(PythonKit)\n",
     "    import PythonKit\n",
-    "#else",
+    "#else\n",
     "    import Python\n",
-    "#endif",
+    "#endif\n",
+    "\n",
+    "// Imaging capabilities\n",
+    "%include \"EnableIPythonDisplay.swift\"\n",
     "let plt = Python.import(\"matplotlib.pyplot\")\n",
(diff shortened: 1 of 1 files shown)


Kshitij09:
@googlebot I signed it!

PR #417, opened 2020-03-24
Preventing Datasets from re-downloading.

### LINKED ISSUE #381: Datasets should not be re-downloaded on each run of a test
Originally, our intent with the dataset unit tests was to test them from a common starting point each run. That meant manually deleting any locally cached versions of the dataset and forcing a re-download. While that is a noble goal, in practice this is slowing down model tests and using a lot of unnecessary bandwidth.

Therefore, we've decided that it's fine to re-use cached local data for these tests and the code that deletes the locally cached copies of the datasets should be removed from our dataset unit tests.

SaiHemanthBR:
fixes #381

All datasets will be downloaded into "~/.s4tf/datasets" and not into a temporary directory like before.

### DIFF
--- a/Datasets/DatasetUtilities.swift
+++ b/Datasets/DatasetUtilities.swift
@@ -23,8 +23,8 @@ public enum DatasetUtilities {
     public static let currentWorkingDirectoryURL = URL(
         fileURLWithPath: FileManager.default.currentDirectoryPath)
         
-    public static let defaultDirectory = FileManager.default.urls(
-            for: .applicationSupportDirectory, in: .userDomainMask).first!
+    public static let defaultDirectory = try! FileManager.default.url(
+            for: .cachesDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
             .appendingPathComponent("swift-models").appendingPathComponent("datasets")
 
     @discardableResult
--- a/Datasets/DatasetUtilities.swift
+++ b/Datasets/DatasetUtilities.swift
@@ -22,6 +22,9 @@ import ModelSupport
 public enum DatasetUtilities {
     public static let currentWorkingDirectoryURL = URL(
         fileURLWithPath: FileManager.default.currentDirectoryPath)
+    
+    public static let defaultDirectory = FileManager.default.homeDirectoryForCurrentUser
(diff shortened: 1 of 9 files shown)


Datasets/DatasetUtilities.swift
```diff
@@ -22,6 +22,9 @@ import ModelSupport
 public enum DatasetUtilities {
     public static let currentWorkingDirectoryURL = URL(
         fileURLWithPath: FileManager.default.currentDirectoryPath)
+    
+    public static let defaultDirectory = FileManager.default.homeDirectoryForCurrentUser
```
BradLarson: Rather than using `homeDirectoryForCurrentUser`, if our goal is to have downloaded datasets persist past a reboot, `applicationSupportDirectory` might be what I'd recommend. If you switch to that, let's use `swift-models` as a subdirectory instead of `.s4tf`.

SaiHemanthBR: I have incorporated the suggested change.

For docker containers, add `-v /my/host/swift-models:/root/.local/share/swift-models/` for persistent.

dabrahams:
@compnerd to suggest standard cache directory.

Datasets/DatasetUtilities.swift
compnerd: Sorry, I think think that `FileManager.url(for:in:appropriateFor:create:)`  https://developer.apple.com/documentation/foundation/filemanager/1407693-url with `.cachesDirectory` might be appropriate for this.  Something like:

```swift
    public static let defaultDirectory =
        FileManager.default.url(for: .cachesDirectory, in: .userDomainMask, appropriateFor: nil, create: true)
            .appendingPathComponent("swift-models")
```

brettkoonce: it was in nscaches, moved to /tmp, now it's moving to a hidden .dir in home...

SaiHemanthBR:
It seems that cachesDirectory is appropriate so, I have changed defaultDirectory to use .cachesDirectory as suggested.

For docker containers, add -v /my/host/swift-models:/root/.cache/swift-models/ for persistence.

SaiHemanthBR:
I think using the Cache directory is the best option. Happy to comply with great coding practices

PR #418, opened 2020-03-24
Remove all Dataset to use Batcher instead

sgugger:
This PR removes all dependency on `Dataset` from swift-models, using `Batcher` instead. More specifically, the old protocol
```swift
public protocol ImageClassificationDataset {
    init()
    var trainingDataset: Dataset<LabeledExample> { get }
    var testDataset: Dataset<LabeledExample> { get }
    var trainingExampleCount: Int { get }
    var testExampleCount: Int { get }
}
```
becomes
```swift
public protocol ImageClassificationDataset {
    associatedtype SourceDataSet: Collection 
    where SourceDataSet.Element == TensorPair<Float, Int32>, SourceDataSet.Index == Int
    init(batchSize: Int)
    var training: Batcher<SourceDataSet> { get }
    var test: Batcher<SourceDataSet> { get }
}
```
All changes are just the natural result of this, with the exception of one addition in MNIST: current implementation only normalizes when `flattened=true`, this PR makes the behavior consistent. I can remove it if it's a problem.


Datasets/ImageClassificationDataset.swift
```diff
@@ -13,11 +13,12 @@
...
 public protocol ImageClassificationDataset {
-    init()
-    var trainingDataset: Dataset<LabeledExample> { get }
-    var testDataset: Dataset<LabeledExample> { get }
-    var trainingExampleCount: Int { get }
-    var testExampleCount: Int { get }
+    associatedtype SourceDataSet: Collection 
+    where SourceDataSet.Element == TensorPair<Float, Int32>, SourceDataSet.Index == Int
+    init(batchSize: Int)
+    var trainingBatcher: Batcher<SourceDataSet> { get }
```
BradLarson: I had used `trainingDataset` originally not to indicate type (we usually don't use types in Swift property names), but to say that it was the training or test dataset as a concept. Would names of `training` and `test` be descriptive enough? For example, in usage: `for batch in dataset.training.sequenced()`.

sgugger: I think `training` and `test` are descriptive enough, yes.

Benchmarks/Models/ImageClassificationTraining.swift
```diff
@@ -40,11 +38,9 @@ where
...
 
         Context.local.learningPhase = .training
-        for epoch in 1...epochs {
-            let trainingShuffled = dataset.trainingDataset.shuffled(
-                sampleCount: dataset.trainingExampleCount, randomSeed: Int64(epoch))
-            for batch in trainingShuffled.batched(batchSize) {
-                let (labels, images) = (batch.label, batch.data)
+        for _ in 1...epochs {
+            for batch in dataset.training.sequenced() {
+                let (images, labels) = (batch.first, batch.second)
```
BradLarson: To reduce noise in the benchmarks, we had used a constant random seed for the shuffling. Is there a way to specify the equivalent in Batcher?

sgugger:
For the shuffle seeding, this is not yet in `Batcher`. Is it important enough to block the merge? There will be a follow-up because the final design of `Batcher` will change. I would rather add the seeding in the final design only, but if we need it earlier, I can add a quick thing for now.

BradLarson:
I created issue #425 as a low-priority reminder for us (or someone else) to add the seeded shuffling functionality at some point in the future.

sgugger:
Good thinking, thanks!

PR #420, opened 2020-03-25
Replace drive letter with %SystemDrive% in win commands

texasmichelle:
Update GPT2-Inference README for consistency with root README.

Follow-up from #412


dabrahams:
@texasmichelle to update with `%SYSTEMDRIVE%`

compnerd:
Please use `%SystemDrive%` rather than `C:` as the system drive may be different.

PR #424, opened 2020-03-27
Fix output shape of Image.resized

kongzii:
Hello,

before this update, calling save on resized image would result to

```
Fatal error: image must be 3-dimensional[1,224,224,3]: file /Users/swiftninjas/s4tf/tensorflow-swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 301
```

because `resized` actually returns a tensor of shape [1, size.0, size.1, channels] and this was saved as a tensor of image.

In loadImagenetteDirectory and loadImagewoofDirectory this was not an issue, because only scalars from the resized image were used. And in ImagenetteBatchers init, it was actually evaded by getting the first dimension from the tensor.

### DIFF
--- a/Datasets/Imagenette/Imagenette.swift
+++ b/Datasets/Imagenette/Imagenette.swift
@@ -120,7 +120,7 @@ func loadImagenetteDirectory(
     let unwrappedLabelDict = labelDict ?? createLabelDict(urls: urls)
     return urls.lazy.map { (url: URL) -> TensorPair<Float, Int32> in
         TensorPair<Float, Int32>(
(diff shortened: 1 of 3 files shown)


sgugger:
Thanks for the PR! It looks good to me. Just making sure with one last round of tests that solving the conflicts with another PR did not break anything and we can merge.

sgugger:
I think you may need to add the same thing to Imagewoof now, since it uses the same thing as Imagenette.

kongzii:
Thanks, updated.

PR #429, opened 2020-03-30
Replace TensorFlow image loading with stb_image

BradLarson:
As part of our efforts to reduce our dependencies on TensorFlow libraries, this replaces the image saving and loading routines with ones provided by the [stb_image](https://github.com/nothings/stb) C single headers. This is inspired by  t-ae's cross-platform image loading library [Swim](https://github.com/t-ae/swim). We're evaluating various image processing options, so this isn't guaranteed to be a permanent implementation, but it works well for our use cases right now.

The interface to the Image type has intentionally been left the same as it was before, in order to restrict all code changes to just Image.swift. I'll follow on with some modifications to the API in a later PR.

This is the last change needed to close out issue #225.


brettkoonce:
i think this is a good pragmatic way to get away from raw ops

Support/CMakeLists.txt
```diff
@@ -1,3 +1,11 @@
+add_library(StbImage
```
compnerd, line 1: Should this be `StbImage` or `stb_image`?

Support/CMakeLists.txt
```diff
@@ -1,3 +1,11 @@
...
+  StbImage/stb_image_write.c
+  StbImage/stb_image.c)
+target_include_directories(StbImage PUBLIC
+  StbImage/include)
+set_target_properties(StbImage PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
```
compnerd, line 7: Is there a strong reason to not just do a `add_subdirectory` and create a separate `CMakeLists.txt` in the subdirectory as is the recommended usage?

Support/CMakeLists.txt
```diff
@@ -1,3 +1,11 @@
+add_library(StbImage
+  StbImage/stb_image_write.c
+  StbImage/stb_image.c)
+target_include_directories(StbImage PUBLIC
+  StbImage/include)
+set_target_properties(StbImage PROPERTIES
```
compnerd, line 6: This is extraneous, it is needed for CMake <3.18 Swift targets only.

Support/CMakeLists.txt
```diff
@@ -22,9 +30,20 @@ set_target_properties(ModelSupport PROPERTIES
...
 
 
+install(TARGETS StbImage
+  ARCHIVE DESTINATION lib/swift/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>
+  LIBRARY DESTINATION lib/swift/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>
+  RUNTIME DESTINATION bin)
+get_swift_host_arch(swift_arch)
+install(FILES
+  $<TARGET_PROPERTY:StbImage,Swift_MODULE_DIRECTORY>/StbImage.swiftdoc
+  $<TARGET_PROPERTY:StbImage,Swift_MODULE_DIRECTORY>/StbImage.swiftmodule
```
compnerd, line 44: These two should be removed - this is not a Swift target.

Support/CMakeLists.txt
compnerd, line 1: Is this library supposed to be dynamically linked or would it be better to statically link it?  If it should always be static, please explicitly mark this as `STATIC`:

```cmake
add_library(stb_image STATIC
```

Support/CMakeLists.txt
```diff
@@ -22,9 +30,20 @@ set_target_properties(ModelSupport PROPERTIES
...
+  DESTINATION lib/swift$<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:_static>/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>/${swift_arch})
```
compnerd, line 45: If this is a static library, then the entirety of the install should be something along the lines of:

```cmake
install(FILES StbImage/module.modulemap
  DESTINATTION lib/swift/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>/StbImage)
```

Support/Image.swift
```diff
@@ -13,14 +13,23 @@
 // limitations under the License.
 
 import Foundation
+import StbImage
```
compnerd, line 16: Can this be an `@_implementationOnly` import?

BradLarson: That makes sense, this is only needed locally and isn't exposed outside here. Updated with this.

Support/CMakeLists.txt
BradLarson: Replaced with your suggestion.

Support/CMakeLists.txt
BradLarson: I've made this static, thanks.

Support/CMakeLists.txt
BradLarson: Removed and replaced with the suggestion below.

Support/CMakeLists.txt
BradLarson: Removed, thanks.

Support/CMakeLists.txt
BradLarson: I've moved this all into a subdirectory and am pulling it in via `add_subdirectory`

Support/CMakeLists.txt
BradLarson: I've gone with the Swift capitalization pattern, even though this is a C library, because that's been the convention for other C libraries that I've seen imported in various projects.

compnerd: In that case, Id really prefer `STBImage` since I believe that `STB` is supposed to be an initialism?

Support/StbImage/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+  include)
+
+
+install(TARGETS StbImage
+  ARCHIVE DESTINATION lib/swift/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>
+  LIBRARY DESTINATION lib/swift/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>
+  RUNTIME DESTINATION bin)
+get_swift_host_arch(swift_arch)
+install(FILES StbImage/module.modulemap
+  DESTINATION lib/swift/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>/StbImage)
```
compnerd, line 14:
```suggestion
```

The library is statically linked and `@_implementationOnly` `import`ed, so there is no client visible aspect to it - it is entirely an implementation detail.  There is nothing to distribute :)

Support/StbImage/module.modulemap
```diff
@@ -0,0 +1,5 @@
+module StbImage {
```
compnerd, line 1: This needs to be moved into `include`, as that is the path that will be searched, not the level above.

Support/StbImage/module.modulemap
```diff
@@ -0,0 +1,5 @@
...
+    header "include/stb_image.h"
+    header "include/stb_image_write.h"
```
compnerd, line 3:
```suggestion
    header "stb_image.h"
    header "stb_image_write.h"
```

This file should be a peer of the headers.

Support/CMakeLists.txt
BradLarson: Good point, renamed to STBImage.

Support/StbImage/CMakeLists.txt
BradLarson: Done.

Support/StbImage/module.modulemap
BradLarson: Ah, that's why. Moved and edited.

Support/StbImage/module.modulemap
BradLarson: Edited.

ematejska:
Looks good.

Tests/SupportTests/ImageTests.swift
```diff
@@ -0,0 +1,97 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+import ModelSupport
+import TensorFlow
+
+final class ImageTests: XCTestCase {
+    let resourceBaseLocation = URL(fileURLWithPath: #file).deletingLastPathComponent()
```
saeta, line 20: will `#file` break in the forthcoming changes to not be full paths?

BradLarson: That's a good point, it probably will. It looks like we have access to `#filePath` in our toolchain (I wasn't sure if we did), so I'll switch this and our other uses of `#file` to `#filePath` in anticipation.

BradLarson: PR #449 is my quick fix to address this.

PR #431, opened 2020-04-01
Adding a ResNet56-CIFAR10 benchmark

BradLarson:
In order to provide a head-to-head comparison with [the tf.keras ResNet56 on CIFAR-10 benchmark model](https://github.com/tensorflow/models/blob/master/official/benchmark/models/resnet_cifar_model.py), this adds a ResNet56 model depth per [the original paper](https://arxiv.org/pdf/1512.03385.pdf) and configures a benchmark of training it on the CIFAR-10 dataset. It also tweaks the ResNet-CIFAR10 example to use the ResNet56 architecture on CIFAR-10, a more common choice for that dataset.

### DIFF
--- a/Benchmarks/Models/ResNetCIFAR10.swift
+++ b/Benchmarks/Models/ResNetCIFAR10.swift
@@ -1,0 +1,49 @@
+// Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Datasets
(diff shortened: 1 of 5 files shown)


Benchmarks/Models/ResNetCIFAR10.swift
```diff
@@ -0,0 +1,49 @@
...
+    func makeTrainingBenchmark(settings: BenchmarkSettings) -> Benchmark {
+        return ImageClassificationTraining<ResNet56, CIFAR10>(settings: settings)
+    }
+}
+
+struct ResNet56: Layer {
+    var model: ResNet
+    
+    init() {
+        model = ResNet(classCount: 10, depth: .resNet56, downsamplingInFirstStage: false, inputFilters: 16)
```
sgugger, line 42: I wonder if `inputFilters` should have a default that comes from the depth (or be set by the depth entirely?). People might forget that it's 16 and not 64 for this version of the ResNet.

BradLarson: I ended up deciding to remove this parameter altogether and just use the two different filter counts for CIFAR-10 and ImageNet-sized images, keyed on the `downsamplingInFirstStage` setting. Looking at other ResNet implementations, they didn't expose this configuration parameter, so I figure we can simplify things by removing that and having it be set internally to match the two image size classes.

sgugger: SGTM

PR #441, opened 2020-04-08
Cyclegan tweaks

brettkoonce:


saeta:
Thank you @brettkoonce !

CycleGAN/main.swift
```diff
@@ -15,14 +15,45 @@
...
+	        remoteRoot: location.deletingLastPathComponent(), localStorageDirectory: directory)
+	}
+
+	datasetFolder = DatasetUtilities.defaultDirectory.appendingPathComponent("CycleGAN", isDirectory: true)
+	downloadZebraDataSetIfNotPresent(to: datasetFolder)
+	trainFolderA = datasetFolder.appendingPathComponent("horse2zebra/trainA")
+	trainFolderB = datasetFolder.appendingPathComponent("horse2zebra/trainB")
+	testFolderA = datasetFolder.appendingPathComponent("horse2zebra/testA")
+	testFolderB = datasetFolder.appendingPathComponent("horse2zebra/testB")
+}
```
BradLarson, line 56: Minor formatting issue: it looks like the indentation here is using tabs, and not four spaces, so the code block is getting indented in a way that is different from the remainder of the code.

CycleGAN/main.swift
```diff
@@ -15,14 +15,45 @@
...
-let datasetFolder = URL(fileURLWithPath: options.datasetPath, isDirectory: true)
-let trainFolderA = datasetFolder.appendingPathComponent("trainA")
-let trainFolderB = datasetFolder.appendingPathComponent("trainB")
-let testFolderA = datasetFolder.appendingPathComponent("testA")
-let testFolderB = datasetFolder.appendingPathComponent("testB")
+var datasetFolder: URL
+var trainFolderA: URL
+var trainFolderB: URL
+var testFolderA: URL
+var testFolderB: URL
```
BradLarson, line 26: Can these still be `let`s, because they are set in both paths of the conditional statement and not reassigned?

brettkoonce: updated!

CycleGAN/main.swift
brettkoonce: tweaked!

PR #444, opened 2020-04-10
Restructuring benchmarks to be batch-based in training, use warmup

BradLarson:
In order to better align the training and inference throughput benchmarks with the ones used for tf.keras, I've converted the training benchmarks to use batches as the fundamental unit of measure, not epochs. I've also introduced a specified warmup period (defaulting to one batch) before batches are counted for throughput.

Logging has been improved, so that you get statistics on examples / second, warmup time, and step time right at the command line. This includes minimum, maximum, and standard deviation so that anomalous outliers can be detected.

Internal setting values have been switched from using magic constants to relying on optionals early and then materializing these into concrete values closer to where options are specified. This should hopefully catch problematic cases earlier in the process.

Finally, the inference benchmarks have been adjusted to use the actual test part of the dataset by default, fixing a crash in the ResNet56 inference benchmark and providing a more realistic measure of inference throughput.

### DIFF
--- a/Benchmarks/Benchmark.swift
+++ b/Benchmarks/Benchmark.swift
@@ -16,7 +16,7 @@ import Foundation
 
 protocol Benchmark {
     var exampleCount: Int { get }
-    func run()
+    func run() -> [Double]
 }
 
 /// Performs the specified benchmark over a certain number of iterations and provides the result to a callback function.
@@ -25,21 +25,32 @@ func measure(
     benchmark: Benchmark
 ) -> BenchmarkResults {
     var timings: [Double] = []
-    for _ in 0..<configuration.settings.iterations {
-        let timing = time(benchmark.run)
-        timings.append(timing)
+    let iterations = configuration.settings.iterations
+    for _ in 0..<iterations {
+        var timedSteps = benchmark.run()
+        timedSteps.removeFirst(configuration.settings.warmupBatches)
+        timings.append(contentsOf: timedSteps)
     }
 
     return BenchmarkResults(
         configuration: configuration, timings: timings, exampleCount: benchmark.exampleCount)
 }
 
+// Returns an uptime-based timestamp in milliseconds.
+func timestampInMilliseconds() -> Double {
+    let divisor: Double = 1_000_000
+    return Double(DispatchTime.now().uptimeNanoseconds) / divisor
+}
+
+/// Calculates and returns the time in milliseconds since the given timestamp.
+func durationInMilliseconds(since timestamp: Double) -> Double {
+    let now = timestampInMilliseconds()
+    return now - timestamp
+}
+
 /// Returns the time elapsed while running `body` in milliseconds.
 func time(_ body: () -> Void) -> Double {
-    let divisor: Double = 1_000_000
-    let start = Double(DispatchTime.now().uptimeNanoseconds) / divisor
+    let start = timestampInMilliseconds()
     body()
-    let end = Double(DispatchTime.now().uptimeNanoseconds) / divisor
-    let elapsed = end - start
-    return elapsed
+    return durationInMilliseconds(since: start)
 }
--- a/Benchmarks/Benchmark.swift
+++ b/Benchmarks/Benchmark.swift
@@ -27,16 +27,19 @@ func measure(
     var timings: [Double] = []
     var warmup: [Double] = []
     let iterations = configuration.settings.iterations
+    let start = timestampInMilliseconds()
     for _ in 0..<iterations {
         var timedSteps = benchmark.run()
         warmup.append(contentsOf: timedSteps.prefix(configuration.settings.warmupBatches))
         timedSteps.removeFirst(configuration.settings.warmupBatches)
         timings.append(contentsOf: timedSteps)
     }
+    let warmupTime = warmup.reduce(0.0, +)
+    let totalTime = durationInMilliseconds(since: start)
 
     return BenchmarkResults(
-        configuration: configuration, warmup: warmup, timings: timings,
-        batchSize: benchmark.batchSize)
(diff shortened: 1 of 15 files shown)


Benchmarks/Models.swift
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-let benchmarkModels: [String: BenchmarkModel] = [
-    "LeNetMNIST": LeNetMNIST(),
-    "ResNetCIFAR10": ResNetCIFAR10(),
+let benchmarkModels: [String: BenchmarkModel.Type] = [
+    LeNetMNIST.name: LeNetMNIST.self,
+    ResNetCIFAR10.name: ResNetCIFAR10.self,
```
saeta, line 17: If there's no need for flexibly mapping from name to `BenchmarkModel.Type`, what do you think of something like the following:

```suggestion
let benchmarkModels = Dictionary([LeNetMNIST.self, ResNetCIFAR10.self].map { ($0.name, $0) })
```

Note: I assume instantiating these `BenchmarkModel` types is expensive, and so that's why we're using their metatypes? Or is there some other reason?

Benchmarks/PrintingStyle.swift
```diff
@@ -128,4 +114,12 @@ extension Array where Element == Double {
     var standardDeviation: Element {
         return Double.sqrt(variance)
     }
+    
+    func printStatistics(indentation: Int) {
```
saeta, line 118: CC @dabrahams for some pretty-print use cases.

Benchmarks/Benchmark.swift
```diff
@@ -25,21 +25,38 @@ func measure(
...
+        warmup.append(contentsOf: timedSteps.prefix(configuration.settings.warmupBatches))
+        timedSteps.removeFirst(configuration.settings.warmupBatches)
+        timings.append(contentsOf: timedSteps)
     }
+    let warmupTime = warmup.reduce(0.0, +)
+    let totalTime = durationInMilliseconds(since: start)
 
     return BenchmarkResults(
-        configuration: configuration, timings: timings, exampleCount: benchmark.exampleCount)
+        configuration: configuration, timings: timings, warmupTime: warmupTime,
```
shabalind, line 41: It would be nice to include `warmupTimings` array as well, similar to `timings`. Once we aggregate to total warmupTime, the information about individual warm-up batches is gone irreversibly.

Benchmarks/Models.swift
```diff
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-let benchmarkModels: [String: BenchmarkModel] = [
-    "LeNetMNIST": LeNetMNIST(),
-    "ResNetCIFAR10": ResNetCIFAR10(),
+let benchmarkModelList: [BenchmarkModel.Type] = [
```
shabalind, line 15: Maybe `benchmarkModelTypes` ?

shabalind: Otherwise distinction between `benchmarkModels` and `benchmarkModelList` is not very clear, solely based on the names.

Benchmarks/Models/ImageClassificationInference.swift
```diff
@@ -40,18 +40,30 @@ where Model: ImageClassificationModel, ClassificationDataset: ImageClassificatio
...
-        }
+        self.images = images
     }
 
-    func run() {
-        for _ in 0..<batches {
-            let _ = model(images)
+    func run() -> [Double] {
+        var batchTimings: [Double] = []
+        if let images = images {
```
shabalind, line 48: It looks like `images` now implicitly encodes `training` vs `inference` modes. Can we make it more explicit?

Benchmarks/Benchmark.swift
BradLarson: That was my original design, but I found the reporting on it to be less than helpful. You really care about the aggregate warmup time, which in most cases seems to be up to the end of the first batch. The fact that we see warmup effects across the first two batches in x10 still strikes me as odd, but I wasn't able to extract any useful information from having statistics on the multiple startup batches I was reporting before. By combining these into a single value, it makes overall timing easier to interpret, and I believe this is consistent with how other benchmarks report this.

Benchmarks/Models.swift
BradLarson: I struggled with the naming here. `benchmarkModelTypes` seems like a better choice, so I've changed to that.

Benchmarks/Models/ImageClassificationInference.swift
BradLarson: The inference case was entirely synthetic before, with a hardcoded image size. That size didn't work for both CIFAR-10 and MNIST-sized networks, so I switched to using real data by default, with the option of providing synthetic data. However, this isn't really exposed to the user performing benchmarks, so I'm working on making a more robust real vs. synthetic option that applies to both inference and training and will add that to this.

Benchmarks/Benchmark.swift
shabalind: Sounds good to me. We can always change this later on in case we need more data on warmup.

Benchmarks/Models/ImageClassificationInference.swift
BradLarson: OK, I've generalized the synthetic / real data selection into `--real` and `--synthetic` options that apply to both the training or inference benchmarks. A new SyntheticImageDataset has been created to represent random image and label data, and preferred input sizes and output label counts are now attached to benchmark model types. Type restrictions are a little tight for datasets in the inference and training loops, but the new Batcher API will allow for easing that (and for generating random batches on the fly). This seems like it should work for now.

PR #445, opened 2020-04-13
Fixed collectURLs extensions filtering

Andr0id100:
A minor bug rendered the extension filtering ineffective as the function repeatedly checks only the first file of the directory.


sgugger:
Thanks for fixing!

PR #446, opened 2020-04-14
Add support for COCO dataset

shabalind:
This PR adds initial scaffolding for COCO dataset family, including:

* `COCO` dataset type, that parses and creates JSON annotation file indexes
* Initial support for masks based on COCO's RLE encoding
* Automatic download of the annotation files and images from the coco server

Some pieces are still missing and can be added later:

* API for model evaluation (i.e., `COCOEval` in pycocotools)


Datasets/COCO/COCO.swift
```diff
@@ -0,0 +1,586 @@
...
+        self.categoryToImages = [:]
+        self.createIndex()
+    }
+
+    mutating func createIndex() {
+        if let info = dataset["info"] {
+            self.info = info as! Info
+        }
+        if let annotations = dataset["annotations"] {
+            let anns = annotations as! [Annotation]
```
BradLarson, line 44: For here and the following cases, you might be able to simplify this (and make it a little safer) by combining the two lines into

```swift
if let annotations = dataset["annotations"] as? [Annotation] {
```

Datasets/COCO/COCO.swift
```diff
@@ -0,0 +1,586 @@
...
+        let contents = try String(contentsOfFile: fileURL.path)
+        let data = contents.data(using: .utf8)!
+        let parsed = try JSONSerialization.jsonObject(with: data)
+        self.dataset = parsed as! Dataset
+        self.info = [:]
+        self.annotations = [:]
+        self.categories = [:]
+        self.images = [:]
+        self.imageToAnnotations = [:]
+        self.categoryToImages = [:]
```
BradLarson, line 35: If these are always initialized to being empty, they could probably be moved up to the property declaration.

Datasets/COCO/COCO.swift
```diff
@@ -0,0 +1,586 @@
...
+            let id = ann["id"] as! AnnotationId
+            annIds.append(id)
+        }
+        return annIds
+    }
+
+    /// Get category ids that satisfy given filter conditions.
+    public func getCategoryIds(
+        categoryNames: [String] = [],
+        supercategoryNames: [String] = [],
```
BradLarson, line 130: It might make sense to have these represented as Sets, not Arrays, because `contains()` on a Set is O(1) vs. O(n) and `contains()` is called fairly frequently in the below. Possibly the same for categoryIds.

Examples/COCO/main.swift
```diff
@@ -0,0 +1,19 @@
+import Datasets
+import Foundation
+import SwiftProtobuf
+
+let coco = COCOVariant.loadInstancesVal2017()
```
BradLarson, line 5: Rather than having a standalone COCO example that only loads the dataset and logs out values from it, maybe we could just rely on the unit tests for now and put in place an example once we have a matching model. We'd probably end up removing or renaming this at that point, anyway, and the tests should probably provide a sufficient means of exercising the dataset.

shabalind:
@BradLarson I applied your suggestions, and also added:
* `ObjectDetectionDataset` protocol  with `CocoDataset` as a single implementation
* Automatic image downloading for the training and evaluation variants of the dataset

shabalind:
@BradLarson Kokoro is passing now. The PR is ready for another round of review.

PR #448, opened 2020-04-16
build: co-locate artifacts

compnerd:
This co-locates the runtime and library artifacts from the sub-builds.
This makes it easier to find the artifacts, reduces the overall work
that needs to be done, and makes it easier to consume the project.


compnerd:
The files are copied, not moved, so there shouldn't be any empty directories.  I think that incrementals can be confusing as the old content will not be removed.  However, future clean builds should avoid those problems.

PR #451, opened 2020-04-20
explicitly set image params for inference test

brettkoonce:


BradLarson:
I'm completely rearchitecting the inference tests in PR #444, which then will allow for specifying size of synthetic data used for training and inference, in part to solve this problem. I've just been holding off merging that until we fix some tests internally around the repository.

Thanks for working on the problem, sorry I didn't make it clearer that this was being addressed there.

brettkoonce:
ahh, didn't see that.  carry on, no worries!

PR #453, opened 2020-04-20
Fixes to make GPT2-WikiText2 trained on TPU with x10

xihui-wu:
1. TextUnsupervised.swift: make number of documents to process configurable
2. TransformerLM.swift 
     - make device consistent while calling bandPart
     - make layers to be 'var' to be copiable.
3. BytePairEncoder.swift: set 'limitedBy' to avoid String index is out of bounds FatalError

### DIFF
--- a/Datasets/TextUnsupervised/TextUnsupervised.swift
+++ b/Datasets/TextUnsupervised/TextUnsupervised.swift
@@ -78,7 +78,7 @@ public struct TextUnsupervised {
         bpe: BytePairEncoder,
         variant: TextUnsupervisedVariant = TextUnsupervisedVariant.wikiText2,
         trainingBatchSize: Int = 8, validationBatchSize: Int = 4, sequenceLength: Int = 1024,
-        trainingDocumentSize: Int = 4, validationDocumentSize: Int = 4
+        trainingDocumentCount: Int = 4, validationDocumentCount: Int = 4
     ) {
         do {
             self.bpe = bpe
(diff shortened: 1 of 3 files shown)


Support/Text/BytePairEncoder.swift
```diff
@@ -207,9 +207,11 @@ extension BytePairEncoder {
...
         case .gpt2:
             for match in matches {
-                let start = token.index(token.startIndex, offsetBy: match.range.lowerBound)
-                let end = token.index(token.startIndex, offsetBy: match.range.upperBound)
-                parts.append(String(token[start..<end]))
+                let start = token.index(token.startIndex, offsetBy: match.range.lowerBound, limitedBy: token.endIndex)
+                let end = token.index(token.startIndex, offsetBy: match.range.upperBound, limitedBy: token.endIndex)
+                if start != nil && end != nil {
+                  parts.append(String(token[start!..<end!]))
+                }
```
BradLarson, line 214: You might be able to simplify this by using something like
```swift
if let start = ..., let end = ... {
    parts.append(String(token[start..<end]))
}
```

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -144,14 +145,14 @@ public struct TextUnsupervised {
     // Only supports CSV files.
     private static func loadDirectory(
         named name: String, in directory: URL, bpe: BytePairEncoder,
-        variantDetails: TextUnsupervisedVariantDetails, batchSize: Int, sequenceLength: Int
+        variantDetails: TextUnsupervisedVariantDetails, batchSize: Int, sequenceLength: Int,
+        documentSize: Int = 4
```
texasmichelle, line 149: How about something like `numberOfDocuments`?

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -144,14 +145,14 @@ public struct TextUnsupervised {
...
     ) throws -> LanguageModelDataset<[[Int]]> {
         downloadIfNotPresent(to: directory, variantDetails: variantDetails)
         let path = directory.appendingPathComponent("\(variantDetails.filename)/\(name).csv")
 
         let documentsFull = try readCSV(in: path)
-        // TODO(michellecasbon): Process a larger number of documents.
-        let documents = Array(documentsFull[0..<4])
+        let documents = Array(documentsFull[0..<documentSize])
```
texasmichelle, line 155: It would be good to use the max of `documentSize` and `documentsFull.count`

Datasets/TextUnsupervised/TextUnsupervised.swift
xihui-wu: Good catch! How about 'documentCount' ? It should be better since the standard in swift always use "count" instead of "num" as I know.

Datasets/TextUnsupervised/TextUnsupervised.swift
xihui-wu: You mean min of ?

texasmichelle: 😆 yes!

PR #460, opened 2020-04-24
Fix CycleGAN build error.

dan-zheng:
`String` has no member called `length`.
Use `String.isEmpty` instead.

---

Fixes `swift build` (on macOS):
```console
$ swift build
swift-models/CycleGAN/main.swift:28:24: error: value of type 'String' has no member 'length'
if options.datasetPath.length != 0 {
   ~~~~~~~~~~~~~~~~~~~ ^~~~~~
```

Confirmed that `swift run CycleGAN` works.

### DIFF
--- a/CycleGAN/main.swift
+++ b/CycleGAN/main.swift
@@ -25,7 +25,7 @@ let trainFolderB: URL
 let testFolderA: URL
 let testFolderB: URL
 
-if options.datasetPath.length != 0 {
+if !options.datasetPath.isEmpty {
     datasetFolder = URL(fileURLWithPath: options.datasetPath, isDirectory: true)
     trainFolderA = datasetFolder.appendingPathComponent("trainA")
     trainFolderB = datasetFolder.appendingPathComponent("trainB")


brettkoonce:
Above is what I tried initially, but on Linux it fails with:

```
Fatal error: Error raised at top level: Error Domain=NSCocoaErrorDomain Code=260
 "The file doesn’t exist.": file /home/skoonce/swift/swift-source/swift/stdlib/$
ublic/core/ErrorType.swift, line 200
```

Because the "" string being passed as the dataset directory, which i couldn't set to be a nil because of parameter requirements in cli defaults thing (is there a workaround)?  This code has been copied to pix2pix as well, fwiw.

dan-zheng:
Using `Optional` sounds better to me! I can quickly look into that, after this patch fixes the build.

@brettkoonce: `swift run CycleGAN` works for me on Linux with this patch. Could you please share more info about your error?

brettkoonce:
not sure, now it's working w/ latest swift/swift-models master here.  #461 is simpler logic, works for me!

PR #464, opened 2020-04-25
[BERT] Specify devices explicitly

texasmichelle:
This resolves device mismatch runtime failures when running on XLA.


texasmichelle:
Requires newly merged changes in swift-apis, which should resolve presubmit failures after tonight's nightly builds.

PR #468, opened 2020-04-27
Add eager / X10 backend selection to the benchmarks

BradLarson:
In order to provide benchmarks for the new x10 backend, in addition to the existing eager backend, this adds the `--x10` and `--eager` options to the Benchmarks target. These options switch between using the X10 and eager backends, respectively, and support has been added within the image classification inference and training benchmarks to support X10 in parallel with the default eager execution.

For now, we're using the first two batches as a warmup period, instead of just the first batch (as is done for the other PerfZero benchmarks), due to an extra recompilation we're experiencing in the second batch. Once that extra recompilation is removed, we'll switch back to a single batch for warmup.

Additionally, to work around some caching artifacts we're seeing with batched data, I'm doing an initial empty pass through the entire training dataset before running the throughput benchmarks in order to warm up any caches involved. This provides more consistent results.

### DIFF
--- a/Benchmarks/Models/ImageClassificationInference.swift
+++ b/Benchmarks/Models/ImageClassificationInference.swift
@@ -52,13 +52,30 @@ where Model: ImageClassificationModel, ClassificationDataset: ImageClassificatio
     }
 
     func run(backend: Backend) -> [Double] {
+        let device: Device
+        switch backend {
+        case .eager: device = Device.defaultTFEager
+        case .x10: device = Device.defaultXLA
+        }
+        model.move(to: device)
+        LazyTensorBarrier()
+
         var batchTimings: [Double] = []
         var currentBatch = 0
         for batch in testDataset.sequenced() {
             if (currentBatch >= self.batches) { break }
             let images = batch.first
+            let deviceImages: Tensor<Float>
+            switch backend {
+            case .eager:
+                deviceImages = images
+            case .x10:
+                deviceImages = Tensor(copying: images, to: device)
+            }
+
             let batchTime = time{
-                let _ = model(images)
+                let _ = model(deviceImages)
+                LazyTensorBarrier()
             }
             batchTimings.append(batchTime)
             currentBatch += 1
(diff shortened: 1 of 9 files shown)


Benchmarks/Models/ImageClassificationInference.swift
```diff
@@ -51,14 +51,31 @@ where Model: ImageClassificationModel, ClassificationDataset: ImageClassificatio
...
 
-    func run() -> [Double] {
+    func run(backend: Backend) -> [Double] {
+        let device: Device
+        switch backend {
+        case .eager: device = Device.defaultTFEager
+        case .x10: device = Device.defaultXLA
+        }
+        model.move(to: device)
+        LazyTensorBarrier()
```
asuhan, line 61: What's the reason for this `LazyTensorBarrier` call? I know we did this before, but it was a quick & dirty way to avoid the random initialization to bundle together with the actual model and delay the steady state by one step. If you initialize the model with the eager backend, then move it to the relevant device(s) for the actual training, this shouldn't be needed.

Benchmarks/Models/ImageClassificationTraining.swift
```diff
@@ -45,29 +45,52 @@ where
...
         var beforeBatch = timestampInMilliseconds()
         var model = Model()
+        model.move(to: device)
         // TODO: Split out the optimizer as a separate specification.
-        let optimizer = SGD(for: model, learningRate: 0.1)
+        var optimizer = SGD(for: model, learningRate: 0.1)
+        optimizer = SGD(copying: optimizer, to: device)
         var batchTimings: [Double] = []
         var currentBatch = 0
+        LazyTensorBarrier()
```
asuhan, line 64: Same comment I made for inference.

Benchmarks/Models/ImageClassificationInference.swift
BradLarson: Good catch. I'd left that in there from when the model was initialized via x10, but looking at the trace I can see this isn't needed now. Removing it and the one below.

Benchmarks/Models/ImageClassificationTraining.swift
BradLarson: Same as above, removing this because it's no longer needed.

brettkoonce:
🥳

PR #470, opened 2020-04-28
Added Oxford IIIT Pets Dataset

Andr0id100:
The current implementation only supports image segmentation but the code can be further adapted for image classification.

I need some help for a few sections.


Datasets/OxfordIIITPets/OxfordIIITPets.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    /// let _ = DatasetUtilities.downloadResource(
+    ///     filename: "images", fileExtension: "tar.gz",
+    ///     remoteRoot: remoteRoot, localStorageDirectory: directory
+    /// )
+
+    /// let _ = DatasetUtilities.downloadResource(
+    ///     filename: "annotations", fileExtension: "tar.gz",
+    ///     remoteRoot: remoteRoot, localStorageDirectory: directory
+    /// )
```
Andr0id100, line 91: The file couldn't be extracted if it was downloaded using the downloadResource function so I had to resort to using Python to download the file.
The issue is better described in this forum post:
https://groups.google.com/a/tensorflow.org/forum/#!topic/swift/xFFuAkHCWO0

Any suggestions to fix this problem will be great.

Datasets/OxfordIIITPets/OxfordIIITPets.swift
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func loadOxfordIIITPets(filename: String, in directory: URL, imageSize: Int) throws -> LazyDataSet {
+    downloadOxfordIIITPetsIfNotPresent(to: directory)
+    let imageURLs = getImageURLs(filename: filename, directory: directory)
+    return imageURLs.lazy.map { (imageURL: URL) -> TensorPair<Float, Int32> in
+        TensorPair<Float, Int32>(
+            first:
+                Image(jpeg: imageURL).resized(to: (imageSize, imageSize)).tensor[0..., 0..., 0..<3]
+                / 255.0,
```
Andr0id100, line 102: Is this the proper way to select channels from an image.
Side note: I had to use 0..<3 because a few images had four channels and I selected only the first 3.

Datasets/OxfordIIITPets/OxfordIIITPets.swift
```diff
@@ -0,0 +1,145 @@
...
+            second: Tensor<Int32>(
+                Image(jpeg: makeAnnotationURL(imageURL: imageURL, directory: directory)).resized(
+                    to: (imageSize, imageSize)
+                ).tensor[0..., 0..., 0...0] - 1
+            )
```
Andr0id100, line 107: There doesn't seem to be any explicit way to read png images and passing the url to the jpeg function seems to work just fine except for reading a single channel image with a label for each pixel as a three channel image with same data in each channel.

Also, can the process of converting Image -> Float Tensor -> Int32 Tensor be improved?

Datasets/OxfordIIITPets/OxfordIIITPets.swift
```diff
@@ -0,0 +1,145 @@
...
+
+func makeAnnotationURL(imageURL: URL, directory: URL) -> URL {
+    let filename = imageURL.deletingPathExtension().lastPathComponent
+    return directory.appendingPathComponent("annotations/trimaps/\(filename).png")
+}
+
+func getImageURLs(filename: String, directory: URL) -> [URL] {
+    let filePath = directory.appendingPathComponent("annotations/\(filename)")
+    let imagesRootDirectory = directory.appendingPathComponent("images", isDirectory: true)
+    let fileContents = try? String(contentsOf: filePath)
```
Andr0id100, line 121: Is this the correct way to read a text file in swift ?

Andr0id100:
Hey @BradLarson just a gentle reminder in case you missed this.

Datasets/OxfordIIITPets/OxfordIIITPets.swift
BradLarson: When I migrated the Image type to use stb_image, I intentionally preserved the same interface, which only supported loading JPEG images before. However, `stbi_load()` used [here](https://github.com/tensorflow/swift-models/blob/master/Support/Image.swift#L68 ) opens a variety of formats, which is how you're able to pull in PNGs. It assumes RGB encoding with no alpha channel, which isn't the case for PNGs.

Ultimately, I want to get away with storing float Tensors within the Image type, and rely on byte arrays that get packaged into appropriate Tensors at the point of use. I believe this will lead to better performance with the X10 backend and improve the conversion process above.

Datasets/OxfordIIITPets/OxfordIIITPets.swift
BradLarson: One possibility is that using Data to download the file remotely is somehow triggering an automatic gunzipping from the server, and what you're getting is in fact the .tar archive instead. I may have seen this before, but haven't had the time to investigate. I can try a quick experiment to see if that's the case and if we can detect that happening before trying to unarchive.

Andr0id100: I ran some experiments and it seems that the issue appears to be resolved in the 0.9 release though there doesn't seem to be anything obvious in the release notes that would fix it.

Andr0id100:
I am working on adding some tests and should be done soon.

BradLarson:
Looks like the OxfordIIITPets.swift file might be missing from the CMakeLists for Datasets, leading to a test failure:

> OxfordIIITPets/OxfordIIITPetsTests.swift /swift-models/Tests/DatasetsTests/XCTestManifests.swift    && :
> /swift-models/Tests/DatasetsTests/OxfordIIITPets/OxfordIIITPetsTests.swift:7:19: error: cannot find 'OxfordIIITPets' in scope
>     let dataset = OxfordIIITPets(batchSize: 64)

Andr0id100:
The conflicts are because of WordSeg's entry in CMakeList. Just adding the file entries for CMake would result in a build failure hence it would be cleaner to just start a new pull request and close this one.
The new PR is at #548

PR #471, opened 2020-04-28
add mnist xla demo

brettkoonce:


BradLarson:
Rather than doing this as a separate example, let's maybe start reworking our existing examples to support X10. We want the examples to demonstrate best practices, so it would make sense to me to have our examples all use X10 where that's the best choice. For the image classification models, that should be relatively straightforward. We're working on improving the models themselves in some of the text and other cases to handle changing shapes.

Also, as of the latest toolchains the environment variables no longer need to be set to have X10 work correctly with GPUs. That's all now autodetected at runtime. Haven't tested the TPU case, but I believe that's also true there. If not, we'll set these for you in TPU Colab instances.

Finally, the LazyTensorBarrier in the inference case should come after the `loss.scalarized()` call, I believe, in order to cut that trace correctly.

brettkoonce:
kk

PR #477, opened 2020-05-01
Create swift.yml

compnerd:
Add an initial GitHub workflow to enable a `swift-format` linting action.


compnerd:
I'll update this once we figure out what to do with the swift-format issues

PR #479, opened 2020-05-01
[ShuffleNetV2] Resolve warning about non-mutating var

texasmichelle:
Change `inputChannels` from variable to constant to resolve compiler warning:

```
Models/ImageClassification/ShuffleNetV2.swift:73:13: warning: variable 'inputChannels' was never mutated; consider changing to 'let' constant
```


texasmichelle:
Would it also be worthwhile to add this to the release branch, to avoid the errors that show up in colab? Or will it get picked up from master?

PR #484, opened 2020-05-04
Update .swift-format

compnerd:
Update the swift-format file to match the configuration from swift-apis


texasmichelle:
What uses this file currently? Is it connected to any presubmit hooks or is it there for IDEs to pick up?

compnerd:
Nothing uses it currently; I dont know of any IDE integration for this.

Issue #480, opened 2020-05-04
[BERT-CoLA] Add epochs


texasmichelle:
CoLA uses a custom data loader, which could be improved by using Epochs instead, now that it is part of the v0.9 release.

This will go a long way to resolving #432.

PR #487, opened 2020-05-05
Use Epochs to load CoLA

sgugger:
This PR changes the way CoLA is loaded to use the Epochs API.
It removes the test set (that was unused) and does the minimal amount of changes to work with the script in BertCoLA.

Two notes where the design changed:
- renamed dev to validation as it's more expressive
- put the closure at the end of the init of the CoLA dataset so that we can use a trailing closure.

Added functionality: we can pass a random generator and thus get reproducible results.


texasmichelle:
The kokoro failure is looking for the functions defined at the bottom of DataUtilities.swift. @BradLarson do we have a common function to use instead?

sgugger:
Yeah I just realized that, I deleted the two files thinking they were useless now but did not try to build after, sorry about that. I can leave DatasetUtilities with them for now, there is a TODO to rely on the swift-models at some point, when it is done, this can be removed.

Examples/BERT-CoLA/main.swift
```diff
@@ -41,21 +41,17 @@ var bertClassifier = BERTClassifier(bert: bert, classCount: 1)
...
-    exampleMap: exampleMapFn,
-    taskDirectoryURL: workspaceURL,
-    maxSequenceLength: maxSequenceLength,
-    batchSize: batchSize,
-    dropRemainder: true)
+  taskDirectoryURL: workspaceURL,
+  maxSequenceLength: maxSequenceLength,
+  batchSize: batchSize,
+  entropy: SystemRandomNumberGenerator()
+) { (example: CoLAExample) -> LabeledTextBatch in
```
texasmichelle, line 49: What are the advantages of moving to a trailing closure?

sgugger: It's more swift-idiomatic from what Dave constantly tells me ;)

The main advantage is that you don't have to define your `exampleMap` function separately and pass it, you can just put in curly brackets at the end (like in the main colab example).

texasmichelle:
This is so helpful!! It cleans up the CoLA implementation considerably. Although there are aspects of `Epochs` that are not entirely clear to me at first glance, it's great that it covers so much of this functionality. Thank you!

sgugger:
Thanks! Hope the design review on Friday will clarify the aspects of Epochs that are unclear.

texasmichelle:
Created issues #488 and #489 for later follow-up.

texasmichelle:
Fixes #480

PR #490, opened 2020-05-06
Load encoded file if BPE is nil in TextUnsupervised

xihui-wu:
Text encoding with Byte Pair Encoder has O(N*N) time complexity https://github.com/tensorflow/swift-models/blob/master/Support/Text/BytePairEncoder.swift#L101. Encoding for full WikiText2 training and testing dataset can take ~100 minutes. 

We wanted to make the cache optional where if the encoded text provided then load them directly.

### DIFF
--- a/Datasets/TextUnsupervised/TextUnsupervised.swift
+++ b/Datasets/TextUnsupervised/TextUnsupervised.swift
@@ -179,7 +179,7 @@ public struct TextUnsupervised {
             let pathPrefix = directory.appendingPathComponent("\(variantDetails.encodedFileName!)/\(name)").path
             for i in 0..<documentCount {
                 encodedDocs += [
-                  NSArray(contentsOf: URL(fileURLWithPath: "\(pathPrefix)/doc_\(i).txt")) as! [Int]
+                  try NSArray(contentsOf: URL(fileURLWithPath: "\(pathPrefix)/doc_\(i).txt"), error: ()) as! [Int]
                 ]
             }
         }
--- a/Datasets/TextUnsupervised/TextUnsupervised.swift
+++ b/Datasets/TextUnsupervised/TextUnsupervised.swift
@@ -57,9 +57,19 @@ public struct TextUnsupervised {
         var fileExtension = "tgz"
     }
 
+    private struct EncodedWikiText2Details: TextUnsupervisedVariantDetails {
+        var variant = TextUnsupervisedVariant.wikiText2
+        var location = URL(string: "https://storage.cloud.google.com/s4tf-hosted-binaries/datasets/WikiText2/")!
+        var trainingDirectoryName = "train"
+        var validationDirectoryName = "test"
+        var filename = "encoded-wikitext-2"
+        var fileExtension = "tgz"
+    }
+
     public let trainingDataset: LanguageModelDataset<[[Int]]>
     public let validationDataset: LanguageModelDataset<[[Int]]>
-    public let bpe: BytePairEncoder
+    // bpe is optional. If it's nil then we skip encoding and load encoded text directly.
+    public let bpe: BytePairEncoder?
     public let variant: TextUnsupervisedVariant
     private let variantDetails: TextUnsupervisedVariantDetails
 
@@ -75,9 +85,8 @@ public struct TextUnsupervised {
     }
 
     public init(
-        bpe: BytePairEncoder,
+        bpe: BytePairEncoder? = nil,
         variant: TextUnsupervisedVariant = TextUnsupervisedVariant.wikiText2,
-        skipEncodingByLoadingFrom: String? = nil,
         trainingBatchSize: Int = 8, validationBatchSize: Int = 4, sequenceLength: Int = 1024,
         trainingDocumentCount: Int = 4, validationDocumentCount: Int = 4
     ) {
@@ -90,20 +99,17 @@ public struct TextUnsupervised {
                 let variantDetails = WikiText103Details()
                 self.variantDetails = variantDetails
             case .wikiText2:
-                let variantDetails = WikiText2Details()
-                self.variantDetails = variantDetails
+                self.variantDetails = (bpe == nil ? EncodedWikiText2Details() : WikiText2Details())
             }
 
             let localStorageDirectory: URL = FileManager.default.temporaryDirectory
                 .appendingPathComponent(
                     variant.rawValue, isDirectory: true)
             self.trainingDataset = try TextUnsupervised.loadTraining(
-                skipEncodingByLoadingFrom: skipEncodingByLoadingFrom,
                 localStorageDirectory: localStorageDirectory, bpe: bpe,
                 variantDetails: variantDetails, batchSize: trainingBatchSize,
                 sequenceLength: sequenceLength, documentCount: trainingDocumentCount)
             self.validationDataset = try TextUnsupervised.loadValidation(
-                skipEncodingByLoadingFrom: skipEncodingByLoadingFrom,
                 localStorageDirectory: localStorageDirectory, bpe: bpe,
                 variantDetails: variantDetails, batchSize: validationBatchSize,
                 sequenceLength: sequenceLength, documentCount: validationDocumentCount)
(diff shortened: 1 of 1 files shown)


Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -77,6 +77,7 @@ public struct TextUnsupervised {
     public init(
         bpe: BytePairEncoder,
         variant: TextUnsupervisedVariant = TextUnsupervisedVariant.wikiText2,
+        skipEncodingByLoadingFrom: String? = nil,
```
dabrahams, line 80: This label is probably too long, but that can be adjusted later.  Regardless, this kind of argument label, which ends with a preposition, needs an explicit parameter name, such as `pathToEncodedText` or something.

xihui-wu: Good go know this thanks. But this parameter is no longer needed in my latest change.

xihui-wu:
> Since the complexity issues are caused by the `BytePairEncoder` implementation, it would be good to find a way to cache at that level, which would impact everything that references it. That would be my preference, but as a shortcut for benchmarks and quick prototyping, I can see how this PR makes sense (even with a bpe cache).
> 
> This approach effectively short-circuits the byte pair encoder entirely, by storing a preprocessed dataset. If you want to go this route for now, it would be worth storing this file in GCS (we might want to consider hosting the data files as well) and adding the path to `TextUnsupervisedVariantDetails`, rather than passing it in. This warrants reconfiguring the `init()` methods. For example, the first one creates an empty bpe, which isn't really useful for anything. Instead, this preprocessed dataset could be loaded, which would make the bpe optional.

Great point:) Updated with EncodedWikiText2Details.

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -57,9 +57,19 @@ public struct TextUnsupervised {
         var fileExtension = "tgz"
     }
 
+    private struct EncodedWikiText2Details: TextUnsupervisedVariantDetails {
```
texasmichelle, line 60: Instead of adding a new variant, how about replacing `filename` with `rawFilename` and adding `var encodedFilename`? That way, the bpe nil check occurs closer to the `embedding()` call, which is a bit clearer IMO.

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -57,9 +57,19 @@ public struct TextUnsupervised {
...
+        var validationDirectoryName = "test"
+        var filename = "encoded-wikitext-2"
+        var fileExtension = "tgz"
+    }
+
     public let trainingDataset: LanguageModelDataset<[[Int]]>
     public let validationDataset: LanguageModelDataset<[[Int]]>
-    public let bpe: BytePairEncoder
+    // bpe is optional. If it's nil then we skip encoding and load encoded text directly.
+    public let bpe: BytePairEncoder?
```
texasmichelle, line 72: This gives us more flexibility in the type of models that can use this dataset 👍

Datasets/TextUnsupervised/TextUnsupervised.swift
dabrahams: Even better!

Datasets/TextUnsupervised/TextUnsupervised.swift
xihui-wu: 👍Added encodedFilename as optional in newer change. Does it LG?

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -144,30 +149,39 @@ public struct TextUnsupervised {
...
+        if let bpe = bpe {
+            let path = directory.appendingPathComponent("\(variantDetails.filename)/\(name).csv")
+            let documentsFull = try readCSV(in: path)
+            let documents = Array(documentsFull[0..<min(documentCount, documentsFull.count)])
+            encodedDocs = documents.map { embedding(for: $0, bpe: bpe) }
+        } else {
+            let pathPrefix = directory.appendingPathComponent("\(variantDetails.encodedFileName!)/\(name)").path
+            for i in 0..<documentCount {
+                encodedDocs += [
+                  NSArray(
```
texasmichelle, line 168: I get a deprecation warning here:
```
/home/michellecasbon/repos/tensorflow/swift-models/Datasets/TextUnsupervised/TextUnsupervised.swift:168:19: warning: 'init(contentsOfFile:)' is deprecated
                  NSArray(
                  ^
```

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -144,30 +149,39 @@ public struct TextUnsupervised {
...
+        downloadIfNotPresent(to: directory, variantDetails: variantDetails, downloadEncodedFile: bpe == nil)
+
+        var encodedDocs: [[Int]] = []
+        if let bpe = bpe {
+            let path = directory.appendingPathComponent("\(variantDetails.filename)/\(name).csv")
+            let documentsFull = try readCSV(in: path)
+            let documents = Array(documentsFull[0..<min(documentCount, documentsFull.count)])
+            encodedDocs = documents.map { embedding(for: $0, bpe: bpe) }
+        } else {
+            let pathPrefix = directory.appendingPathComponent("\(variantDetails.encodedFileName!)/\(name)").path
```
texasmichelle, line 165: If a caller references a dataset without an `encodedFileName` and does not include a bpe, this will crash. Potentially add a precondition on creation.

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -45,21 +46,24 @@ public struct TextUnsupervised {
...
         var validationDirectoryName = "test"
         var filename = "wikitext-2"
+        var encodedFileName: String? = "encoded-wikitext-2"
         var fileExtension = "tgz"
     }
 
     public let trainingDataset: LanguageModelDataset<[[Int]]>
     public let validationDataset: LanguageModelDataset<[[Int]]>
-    public let bpe: BytePairEncoder
+    // bpe is optional. If it's nil then we skip encoding and load encoded text directly.
```
texasmichelle, line 65: I think a description of this behavior would be better placed either in front of `init()` or `loadDirectory()`. I recommend following the [API design doc](https://swift.org/documentation/api-design-guidelines/) as a guide (please do not interpret this as a request to overhaul the whole file - just this one behavior).

Datasets/TextUnsupervised/TextUnsupervised.swift
xihui-wu: I use https://developer.apple.com/documentation/foundation/nsarray/1414742-write and https://developer.apple.com/documentation/foundation/nsarray/1413844-init to write and read arrays between file. They are deprecated. Do you know any better way to do it ?

Datasets/TextUnsupervised/TextUnsupervised.swift
xihui-wu: Good point.

Datasets/TextUnsupervised/TextUnsupervised.swift
texasmichelle: `readCSV()` isn't doing anything fancy. Doubtless there's a better way, but using that approach seems reasonable.

xihui-wu: Find it. I should have switched to https://developer.apple.com/documentation/foundation/nsarray/2879134-init.

Datasets/TextUnsupervised/TextUnsupervised.swift
```diff
@@ -135,39 +139,73 @@ public struct TextUnsupervised {
...
+    /// then download the encoded file instead.
+    ///
+    /// - Parameter name: name of the dataset. Ususally 'train' or 'test'.
+    /// - Parameter directory: directory that files are read from.
+    /// - Parameter bpe: byte pair encoder used for encoding text.
+    /// - Parameter variantDetails: an object containing information of filename, location, etc.
+    /// - Parameter batchSize: number of sequences in a batch.
+    /// - Parameter sequenceLength: number of characters in a sequence.
+    /// - Parameter documentCount: number of documents to proceed. (Refer func readCSV() to see how
+    ///   a text file is chunked into documents.)
```
texasmichelle, line 173: This documentation is a big improvement 👍 Thank you!!

Issue #434, opened 2020-04-02
GPT2 and BERT alignment


texasmichelle:
Refactor the GPT-2 and BERT models to make it clear which concepts are shared and which are distinct. For example, transformers and multi-head attention are shared concepts, but naming collisions have resulted in files such as TransformerBERT.swift and structs such as `MultiHeadAttentionGPT2`. As our collection of transformer-based language models increases, this becomes important for code reuse and maintenance.

Investigate alignment with [HuggingFace APIs](https://github.com/huggingface/transformers), which do a great job of unifying a wide variety of transformer-based models.

PR #494, opened 2020-05-07
[BERT] Rename custom embedding layer

### LINKED ISSUE #493: [BERT] Rename custom Embedding layer
BERT uses a [custom `Embedding` layer](https://github.com/tensorflow/swift-models/blob/master/Models/Text/Embeddings.swift), which causes ambiguity errors when other models attempt to use the [standard version](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Layers/Embedding.swift).

```
error: 'Embedding' is ambiguous for type lookup in this context
  private func setEmbedding(_ embedding: inout Embedding<Float>, to p: EmbeddingParameters) {
                                               ^~~~~~~~~
TensorFlow.Embedding:1:15: note: found this candidate
public struct Embedding<Scalar> : Module where Scalar : TensorFlowFloatingPoint {
              ^
/usr/local/google/home/michellecasbon/repos/texasmichelle/swift-models/Models/Text/Embeddings.swift:21:15: note: found this candidate
public struct Embedding<Scalar: TensorFlowFloatingPoint>: Module, Regularizable {
              ^
```

Identify the differences and rename this version accordingly. If appropriate, request changes in swift-apis.

texasmichelle:
Fixes #493

Change filename and all references.

### DIFF
--- a/Models/Text/BERT.swift
+++ b/Models/Text/BERT.swift
@@ -48,9 +48,9 @@ public struct BERT: Module, Regularizable {
     @noDerivative public let typeVocabularySize: Int
     @noDerivative public let initializerStandardDeviation: Scalar
 
-    public var tokenEmbedding: Embedding<Scalar>
-    public var tokenTypeEmbedding: Embedding<Scalar>
-    public var positionEmbedding: Embedding<Scalar>
+    public var tokenEmbedding: RegularizableEmbedding<Scalar>
+    public var tokenTypeEmbedding: RegularizableEmbedding<Scalar>
+    public var positionEmbedding: RegularizableEmbedding<Scalar>
     public var embeddingLayerNorm: LayerNorm<Scalar>
     @noDerivative public var embeddingDropout: Dropout<Scalar>
     public var embeddingProjection: [Dense<Scalar>]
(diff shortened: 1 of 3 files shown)


xihui-wu:
Thanks for the nice cleanup! Can you explain a bit the reason calling it RegularizableEmbedding?

texasmichelle:
I'm not convinced it's the best name, but it's better than `EmbeddingBERT` (as an analog to [`EmbeddingGPT2`](https://github.com/tensorflow/swift-models/blob/master/Models/Text/GPT2/TransformerLM.swift#L294) 😆 
I chose this because it differs from the standard `Embedding` in conformance to [`Regularizable`](https://github.com/tensorflow/swift-models/blob/master/Models/Text/WeightDecayedAdam.swift#L18).

xihui-wu:
I see. That's good for now. Do you have a sense later how to organize Embedding, RegularizableEmbedding and EmbeddingBERT these three ?

texasmichelle:
Ideally, they both use the standard `Embedding`, which may mean adding functionality to the core library. Not something I want to tackle right now, but it's for sure part of #434.

Issue #474, opened 2020-04-28
In the Neural Collaborative Filtering model, generalize to arbitrary layers and improve parameters


BradLarson:
Within the [recently added Neural Collaborative Filtering model](https://github.com/tensorflow/swift-models/pull/406), the number of layers is hardcoded, but is intended to be variable. To generalize this model, these layers should be built in an array, using `differentiableReduce()` to process through them.

Additionally, `mfDim` and `mfReg` [as parameter names](https://github.com/tensorflow/swift-models/blob/master/Models/Recommendation/NeuMF.swift#L58) aren't the most descriptive, and we should update this model and the matching example to have better naming for these parameters.

BradLarson:
Resolved via PR #501.

PR #501, opened 2020-05-09
In the Neural Collaborative Filtering model, generalize to arbitrary layers and improve parameters (#474)

AniketNavlur:
* Generalized the model by building the layers in an array and processing through `differentiableReduce()`

* Using more descriptive names for model parameters

* Updated NeuMF example to work with the new parameter names

### DIFF
--- a/Models/Recommendation/NeuMF.swift
+++ b/Models/Recommendation/NeuMF.swift
@@ -26,73 +26,71 @@ public struct NeuMF: Module {
     public typealias Scalar = Float
     @noDerivative public let numUsers: Int
     @noDerivative public let numItems: Int
-    @noDerivative public let mfDim: Int
-    @noDerivative public let mfReg: Scalar
+    @noDerivative public let numLatentFeatures: Int
+    @noDerivative public let matrixRegularization: Scalar
     @noDerivative public var mlpLayerSizes: [Int] = [64, 32, 16, 8]
-    @noDerivative public var mlpLayerRegs: [Scalar] = [0, 0, 0, 0]
+    @noDerivative public var mlpRegularizations: [Scalar] = [0, 0, 0, 0]
 
-    public var mfUserEmbed: Embedding<Scalar>
-    public var mfItemEmbed: Embedding<Scalar>
-    public var mlpUserEmbed: Embedding<Scalar>
-    public var mlpItemEmbed: Embedding<Scalar>
-    public var dense1: Dense<Scalar>
-    public var dense2: Dense<Scalar>
-    public var dense3: Dense<Scalar>
-    public var finalDense: Dense<Scalar>
+    public var mfUserEmbedding: Embedding<Scalar>
+    public var mfItemEmbedding: Embedding<Scalar>
+
+    public var mlpUserEmbedding: Embedding<Scalar>
+    public var mlpItemEmbedding: Embedding<Scalar>
+    public var mlpLayers: [Dense<Scalar>]
+
+    public var neuMFLayer: Dense<Scalar>
 
     /// Initializes a NeuMF model as per the dataset from the given hyperparameters.
     ///
     /// -Parameters
     /// - numUsers: Total number of users in the dataset.
     /// - numItems: Total number of items in the dataset.
-    /// - mfDim: Embedding size of the matrix factorization model.
-    /// - mfReg: Regularization for the matrix factorization embeddings.
+    /// - numLatentFeatures: Embedding size of the matrix factorization model.
+    /// - matrixRegularization: Regularization for the matrix factorization embeddings.
     /// - mlpLayerSizes: The sizes of the layers in the multi-layer perceptron model.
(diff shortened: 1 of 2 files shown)


Models/Recommendation/NeuMF.swift
```diff
@@ -26,95 +26,95 @@ public struct NeuMF: Module {
...
-        mfDim: Int,
-        mfReg: Float,
-        mlpLayerSizes: [Int],
-        mlpLayerRegs: [Float]
+        numUsers : Int,
+        numItems : Int,
+        mfFeatures : Int,
+        mfRegularization : Float,
+        mlpLayerSizes : [Int],
+        mlpRegularizations : [Float]
```
BradLarson, line 61: Prefixes on parameters, such as `mf`, generally aren't used in Swift. It isn't immediately clear what the prefix stands for in this case. My suggestion in the previous pull request was that `embeddingSize` and `regularization` might be better choices as names, but I'm open to others. `mlp` is a bit better known as an abbreviation in the field, so it's probably fine to leave as-is.

Models/Recommendation/NeuMF.swift
```diff
@@ -26,95 +26,95 @@ public struct NeuMF: Module {
...
-        let userEmbedMf = self.mfUserEmbed(userIndices)
-        let itemEmbedMf = self.mfItemEmbed(itemIndices)
+        let userEmbedMf = mfUserEmbed(userIndices)
+        let itemEmbedMf = mfItemEmbed(itemIndices)
 
-        // Concatenate MF and MLP parts
         let mfVector = userEmbedMf * itemEmbedMf
+
         var mlpVector = userEmbedMlp.concatenated(with: itemEmbedMlp, alongAxis: -1)
+         mlpVector = mlpLayers.differentiableReduce(mlpVector){ $1($0) }
```
BradLarson, line 112:
```suggestion
        mlpVector = mlpLayers.differentiableReduce(mlpVector){ $1($0) }
```

Models/Recommendation/NeuMF.swift
```diff
@@ -26,95 +26,95 @@ public struct NeuMF: Module {
...
-        // TODO: Extend it for n layers by using for loop
-        // Currently only for 4 layers
-        dense1 = Dense(
-            inputSize: self.mlpLayerSizes[0], outputSize: self.mlpLayerSizes[1], activation: relu)
-        dense2 = Dense(
-            inputSize: self.mlpLayerSizes[1], outputSize: self.mlpLayerSizes[2], activation: relu)
-        dense3 = Dense(
-            inputSize: self.mlpLayerSizes[2], outputSize: self.mlpLayerSizes[3], activation: relu)
-        finalDense = Dense(inputSize: (self.mlpLayerSizes[3] + self.mfDim), outputSize: 1)
+        for (inpSz, outSz) in zip(mlpLayerSizes, mlpLayerSizes[1...]) {
```
BradLarson, line 88: `inpSz` and `outSz` are a little hard to parse as names, and not what we'd typically see in Swift. Expanding them slightly to `inputSize` and `outputSize` might make this easier to follow.

Models/Recommendation/NeuMF.swift
```diff
@@ -26,95 +26,95 @@ public struct NeuMF: Module {
...
-        let userEmbedMlp = self.mlpUserEmbed(userIndices)
-        let itemEmbedMlp = self.mlpItemEmbed(itemIndices)
+        let userEmbedMlp = mlpUserEmbed(userIndices)
+        let itemEmbedMlp = mlpItemEmbed(itemIndices)
 
         // MF part
-        let userEmbedMf = self.mfUserEmbed(userIndices)
-        let itemEmbedMf = self.mfItemEmbed(itemIndices)
+        let userEmbedMf = mfUserEmbed(userIndices)
+        let itemEmbedMf = mfItemEmbed(itemIndices)
```
BradLarson, line 107: Would it make sense to expand the abbreviation of `Embed` to `Embedding`? Also, if we're going to use acronyms like MLP as a suffix, we should probably capitalize all letters of the acronym (we don't do this for prefixes, due to them starting with lowercase letters).

Models/Recommendation/NeuMF.swift
BradLarson: Looking at the paper, there's a clear distinction between the matrix factorization and multi-layer perceptron portions, so it can see why it might be useful to call those out separately. `mf` is still an uncommon acronym, though, so would we want to expand that to `matrixFactorization`? Again, I'm open to other suggestions.

AniketNavlur: I feel `matrixFactorizationRegularization` will be a too long variable name. How about `matrixRegularization` instead?

BradLarson: Those sound like reasonable naming choices, and if you want to go ahead with those and with the other small suggestions here, I'd be happy to bring this in as such.

PR #507, opened 2020-05-12
Converting benchmarks to Epochs, removing old Batcher CIFAR-10 dataset

BradLarson:
Continuing Sylvain's work in transitioning the Swift models from using Batcher-based datasets to the new Epochs API, this reworks the Benchmarks interfaces to use Epochs datasets. It removes the last dependency on the old Batcher-based CIFAR-10 dataset, so that can now be deleted.

By lifting the constant mean and standard deviation tensors out of the `makeBatch()` function and instead creating them once and passing them into `makeBatch()`, the CIFAR-10 dataset now benchmarks at the same performance as the previous Batcher implementation.

This pull request is ready for review.

Known issues to investigate later:

- Without placing a trivial eager-mode calculation at the beginning of the training benchmark, the eager runtime consumes all available GPU memory on initialization of the ResNet56 model and causes the X10 compiler to run out of memory during compilation. With this workaround in place, GPU memory peaks at 1.6 GB out of 8 GB on a GTX 1080. I'll create an issue to investigate this further.
- Switching to synthetic training data leads to a crash in the XLA compiler when targeting GPU devices. Results are correctly returned before this happens, and the synthetic benchmarks are currently unused for anything, so this doesn't impact the function of our benchmarks. A separate issue will be created to track this. The error is as follows:

> F tensorflow/compiler/xla/xla_client/xla_util.cc:90] Internal: From /job:localservice/replica:0/task:0:
> Unable to launch convolution with type forward and algorithm (1, none)
> 	 [[{{node XRTExecute}}]]
> Aborted (core dumped)


Benchmarks/Models/SyntheticImageDataset.swift
```diff
@@ -16,32 +16,69 @@ import Batcher
...
+    labels: Int,
+    dimensions: [Int],
+    entropy: Entropy,
+    device: Device
+  ) {
+    precondition(labels > 0)
+    precondition(dimensions.count == 3)
+
+    // Training data
+    let trainingSamples = [Int](repeating: 0, count: batchSize)
```
sgugger, line 56: A range `(0..<batchSize)` would use less memory, but that's nitpicking.

Benchmarks/Models/SyntheticImageDataset.swift
```diff
@@ -16,32 +16,69 @@ import Batcher
...
+          makeSyntheticBatch(samples: $0, dimensions: dimensions, labels: labels, device: device)
         }
+      }
 
-        training = Batcher<SourceDataSet>(
-            on: syntheticDataset, batchSize: batchSize, numWorkers: 1, shuffle: true)
-        
-        test = Batcher<SourceDataSet>(on: syntheticDataset, batchSize: batchSize, numWorkers: 1)
+    // Validation data
+    let validationSamples = [Int](repeating: 0, count: batchSize)
```
sgugger, line 65: Same here

Benchmarks/Models/SyntheticImageDataset.swift
BradLarson: Good suggestion. This was a quick-and-dirty implementation, and this part probably wouldn't impact performance by too much, but the use of a range at least simplifies the code. Done.

Benchmarks/Models/SyntheticImageDataset.swift
BradLarson: Implemented, as per above.

Issue #506, opened 2020-05-12
[WordSeg] Fix dropout issues


sgugger:
Currently dropout is not used on the embeddings in the encoder and the decoder because it breaks AD (at least that's what the comment say). See [here](https://github.com/tensorflow/swift-models/blob/a5e1b1d558b174a9590a66b2ffe6978faefff8df/Models/Text/WordSeg/Model.swift#L129) and [there](https://github.com/tensorflow/swift-models/blob/a5e1b1d558b174a9590a66b2ffe6978faefff8df/Models/Text/WordSeg/Model.swift#L175).

sgugger:
So in both case, the dropout should be applied on the matrix of hidden states (I think S4TF returns a list of hidden states and not a matrix batchSize x SequenceLength x hiddenDimension).

dan-zheng:
Dropout added in https://github.com/tensorflow/swift-models/pull/550.

PR #510, opened 2020-05-13
Fix some WordSeg AutoDiff todos.

dan-zheng:
- Inline stale workaround functions.
  - These functions were added to work around differentiation crashes.
    These crashes are no longer reproducible.
- Remove duplicate derivative for `RecurrentLayer.callAsFunction(_:initialState:)`.
  - This derivative wrt `(self, inputs, initialState)` was [upstreamed to
    tensorflow/swift-apis](https://github.com/tensorflow/swift-apis/blob/763fa17a91bdc6d241d123a923dbc0ea1229d808/Sources/TensorFlow/Layers/Recurrent.swift#L380-L422).

### DIFF
--- a/Models/Text/WordSeg/Model.swift
+++ b/Models/Text/WordSeg/Model.swift
@@ -124,11 +124,8 @@ public struct SNLM: EuclideanDifferentiable, KeyPathIterable {
   /// Returns the hidden states of the encoder LSTM applied to the given sentence.
   public func encode(_ x: CharacterSequence) -> [Tensor<Float>] {
     let embedded = drop(embEnc(x.tensor))
-    // TODO: If I inline `makeEncoderInput`, it breaks AD.
-    let encoderStates = lstmEnc(makeEncoderInput(embedded))
-    // TODO: Need to add dropout here, but it breaks AD.
-    // TODO: If I inline `computeEncoderResult`, it breaks AD.
-    return computeEncoderResult(encoderStates)
+    let encoderStates = lstmEnc(embedded.unstacked().differentiableMap { $0.rankLifted() })
+    return encoderStates.differentiableMap { $0.hidden.squeezingShape(at: 0) }
   }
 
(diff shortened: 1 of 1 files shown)


Models/Text/WordSeg/Model.swift
```diff
@@ -347,56 +344,10 @@ public struct MLP: Layer {
...
-func rankLift(_ x: Tensor<Float>) -> Tensor<Float> {
-  return x.rankLifted()
-}
-
-@differentiable
-func makeEncoderInput(_ x: Tensor<Float>) -> [Tensor<Float>] {
-  return x.unstacked().differentiableMap(rankLift)
-}
-
 // TODO: Move this derivative into tensorflow-apis
```
marcrasi, line 347: fyi, I think I did move this into tensorflow-apis (https://github.com/tensorflow/swift-apis/pull/663), so you should be able to completely delete this extension

dan-zheng: Thank you, done!

Models/Text/WordSeg/Model.swift
```diff
@@ -124,11 +124,8 @@ public struct SNLM: EuclideanDifferentiable, KeyPathIterable {
   /// Returns the hidden states of the encoder LSTM applied to the given sentence.
   public func encode(_ x: CharacterSequence) -> [Tensor<Float>] {
     let embedded = drop(embEnc(x.tensor))
-    // TODO: If I inline `makeEncoderInput`, it breaks AD.
-    let encoderStates = lstmEnc(makeEncoderInput(embedded))
-    // TODO: Need to add dropout here, but it breaks AD.
```
texasmichelle, line 129: Do we still need dropout between these layers?

dan-zheng: I believe so. https://github.com/tensorflow/swift-models/issues/506 tracks re-adding dropout.

PR #512, opened 2020-05-13
[WordSeg] Add file download to dataset

### LINKED ISSUE #500: [WordSeg] Add dataset
Add a dataset for use with the WordSeg model. With a bit of modification, `TextUnsupervised` could be a good fit.

Clean up structures like `DataSet` and `Alphabet` in the model.

texasmichelle:
Fixes #500 

Move dataset tests into Tests/DatasetTests
Add test for file download initializer


Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -79,6 +88,31 @@ public struct WordSegDataset {
     return nil
   }
 
+  public init?() throws {
```
BradLarson, line 91: Does this initializer and the one below it need to be optional? It seems like we'd cover the error cases just by them being able to throw errors. I don't think I see a case where we return nil in the initialization code, but I could be missing something.

texasmichelle: SGTM and is consistent with other datasets 👍

PR #526, opened 2020-05-14
[WordSeg] Add example

texasmichelle:


texasmichelle:
The test failure is indeed puzzling. My hunch is that kokoro was using a cached image, so I'm curious to see whether it passes without any changes to `TextUnsupervised`.

Do we have an open issue for training loop creation? Would that be here or in another repo?

texasmichelle:
I couldn't get past the first epoch with the full dataset, so I added support for a smaller dataset, which trains quickly. The full dataset gives us our performance target and the flexibility to use a smaller sample will help get us there.

PR #531, opened 2020-05-15
Adding a ResNet50 benchmark with synthetic ImageNet-sized imagery

BradLarson:
This adds the beginnings of a ResNet50 v1.5 image classification benchmark on ImageNet, starting with synthetic data only due to the lack of ImageNet as a full dataset right now. This also fixes how synthetic data preferences are read from defaults and cleans up a few parameters in the Package.swift file.

The default batch size of 128 just fits into GPU memory with X10 on a V100 instance, matching the batch sized used in tf.keras benchmarks. The eager pathway can only support up to a batch size of 64 on the same GPU, as opposed to the size of 120 used for tf.keras eager benchmarks.

Additionally, this removes the hardcoded epoch sizes in favor of a new enum type that contains either batches or epochs as the duration of the benchmark. This was enabled by the move to Epochs for the datasets, which now have epochs that you can iterate over independently of the batches.


Benchmarks/Models/ResNetImageNet.swift
```diff
@@ -0,0 +1,78 @@
...
+import TensorFlow
+
+// This benchmark currently is purely synthetic, as we don't yet have ImageNet as a real dataset.
+enum ResNetImageNet: BenchmarkModel {
+    static var name: String { "ResNetImageNet" }
+
+    static func examplesPerEpoch(for variety: BenchmarkVariety) -> Int {
+        switch(variety) {
+        case .inferenceThroughput: return 150000
+        case .trainingThroughput: return 1200000
```
shabalind, line 26: Is there any way not to hardcode those numbers, but compute them? (I am afraid of it getting unintentionally out of sync.)

BradLarson: So this caused me to pull on a thread, and as a result I've completely removed the hardcoded epoch counts in favor of a new .batches(value) / .epochs(value) enum that takes advantage of Epoch's ability to iterate on epochs as well as individual batches. It expands the scope of the PR, but I like how it simplifies the benchmarks.

shabalind: 👍

PR #533, opened 2020-05-15
Add Handwriting Synthesis model example

tanmayb123:
This adds a demo of using Swift for TensorFlow to generate handwriting from text automatically. For example: https://imgur.com/a/xpGGfcU

Code based off of: https://github.com/t-vi/pytorch-tvmisc/blob/master/misc/graves_handwriting_generation.ipynb
Original paper: https://arxiv.org/abs/1308.0850
Pretrained model from: https://github.com/t-vi/pytorch-tvmisc/releases

There's still a lot of work to be done. I'm planning on refactoring the code a bit, making it cleaner, easier to read through. However, I did want to share this as a draft to get some high level feedback before I get some more low level, nitty-gritty feedback after my refactor.
I also need to add a readme, some documentation, etc.

I'd appreciate any thoughts/ideas.


BradLarson:
At a high level, even breaking this up into separate files would do a lot to help code readability, but I'm guessing you have that planned.

For the weights, if we could convert those to a TensorFlow v2 checkpoint format (or SavedModel), we could use the built-in CheckpointReader to load them and avoid the h5py dependency. We'd be able to host the weights on our bucket, once converted.

tanmayb123:
Glad to _be_ back!

I'm working on an overhaul of the model & code which should generate better output and will also be a lot cleaner.

tanmayb123:
Hey, sorry for the delay. Should be able to submit the cleanup by tonight.

tanmayb123:
Finished. This PR is ready for review now. I've been busy with a product release for some time but should be available now to incorporate feedback from the review process.

BradLarson:
I hope you don't mind, but I just pushed a change that reworks the Python-based checkpoint loading into a function that uses our CheckpointReader and a hosted TF checkpoint archive. This eliminates the need for the manual model downloading, as well as having the h5py module in your local Python setup.

I've also added a Swift Package Manager target for this example, which was missing. The example itself appears to need two command-line arguments to be supplied or the example crashes, so you will probably want to either warn people about their need if missing or provide defaults. I'd highly recommend creating a Readme file that explains the model and its usage, which could be where you'd provide good example inputs.

I added a function to write out the image when done, because the visualize() function didn't work at the command line on my workstation. The image that was written for command-line options of `"text" 0.5` was all white, so I don't know if I picked bad values.

BradLarson:
Checking back in with this model, I think we might be ready to do a final review and bring this in. However, I think we'll need a Readme to explain the proper usage of the example, as well as sample options that can be used. As I pointed out earlier, I'd like to make sure that I was using the right options, because I'm currently getting an all-white image as output from the model with the above settings.

tanmayb123:
Thanks for the additions, Brad! I'll go ahead and test out the code to make sure it works, and write the readme.

BradLarson:
It's been a little while, so I'm checking in on this PR. Do you still have an interest in finishing the Readme and verifying correct operation? I can't seem to make this work using the random parameters I've tried, so any documentation that can allow me to verify that the model is functional would be a huge step forward, and should let us pull this in (assuming the model is working right).

BradLarson:
I wanted to check in again and see if you'd been able to make any progress on the documentation for the model so that we can test it an incorporate it into the repository. It would be a great model to have, but I unfortunately can't verify its correct operation right now. If you don't have the time right now to work on this, I totally understand.

We might close a few of these draft PRs if they remain inactive, and then can revisit them when they are ready to move forward.

shabalind:
Closing for now. Feel free to reopen if there is more progress!

PR #534, opened 2020-05-15
Add Support for ELECTRA

Shashi456:
This PR adds support for ELECTRA, the current state of the art in pretrained language models.
Source: ["ELECTRA: Pre-training Text Encoders as Discriminators Rather Than Generators"]( https://arxiv.org/abs/2003.10555 ) 

Electra has two parts: a generator and a discriminator, This PR adds support for the discriminator (which is the only relevant for finetuning and downstream tasks), I'll add support for the generator in a follow-up PR.
image

This works without any additional model changes because the discriminator has exactly the same model as BERT and other variants and the tokenization scheme is the same. 

I ran into two issues:
1] Although the `hidden_size = embedding_size` generally for Electra, but in the case of `electra_small`, the `embedding size` is `128 `while the `hidden size` is `256`, any idea how I could solve in that particular case?
2] Loading the checkpoint, when I looked at the directory, this is what I get any idea. how I could use the checkpoint reader to load this?
image

Ideally, the ultimate goal would be to, add support for the generator and try to actually train an `electra_small` model, which can be done on a single gpu within a day.


brettkoonce:
very cool!

BradLarson:
It looks like this does not build cleanly, at least with swift-models head and the currently nightly toolchain. There appear to be errors on lines 205, 580, 582, 584, and 603 in Bert.swift, largely dealing with the new enum case. Could you fix those and verify this builds cleanly locally?

Shashi456:
@bradlarson, could you help me with the doubts I posted in the top message :P?

BradLarson:
I can't answer your first question, but for the second, the checkpoint directory you describe should be readable by the CheckpointReader that's contained within swift-models. This model currently uses the TensorFlowCheckpointReader in swift-apis that lacks some of this functionality. PR #554 starts to address this by replacing the old TensorFlowCheckpointReader with the newer CheckpointReader. That should make it easy to read this checkpoint.

xihui-wu:
@Shashi456 I'm trying to answer you first question -- https://github.com/tensorflow/swift-models/blob/master/Models/Text/BERT.swift#L398 You should be able to do same way with Albert

dabrahams:
Bump; we're really interested in this PR. Can you rebase it on master so we can land it?

Thanks!

Shashi456:
@dabrahams @shabalind. I'll use the new checkpoint and all other changes asap and push it by tomorrow. Sorry for the delay

Shashi456:
I seem to have done some errors while rebasing on top of master, I'll start a new pull request on a fresh branch and close this. Sorry!

PR #486, opened 2020-05-04
CI: add swift-format based linting/auto-formatting

compnerd:
Introduce swift-format based auto-formatting of PRs on the repository
via GitHub actions.  This will automatically perform the required style
changes and commit them to the PR.

PR #535, opened 2020-05-15
WordSeg: use Tensors everywhere

sgugger:
Test using Tensor of floats everywhere to avoid copying back to the CPU needlessly.
This leads to an AD failure (or I did something stupid) since the model does not produce any gradients anymore. Any help welcome :)


Models/Text/WordSeg/Lattice.swift
```diff
@@ -232,7 +232,7 @@ extension Lattice.Node {
 
 extension Lattice.Edge {
   public func isAlmostEqual(to other: Self, tolerance: Float) -> Bool {
-    let diffP = abs(self.logp - other.logp)
+    let diffP = abs(self.logp - other.logp).scalarized()
```
texasmichelle, line 235: This line is only necessary because `logp` was previously `Float`. Now that it's `Tensor<Float>`, the check in the return statement can use `Tensor.isAlmostEqual`, similar to `self.score` and `self.totalScore`.

Models/Text/WordSeg/SemiRing.swift
```diff
@@ -105,8 +97,8 @@ extension SemiRing {
   // TODO(abdulras) see if we can use ulp as a default tolerance
   @inlinable
   public func isAlmostEqual(to other: Self, tolerance: Float) -> Bool {
-    let diffP = abs(self.logp - other.logp)
-    let diffR = abs(self.logp - other.logp)
+    let diffP = abs(self.logp - other.logp).scalarized()
+    let diffR = abs(self.logp - other.logp).scalarized()
```
texasmichelle, line 101: Similarly, these can also use `Tensor.isAlmostEqual`

sgugger:
I don't have the ability to run the swift-format used in google3, so unless the instructions to build it have become easier than last time I checked, it would be easier if you could run it through the branch.

(I thought swift-format was now auto-triggered?)

Models/Text/WordSeg/Lattice.swift
sgugger: I did the quick fix for the merge conflict but this makes more sense, thanks!

Models/Text/WordSeg/SemiRing.swift
sgugger: Same as above.

Examples/WordSeg/main.swift
```diff
@@ -72,15 +72,15 @@ for epoch in 1...maxEpochs {
...
+    let (loss, gradients) = valueWithGradient(at: model) { model -> Tensor<Float> in
       let lattice = model.buildLattice(sentence, maxLen: maxLength)
       let score = lattice[sentence.count].semiringScore
       let expectedLength = exp(score.logr - score.logp)
       let loss = -1 * score.logp + lambd * expectedLength
       return loss
     }
 
-    trainingLossSum += loss
+    trainingLossSum += loss.scalarized()
```
BradLarson, line 83: We may want to have the training and validation loss sums be expressed as Tensor<Float> to avoid the `scalarized()` usage here, to make this a little easier on X10. It's a small thing compared to the other challenges this presents to X10, but it's a quick fix.

Models/Text/WordSeg/Lattice.swift
```diff
@@ -54,13 +54,13 @@ public struct Lattice: Differentiable {
         SemiRing(
           logp: logp,
           // TODO(abdulras): this should really use integeral pow
-          logr: logp + logf(powf(Float(sentence.count), Float(order))))
+          logr: logp + Tensor(logf(powf(Float(sentence.count), Float(order)))))
```
BradLarson, line 57: This isn't the only case of this, but to make sure of device consistency in creation of new tensors, we should probably add `on: logp.device` so that this is at least the same device as the incoming Tensor argument. Otherwise, it will always be on the default TF eager device.

Models/Text/WordSeg/Model.swift
```diff
@@ -197,10 +187,9 @@ public struct SNLM: EuclideanDifferentiable, KeyPathIterable {
   }
 
   // MARK: - buildLattice
-
-  func get_logp_lex(_ logp_lex: [Float], _ candidate: CharacterSequence) -> Float {
+  func get_logp_lex(_ logp_lex: Tensor<Float>, _ candidate: CharacterSequence) -> Tensor<Float> {
     guard let index = parameters.strVocab.dictionary[candidate] else {
-      return -Float.infinity
+      return Tensor(-Float.infinity)
```
BradLarson, line 192: Same as above, maybe specifying `on: logp_lex.device` would aid in compatibility.

texasmichelle:
I ran lint and it came back clean 👍

texasmichelle:
When #486 is merged, linting should be automatic 🎉

sgugger:
Eh those random failures are annoying in the sense they are not always giving the same error message! Thanks for unblocking me :)

sgugger:
As discussed, I'll add @BradLarson optimizations in a separate PR.

texasmichelle:
Yeah the tests have been picking up old images - in this case, the failure had nothing to do with your changes. Thank you for these updates!

PR #538, opened 2020-05-15
[WordSeg] Add validation

texasmichelle:


Examples/WordSeg/main.swift
```diff
@@ -92,10 +92,26 @@ for epoch in 1...maxEpochs {
     }
   }
 
+  Context.local.learningPhase = .inference
+  var validationLossSum: Float = 0
+  var validationBatchCount = 0
+  var validationCharacterCount = 0
+  for sentence in dataset.validation! {
```
saeta, line 99:
```suggestion
  guard let validationDataset = dataset.validation else { return }
  Context.local.learningPhase = .inference
  var validationLossSum: Float = 0
  var validationBatchCount = 0
  var validationCharacterCount = 0
  for sentence in validationDataset {
```

This avoids crashing when the validation dataset isn't provided. WDYT?

texasmichelle: Yes, this is a good edit - thank you!

PR #540, opened 2020-05-16
Loading multiple GPT2 checkpoints on same computer.

leoxzhao:
GPT2 initializer tries to load parameters by directly referencing `Transformer` directory under temporary directory, which happens to be where CheckpointReader keeps those files if modelName is `Transformer`. This works if there is only one checkpoint on the system, for example 117M. Using GPT2 with other checkpoints will break with shape errors.

```
Fatal error: Input to reshape is a tensor with 5120 values, but the requested shape has 5100: file /Users/swiftninjas/s4tf/tensorflow-swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 301
```
This change uses the directory name as part of the modelName parameter of CheckpointReader, for example: GPT2-117M, GPT2-345M to avoid mixing files. It also uses CheckpointReader's localCheckpointLocation to get the directory of the locally stored checkpoint files.

### DIFF
--- a/Models/Text/GPT2/GPT2.swift
+++ b/Models/Text/GPT2/GPT2.swift
@@ -49,7 +49,6 @@ public class GPT2 {
                 "model.ckpt.meta",
                 "vocab.bpe",
             ]
-            let FS: FileManager = FileManager.default
 
             let reader: CheckpointReader = try CheckpointReader(
                 checkpointLocation: checkpoint,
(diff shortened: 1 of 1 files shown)


Models/Text/GPT2/GPT2.swift
```diff
@@ -50,15 +50,16 @@ public class GPT2 {
                 "vocab.bpe",
             ]
             let FS: FileManager = FileManager.default
```
texasmichelle, line 52: This line can also be removed since `FS` is not used elsewhere.

leoxzhao: done.

PR #541, opened 2020-05-16
Add Support for SNLI 

Shashi456:
I've checked the individual components and they all work, I've adopted the code mostly from the CoLA dataset.

Hope to flesh it out in properly.


ematejska:
We'll review once it's out of draft unless you're asking for a review now. Thanks!

PR #542, opened 2020-05-18
[WordSeg] Train until loss stops improving.

texasmichelle:
Remove fixed number of training epochs. Instead, continue training until validation loss has not improved for 5 consecutive epochs.

### DIFF
--- a/Examples/WordSeg/main.swift
+++ b/Examples/WordSeg/main.swift
@@ -22,7 +22,9 @@ let ndim = 512  // Hidden unit size.
 // Training flags
 let dropoutProb = 0.5  // Dropout rate.
 let order = 5  // Power of length penalty.
-let maxEpochs = 10  // Maximum number of training epochs.
+let maxEpochs = 1000  // Maximum number of training epochs.
+var validationLossHistory = Array<Float>() // Keep track of validation loss.
+var noImprovements = 0 // Consecutive epochs without improvements to loss.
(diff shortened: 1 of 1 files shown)


Examples/WordSeg/main.swift
```diff
@@ -93,12 +95,24 @@ for epoch in 1...maxEpochs {
...
+
     print(
       """
       [Epoch \(epoch)] \
-      Training loss: \(trainingLossSum / Float(trainingBatchCount))
+      Training loss: \(trainingLoss))
       """
     )
+
+    // Stop training when loss stops improving.
```
sgugger, line 107: The training loss should not be considered for the criterion, only the validation loss (they also only use the validation loss in the PyTorch implementation).

Examples/WordSeg/main.swift
```diff
@@ -93,12 +95,24 @@ for epoch in 1...maxEpochs {
...
+    lossHistory.append(trainingLoss)
```
sgugger, line 108: There should be two different arrays for training and validation loss as you can't compare them directly.

texasmichelle: This will only be executed if there is no validation dataset. I added this to protect from runaway training times.

sgugger: Ah! Didn't catch the guard above in the diff. It works then!

PR #543, opened 2020-05-18
[GPT2] Add generation unwrapping protection

texasmichelle:
Every once in a while, GPT2-Inference will crash with this error:
```
Illegal instruction (core dumped)
```

This change adds unwrapping protection in case something goes wrong during generation.

### DIFF
--- a/Models/Text/GPT2/GPT2.swift
+++ b/Models/Text/GPT2/GPT2.swift
@@ -20,8 +20,9 @@ public class GPT2 {
     public static let remoteCheckpoint: URL =
         URL(string: "https://storage.googleapis.com/gpt-2/models/117M/model.ckpt")!
 
-    enum GPT2Error: Error {
+    public enum GPT2Error: Error {
         case invalidEncoding(id: Int32)
+        case invalidSeed(seed: Tensor<Int32>)
     }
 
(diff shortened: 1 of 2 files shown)


Models/Text/GPT2/GPT2.swift
```diff
@@ -154,7 +154,7 @@ public class GPT2 {
             randomCategorialLogits: logits.squeezingShape(at: 1),
             sampleCount: 1)
 
-        let id = Int32(seed[0][0])!
+        guard let id = Int32(seed[0][0]) else { return "" }
```
BradLarson, line 157: In the cases where a seed isn't generated correctly, is it preferred to quietly return an empty string over throwing a GPT2Error of some kind?

texasmichelle: that's a good point. We can throw here and continue from main instead.

texasmichelle:
I'm still seeing `Illegal instruction (core dumped)`, but this will hopefully aid in eventually locating the root cause.

compnerd:
@texasmichelle I'm not sure I follow how this helps in locating the root cause?  This hides the error, previously, we would actually hit the unreachable codepath, allowing us to understand whats going on in in a debugger

PR #544, opened 2020-05-18
[BERT] Remove custom optimizer protocol

texasmichelle:
The `WeightDecayedAdam` optimizer uses a custom protocol that overwrites the [swift-apis definition](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Optimizers/Optimizer.swift#L18).

This change removes the custom protocol and updates `WeightDecayedAdam` to conform to the standard `Optimizer`.


Models/Text/WeightDecayedAdam.swift
```diff
@@ -102,7 +93,8 @@ where
         precondition(0 <= beta1 && beta1 <= 1, "Beta parameter must be between 0 and 1")
         precondition(0 <= beta2 && beta2 <= 1, "Beta parameter must be between 0 and 1")
 
-        self.learningRate = learningRate
+        self.scheduledLearningRate = learningRate
```
xihui-wu, line 96: How about changing this parameter name to scheduledLearningRate ?

Models/Text/WeightDecayedAdam.swift
```diff
@@ -102,7 +93,8 @@ where
...
+        self.learningRate = self.scheduledLearningRate(forStep: step)
```
xihui-wu, line 97: Looks like self.learningRate is not used anywhere? Or you meant to use it in line 132? If so probably can add a property observer to step to update it.

texasmichelle: Good catch! I removed the local var on line 132.
I don't see how to use a property observer here, since `step` is a method parameter. Could you give more details about what you have in mind?

xihui-wu: I mean this, defining 'step' as follows, so that learningRate will be updated whenever step is increased by 1.
public var step: UInt64 = 0 {
    didSet {
        learningRate = scheduledLearningRate(forStep: step)
    }
}

texasmichelle: oh I see, how cool! Never used that language feature before :)

texasmichelle: Do we still need this line in `init` or will the property observer kick in when `step` is initialized?

texasmichelle: (I get a compiler error otherwise, so I left it)

xihui-wu: Ah I see. Probably because learningRate is a stored property and required to be explicitly initialized. Or define it as computed property? https://docs.swift.org/swift-book/LanguageGuide/Properties.html

texasmichelle: It's part of the `Optimizer` protocol, so no redefining

texasmichelle: Unless... I wonder if that would still conform. Trying it now!

texasmichelle: No, they have different type signatures.

Models/Text/WeightDecayedAdam.swift
```diff
@@ -136,7 +133,6 @@ where
         let denominator = Model.TangentVector.sqrt(secondMoments).adding(epsilon)
         let weightDecay = model.regularizationValue.scaled(by: weightDecayRate)
         let update = firstMoments ./ denominator + weightDecay
-        var learningRate = self.learningRate(forStep: step)
         if useBiasCorrection {
             let step = Float(self.step)
             learningRate *= sqrtf(1 - powf(beta2, step)) / (1 - powf(beta1, step))
```
texasmichelle, line 138: As cool as it is, I'm not crazy about the property observer approach because it could have unintended consequences if this line is executed multiple times without the step being incremented. Since there's a `step += 1` at the start of this function, that's not possible, but if that changes over time this would be a tricky bug to find.

xihui-wu: Sure, make sense to me not using property observer. The other reason is that we don't want to make it half-way updated in observer and half-way updated in update function.

texasmichelle: Easy fix: moving this assignment to the property observer 🎉

PR #546, opened 2020-05-19
[BERT] Remove custom embedding

texasmichelle:
Replace custom `Embedding` with the standard layer from swift-apis.


Models/Text/WeightDecayedAdam.swift
```diff
@@ -36,6 +36,12 @@ extension LayerNorm: Regularizable {
     }
 }
 
+extension Embedding: Regularizable {
```
xihui-wu, line 39: How about moving extensions of Embedding together with Dense and LayerNorm to BERT.swift, and leave this WeightDecayedAdam.swift self-contained ?

texasmichelle: That's a good idea. I moved everything except the custom optimizer, since it will go away with #544

PR #547, opened 2020-05-19
[WordSeg] Reduce learning rate on plateau

texasmichelle:
Automatically reduce the learning rate when training loss stagnates.

Move training termination code into a function to reduce clutter in the main loop.

### DIFF
--- a/Examples/WordSeg/main.swift
+++ b/Examples/WordSeg/main.swift
@@ -187,18 +187,16 @@ func reduceLROnPlateau(
   lossHistory: [Float], optimizer: Adam<SNLM>,
   factor: Float = 0.25
 ) {
-  let patience = 1
   let threshold: Float = 1e-4
   let minDecay: Float = 1e-8
-  if lossHistory.count <= patience { return }
-  guard let loss = lossHistory.last else { return }
+  if lossHistory.count < 2 { return }
+  let window = Array(lossHistory.suffix(2))
+  guard let previous = window.first else { return }
+  guard let loss = window.last else { return }
 
-  let window = Array(lossHistory.suffix(patience + 1))
-
-  if loss >= window.min()! * (1 - threshold) && loss != window.max() { return }
-
-  let newLR = optimizer.learningRate * factor
-  if newLR > minDecay {
-    optimizer.learningRate = newLR
-  }
+  if loss <= previous * (1 - threshold) { return }
+    let newLR = optimizer.learningRate * factor
+    if optimizer.learningRate - newLR > minDecay {
+      optimizer.learningRate = newLR
+    }
 }
--- a/Examples/WordSeg/main.swift
+++ b/Examples/WordSeg/main.swift
@@ -188,12 +188,13 @@ func reduceLROnPlateau(
   factor: Float = 0.25
 ) {
   let patience = 1
+  let threshold: Float = 1e-4
   if lossHistory.count <= patience { return }
   guard let loss = lossHistory.last else { return }
 
   let window = Array(lossHistory.suffix(patience + 1))
 
-  if window.min() == loss && window.max() != loss { return }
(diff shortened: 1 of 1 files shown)


Examples/WordSeg/main.swift
```diff
@@ -163,3 +164,36 @@ func usage() -> Never {
...
+  lossHistory: [Float], optimizer: Adam<SNLM>,
+  factor: Float = 0.25
+) {
+  let patience = 1
+  if lossHistory.count <= patience { return }
+  guard let loss = lossHistory.last else { return }
+
+  let window = Array(lossHistory.suffix(patience + 1))
+
+  if window.min() == loss && window.max() != loss { return }
```
sgugger, line 196: The PyTorch implementation uses a threshold of 1e-4 to determine if there is an improvement: a is considered a better loss than b if a < b * (1-threshold) so here, there should be a similar test between `window.max()` and `loss` (probably `&& window.max() * (1-threshold) > loss`).

Examples/WordSeg/main.swift
```diff
@@ -163,3 +164,36 @@ func usage() -> Never {
...
+
+  optimizer.learningRate = max(0.0, optimizer.learningRate * factor)
```
sgugger, line 198: The PyTorch implementation has a minimal decay of 1e-8: if `optimizer.learningRate - optimizer.learningRate * factor <= 1e-8` nothing is done.

Examples/WordSeg/main.swift
texasmichelle: Done

Examples/WordSeg/main.swift
texasmichelle: Done

Examples/WordSeg/main.swift
```diff
@@ -163,3 +164,41 @@ func usage() -> Never {
...
+  let minDecay: Float = 1e-8
+  if lossHistory.count <= patience { return }
+  guard let loss = lossHistory.last else { return }
+
+  let window = Array(lossHistory.suffix(patience + 1))
+
+  if loss >= window.min()! * (1 - threshold) && loss != window.max() { return }
+
+  let newLR = optimizer.learningRate * factor
+  if newLR > minDecay {
```
sgugger, line 201: It should be `oldLR - newLR > minDecay` here (sorry if I was unclear before).

Examples/WordSeg/main.swift
```diff
@@ -163,3 +164,41 @@ func usage() -> Never {
...
+) {
+  let patience = 1
+  let threshold: Float = 1e-4
+  let minDecay: Float = 1e-8
+  if lossHistory.count <= patience { return }
+  guard let loss = lossHistory.last else { return }
+
+  let window = Array(lossHistory.suffix(patience + 1))
+
+  if loss >= window.min()! * (1 - threshold) && loss != window.max() { return }
```
sgugger, line 198: The first part of the test is always true: `window.min() ! * (1 - threshold) < window.min()! <= loss`.
The threshold is to consider there has been an improvement: we say newLoss is better than anything before if
`newLoss <= oldLoss * (1-threshold)` otherwise we consider there has not really been an improvement. 

Here we want to return if there has been an improvement during the window of time. I think that simply means `window.min()! <= window.max() * (1-threshold)` (also not sure why min needs a ! but not max?) If that test is false, it means all the loss in the window did not significantly improve compared to the threshold.

texasmichelle: In this case, I believe `window.max()` does not require `!` because it is valid to compare an optional, whereby `window.min()` requires unwrapping to be used with `*`.

I attempted to clarify this function by assigning current and previous loss explicitly, rather than using min and max over a window (which was originally intended to support variable window sizes/patience values, but added too much complexity).

PR #550, opened 2020-05-19
[WordSeg] Add dropout to SNLM.

dan-zheng:
Add dropout to `SNLM.encode` and `SNLM.decode`.
Make all dropout invocations more visible using local variable mutations.

### DIFF
--- a/Models/Text/WordSeg/Model.swift
+++ b/Models/Text/WordSeg/Model.swift
@@ -124,9 +124,13 @@ public struct SNLM: EuclideanDifferentiable, KeyPathIterable {
 
   /// Returns the hidden states of the encoder LSTM applied to the given sentence.
   public func encode(_ x: CharacterSequence) -> [Tensor<Float>] {
-    let embedded = dropout(encoderEmbedding(x.tensor))
+    var embedded = encoderEmbedding(x.tensor)
+    embedded = dropout(embedded)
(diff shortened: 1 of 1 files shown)


Models/Text/WordSeg/Model.swift
```diff
@@ -124,9 +124,13 @@ public struct SNLM: EuclideanDifferentiable, KeyPathIterable {
...
   public func encode(_ x: CharacterSequence) -> [Tensor<Float>] {
-    let embedded = dropout(encoderEmbedding(x.tensor))
+    var embedded = encoderEmbedding(x.tensor)
+    embedded = dropout(embedded)
     let encoderStates = encoderLSTM(embedded.unstacked().differentiableMap { $0.rankLifted() })
-    return encoderStates.differentiableMap { $0.hidden.squeezingShape(at: 0) }
+    var encoderResult = Tensor(
+      stacking: encoderStates.differentiableMap { $0.hidden.squeezingShape(at: 0) })
+    encoderResult = dropout(encoderResult)
+    return encoderResult.unstacked()
```
dan-zheng, line 133: Note: this code looks convoluted because it performs lots of tensor stacking/unstacking.

This is because recurrent layer APIs take and return `[Tensor<Scalar>]`, while operations like `dropout` take and return `Tensor<Scalar>`. We could consider adding recurrent layer APIs that take and return stacked `Tensor<Scalar>` to improve usability.

PR #553, opened 2020-05-19
WordSeg benchmarks

BradLarson:
This adds three distinct benchmark cases for the WordSeg model: `WordSegScore`, `WordSegScoreAndGradient`, and `WordSegViterbi`. The test sentence length for these benchmarks is specified via the `batchSize` parameter. The `--training` or `--inference` parameter is ignored for these benchmarks, but one still must be specified due to the current architecture.

For example, to run a score and gradient benchmark with a test sentence length of 14:

```
swift run -c release Benchmarks measure --benchmark WordSegScoreAndGradient --batchSize 14 --training
```

This is ready for review for correctness.


Models/Text/WordSeg/Lattice.swift
```diff
@@ -129,9 +129,9 @@ public struct Lattice: Differentiable {
     self.positions = positions
   }
 
-  mutating func viterbi(sentence: String) -> [Edge] {
+  public mutating func viterbi(sentence: CharacterSequence) -> [Edge] {
```
texasmichelle, line 132: Considerations for rebasing: `CharacterSequence` might make more sense here as an input parameter. Another option is leaving it as `String` with whitespace removal inside this method, or overloading to support both.

Benchmarks/Models/WordSeg.swift
```diff
@@ -0,0 +1,178 @@
...
+
+    static func score(model: SNLM, sentence: CharacterSequence) {
+        let lattice = model.buildLattice(sentence, maxLen: maximumSequenceLength)
+        let score = lattice[sentence.count].semiringScore
+        let _ = score.logr + score.logp
+    }
+
+    static func scoreAndGradient(model: SNLM, sentence: CharacterSequence) {
+        let lambd: Float = 0.00075
+        var mutableModel = model
```
texasmichelle, line 163: Does the model need to be mutable?

Benchmarks/Models/WordSeg.swift
```diff
@@ -0,0 +1,178 @@
...
+
+        let (loss, gradients) = valueWithGradient(at: mutableModel) { model -> Tensor<Float> in
```
texasmichelle, line 165: Do we get more realistic benchmarks by including this assignment, even though `loss` and `gradients` aren't used (I noticed this in `viterbi()` as well)?

BradLarson: For eager mode, I don't think it makes a difference, everything just runs eagerly. For X10, we may need to use the values in some capacity, because for image classification benchmarks if the gradient isn't touched after calculation the trace won't run through the backwards pass. I left this in there for now, because I wasn't sure how X10 would optimize around this.

texasmichelle: That makes sense for X10. We'll likely have to add a scalarize statement just afterwards. To get rid of the warnings for now, how about using underscores?

Benchmarks/Models/WordSeg.swift
texasmichelle: (this also throws a warning)

Models/Text/WordSeg/Lattice.swift
BradLarson: I've kept CharacterSequence here and changed the input in the Viterbi function in the validation phase for the WordSeg example to use `record.numericalizedText`. I'm hoping that's functionally correct.

PR #554, opened 2020-05-20
[BERT] Switch to use CheckpointReader to load tensors from checkpoint.

leoxzhao:
Switch to use CheckpointReader to load tensors from checkpoint (it was using TensorFlow API directly). Also add a simple command line parameter for the example app, BERT-CoLA, which takes "BERT", "ALBERT", "RoBERTa" to switch between models.

`swift run BERT-CoLA RoBERTa` ends up with this error.

```
Training BERT for the CoLA task!
[Epoch 1]
Fatal error: Input to reshape is a tensor with 1 values, but the requested shape has 786432: file /Users/danielzheng/swift-debug/tensorflow-swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 301
Illegal instruction: 4
```

This happens in `master` as well. So, it is not an issue introduced by the change.

BERT base and ALBERT base can start training.


leoxzhao:
I made this change to load different models easier and also verify different language models. I think BERT is in best shape. ALBERT's checkpoint is pointing at an TfHub checkpoint, therefore needs some special treatment. I don't know how RoBERTa's checkpoint was generated. The original checkpoint was in PyTorch format.

I open this PR to see if those code changes can be useful to anyone. It is fine to leave it out of master.

Support/Checkpoints/CheckpointReader.swift
```diff
@@ -70,9 +70,21 @@ open class CheckpointReader {
         let checkpointBase = finalCheckpointLocation.lastPathComponent
         let indexReader: CheckpointIndexReader
         if finalCheckpointLocation.isFileURL {
-            self.localCheckpointLocation = finalCheckpointLocation
```
BradLarson, line 73: The changes here should be unnecessary. The TF Hub ALBERT snapshot is an archived SavedModel. The CheckpointReader will download that, unarchive it, and recurse through the directories within to find the checkpoint location automatically. I just tried 

```swift
let url = URL(string:
            "https://storage.googleapis.com/tfhub-modules/google/albert_base/1.tar.gz")!
let reader = try! CheckpointReader(checkpointLocation: url, modelName: "ALBERT")
```

locally, using the current CheckpointReader, and it properly loaded the checkpoint and found 30 tensors within. I believe you should be safe in reverting the changes to this file and still having the rest work.

Models/Text/BERT/BERTCheckpointReader.swift
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
BradLarson, line 1: Looks like this file needs to be added to the Models/Text/CMakeLists.txt

Support/Checkpoints/CheckpointReader.swift
leoxzhao: You are right, ALBERT is fine. This change is for loading BERT model hosted in TensorFlow Hub, https://tfhub.dev/tensorflow/bert_en_uncased_L-12_H-768_A-12/1, where weights are stored in `variables/variables.index`, and `variables.data-00000-of-00001`. We don't need it for this repo. I'll revert this change.

Models/Text/BERT/BERTCheckpointReader.swift
leoxzhao: I added this file into CMakeLists.txt. I don't have cmake setup properly to verify if it is adequate though. We need to rely on CI.

leoxzhao: `CMakeLists.txt` was recently updated to remove `RegularizableEmbedding.swift`. I rebased it to fix the conflict.

BradLarson: That looked like it fixed the compile issue, thanks.

Examples/BERT-CoLA/main.swift
```diff
@@ -18,13 +18,41 @@ import ModelSupport
...
+            "https://storage.googleapis.com/bert_models/2018_10_18/uncased_L-12_H-768_A-12.zip")!
+        reader = try! CheckpointReader(checkpointLocation: url, modelName: "BERT")
+    }
+} else {
+    modelName = "BERT"
+    bertPretrained = BERT.PreTrainedModel.bertBase(cased: false, multilingual: false)
+    let url = URL(string:
+        "https://storage.googleapis.com/bert_models/2018_10_18/uncased_L-12_H-768_A-12.zip")!
+    reader = try! CheckpointReader(checkpointLocation: url, modelName: "BERT")
+}
```
xihui-wu, line 50: At first glance, this block of logic should be kept in BERT.swift. Can you refactor ?

leoxzhao: After your change, I see what were you asking.

The reason I lifted checkpoint path out of BERT was that I wanted to try out different checkpoints. As you can see from this page: https://github.com/google-research/bert, there are 24 smaller BERT models in there. Right now, the path and configuration are tightly embedded into the code. By supplying checkpoint URL from outside, I can just change a few parameters to load those different checkpoints without adding new enum items.

In summary, your change is good. Ideally we can come up with a solution to load different checkpoints without changing BERT's code.

Issue #385, opened 2020-03-13
Resolve occasional crash in TransformerDemo


texasmichelle:
TransformerDemo occasionally crashes unexpectedly. Figure out why!

texasmichelle:
I believe this is related to the context size.  Every time `GPT2.generate()` is called, the attention context is mutated ([`states` in `TransformerLM.callAsFunction()`](https://github.com/tensorflow/swift-models/blob/master/Models/Text/GPT2/TransformerLM.swift#L327)). When `states` surpasses 1024 tokens, the app crashes with this error:

```
Fatal error: indices[0,0] = 1024 is not in [0, 1024): file /swift-base/tensorflow-swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 300
```

The context size is defined when loading the pre-trained model from a checkpoint (`n_ctx` in hparams.json). To prevent this source of crashing, the context needs to roll over once it reaches this limit.

texasmichelle:
Anecdotally, this occurs more frequently with the Windows & Mac UIs than its console counterpart 🤔

texasmichelle:
At least one source of these crashes is caused by [`BytePairEncoder.unicodeToBytes`](https://github.com/tensorflow/swift-models/blob/master/Support/Text/BytePairEncoder.swift#L171). This computed property contains a 1:1 mapping between `UnicodeScalar` and `UInt8`, which is incorrect since unicode scalars can be composed of more than 1 byte.

To reproduce:
```
let token = "Å"
```
`token.utf8` is a collection with 2 elements, whereas `token.unicodeScalars` is a collection with 1 element. Decoding this token results in a fatal error:
```
BytePairEncoder.decode(token: token)
```
```
Fatal error: Unexpectedly found nil while unwrapping an Optional value: file /tmp/tmp23rjche4/swift-install/package/.build/checkouts/swift-models/Support/Text/BytePairEncoder.swift, line 290
Current stack trace:
0    libswiftCore.so                    0x00007fe154f96160 swift_reportError + 50
1    libswiftCore.so                    0x00007fe155008c20 _swift_stdlib_reportFatalErrorInFile + 115
2    libswiftCore.so                    0x00007fe154cad50e <unavailable> + 1504526
3    libswiftCore.so                    0x00007fe154cad067 <unavailable> + 1503335
4    libswiftCore.so                    0x00007fe154cacd43 <unavailable> + 1502531
[...]

PR #556, opened 2020-05-20
[GPT2] Capture an error on unicode support bug

texasmichelle:
This is a temporary patch to prevent crashes caused by a bug in `BytePairEncoder` support for multibyte unicode scalars, described in #385

### DIFF
--- a/Models/Text/GPT2/GPT2.swift
+++ b/Models/Text/GPT2/GPT2.swift
@@ -22,7 +22,6 @@ public class GPT2 {
 
     public enum GPT2Error: Error {
         case invalidEncoding(id: Int32)
-        case invalidSeed(seed: Tensor<Int32>)
     }
 
     public var model: TransformerLM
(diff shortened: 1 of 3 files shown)


Models/Text/GPT2/GPT2.swift
```diff
@@ -155,15 +154,13 @@ public class GPT2 {
...
-          throw GPT2Error.invalidSeed(seed: seed)
-        }
+        let id = Int32(seed[0][0])!
         if id == Int32(endOfTextId) {
             // Replace with newline.
             return "\r\n"
         }
         if let token: String = bpe.vocabulary.token(forId: Int(id)) {
-            let decodedToken = BytePairEncoder.decode(token: token)
+            let decodedToken = try BytePairEncoder.decode(token: token)
```
compnerd, line 163: Should this be `try!` or doesn't `decodedToken` become optional?

Support/Text/BytePairEncoder.swift
```diff
@@ -15,6 +15,10 @@
 import Foundation
 
 public struct BytePairEncoder {
+    public enum BPEError: Error {
+        case unsupported
```
compnerd, line 19: I think that `invalidEncoding` is better than `unsupported` ... it is supported, just cannot be encoded in a single byte.

texasmichelle: Changed to `invalidGeneratedToken`, to reflect that it's an intermediate encoded value generated by the bpe.

Models/Text/GPT2/GPT2.swift
texasmichelle: `decode()` does not return an optional and if it fails we just want to propagate the error, so neither `try!` nor `try?` make sense here.

PR #558, opened 2020-05-20
Remove all OldMNIST*

xihui-wu:


BradLarson:
This is great, thanks for the hard work.

PR #560, opened 2020-05-21
[WordSeg] Improvements for X10

sgugger:
Follow-up from #535 with some improvements to make it better on X10.

### DIFF
--- a/Examples/WordSeg/main.swift
+++ b/Examples/WordSeg/main.swift
@@ -73,7 +73,7 @@ print("Starting training...")
 
 for epoch in 1...maxEpochs {
   Context.local.learningPhase = .training
-  var trainingLossSum: Float = 0
+  var trainingLossSum: Tensor<Float> = .zero
   var trainingBatchCount = 0
   for sentence in dataset.training {
     let (loss, gradients) = valueWithGradient(at: model) { model -> Tensor<Float> in
@@ -84,7 +84,7 @@ for epoch in 1...maxEpochs {
       return loss
     }
 
-    trainingLossSum += loss.scalarized()
+    trainingLossSum += loss
     trainingBatchCount += 1
     optimizer.update(&model, along: gradients)
 
@@ -97,7 +97,7 @@ for epoch in 1...maxEpochs {
   }
 
   // Decrease the learning rate if loss is stagnant.
-  let trainingLoss = trainingLossSum / Float(trainingBatchCount)
+  let trainingLoss = trainingLossSum.scalarized() / Float(trainingBatchCount)
   trainingLossHistory.append(trainingLoss)
   reduceLROnPlateau(lossHistory: trainingLossHistory, optimizer: optimizer)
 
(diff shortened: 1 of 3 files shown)


Examples/WordSeg/main.swift
```diff
@@ -73,7 +73,7 @@ print("Starting training...")
 
 for epoch in 1...maxEpochs {
   Context.local.learningPhase = .training
-  var trainingLossSum: Float = 0
+  var trainingLossSum: Tensor<Float> = .zero
```
texasmichelle, line 76: I don't think the changes in this file will help with X10 since there is usually a `LazyTensorBarrier` after `optimizer.update()`.

Models/Text/WordSeg/Lattice.swift
```diff
@@ -54,7 +54,7 @@ public struct Lattice: Differentiable {
         SemiRing(
           logp: logp,
           // TODO(abdulras): this should really use integeral pow
-          logr: logp + Tensor(logf(powf(Float(sentence.count), Float(order)))))
+          logr: logp + Tensor(logf(powf(Float(sentence.count), Float(order))), on: logp.device))
```
texasmichelle, line 57: There are a few more places the device could be specified, like `SNLM`. Essentially anywhere a `Tensor` is initialized.

Examples/WordSeg/main.swift
texasmichelle: I guess this is optional since @asuhan has [this](https://github.com/tensorflow/swift-models/commit/be1e2fbdc9792b6636dd87e3f2d77695413f8b2e) in the works. @asuhan can you get what you have ready to merge so that we can get X10 benchmarks running?

Models/Text/WordSeg/Lattice.swift
texasmichelle: I guess this is optional since @asuhan has [this](https://github.com/tensorflow/swift-models/commit/be1e2fbdc9792b6636dd87e3f2d77695413f8b2e) in the works. @asuhan can you get what you have ready to merge so that we can get X10 benchmarks running?

Examples/WordSeg/main.swift
BradLarson: We've found that the use of `scalarized()` within a training loop can lead to decreased performance (see [the discussion in this PR](https://github.com/tensorflow/swift/pull/462)). One concern that I have for the use of `.zero` is that I think it will be allocated on the default (eager) device. We don't have an XLA device defined here to switch to, but [this](https://github.com/tensorflow/swift-models/commit/be1e2fbdc9792b6636dd87e3f2d77695413f8b2e#diff-0e746268d29de0bb271ecbf56851808aR76) is what Alex did in his variant.

Models/Text/WordSeg/Lattice.swift
asuhan: @texasmichelle I don't think I'll have time for it until the weekly merge is sorted out, unfortunately. I can definitely look into it next week.

Examples/WordSeg/main.swift
texasmichelle: very interesting about decreased perf! Thanks for the link

texasmichelle:
Thanks for the follow-up!!

asuhan:
This is now replaced by https://github.com/tensorflow/swift-models/pull/593 since we reworked portions of the model to use less tensors.

PR #563, opened 2020-05-21
Add PersonLab human pose estimator

joaqo:
Hi, this PR adds PersonLab. Any comments are very welcome!

@BradLarson I improved the benchmarking code, as the old benchmarks I was running were very inaccurate (I wasn't considering that the first inference step is much slower than the following ones). Ran these benchmarks using the newly released 0.9 toolchain, and the results are really really good. Around 250 fps on a 1080, several times faster than on the 0.8 toolchain! I've tried several pose estimators and this is the fastest one I've seen yet, go Swift!

Gave X10 a try, and the backbone's speed remains mostly unchanged, but the decoder becomes much slower for some reason.

There are some things I am not sure how to deal with:
* I have stored the checkpoints in the release page of my fork of this repo, they are only 10MB so maybe I could add them to this repo?
* I added SwiftCV as a dependency, which I can see could be undesirable for swift-models as a whole, as it needs the user to have openCV4 installed in their machines to work. I couldn't think of a way to get around this dependency for the demo though.

Also, for an introduction to how the model works, and an explanation of how some parts of the codebase works please feel free to watch the first part of this [video](https://www.youtube.com/watch?v=WxFPrypPBpU&amp=&feature=youtu.be). Also the README has install/usage instructions.


BradLarson:
Thanks for contributing this, I'll review and provide more specific feedback as I go along. As a first step, it looks like you might need to sign the Contributor License Agreement and notify the bot when done.

Beyond that, my big concern is the addition of the dependency to SwiftCV for the entire swift-models repository. For loading still images, we can use the Image type within swift-models, which encapsulates platform-independent image loading and saving, but I'm not sure what we can use for a video feed, drawing overlays, and display of the results. I may have to do a little digging to see what our options are.

joaqo:
@googlebot I signed it!

joaqo:
Is there anything I can do to help? I realize that having SwiftCV as dependency for drawing the poses is a big pain point, I actually tried to avoid it myself when first developing the model. Maybe GPUImage2 would be better? Maybe avoiding drawing poses altogether?

joaqo:
Great, thanks! Just made the changes you suggested. Let me know if I can help with anything else.

BradLarson:
I've tried to figure out a way to have SwiftCV be compatible with swift-models overall, but as you can see from the Kokoro build failure, adding SwiftCV as a dependency fails to build on any system without OpenCV installed. That would be a significant problem for anyone who needs swift-models as a dependency or who just wants to quickly clone and run examples from the project.

The model itself is great, and I want to have it in here, so my suggestion as an initial implementation (we can build upon this later) would be to focus on just the still image case, and use our existing methods for loading and saving images. The challenge would be to find a way to perform the image annotations within that, but maybe there's something we could do for drawing lines overlaid on the source image.

Reducing it to that first case would allow us to pull in the model itself, which is the important part, and then we would later figure out how to extend it to video and other sources in a way that still allows the project to build across various platforms.

Sorry for the hassle, but we do want swift-models to be easily accessible to a variety of systems with minimal setup involved.

joaqo:
Hi Brad, I completely understand, I would've made the same decision.

The only thing left to resolve I think is how to draw the poses on the still images. Is there any non-iOS Swift library which you recommend for this? I haven't been able to find any, so I suspect I may have to just do it by hand on a pixel by pixel basis, which I am totally up to if it is whats needed.

Let me know what you think and I'll get on it, thanks!

joaqo:
Hi @BradLarson, just to clarify my last post, I will remove the video inference code and will just focus on single image inference, so that we can get rid of SwiftCV as a dependency. First though, I wanted to see if we could agree on a way of handling the drawing of the poses on the images.

My first instinct is to just edit the array storing the image on a pixel by pixel basis to do the drawings. I think if I do this it would make sense to abstract most of this code away from this particular model and maybe place a couple of basic drawing primitives in some general location of this repo, as it could be useful for other models.

This is of course if you don't have any suggestions on other libraries I could use to manage the drawing, or maybe forego drawing altogether.

Thanks!

BradLarson:
To minimize dependencies, and to guarantee this works across platforms, maybe some kind of general Bresenham line drawing that operates on a pixel array (or Tensor) could be an interesting addition. I see some Swift examples of this [here](https://github.com/RockLobster/Bresenham-Swift) and [here](https://github.com/MainasuK/Bresenham-Line) that could act as a basis.

For [SwiftPlot](https://github.com/KarthikRIyer/swiftplot), Karthik leveraged AGG to do the drawing across platforms, but that's a challenge to get incorporated. I think there also have been some recent compilation issues with that and Colab, if I remember correctly. Nothing that couldn't be solved, but if we just need lines to be overlaid on images it might be handy to have our own means of annotating that.

An interesting possibility with doing the drawing on Tensors is the extension of this to be used in producing segmentation maps or in data augmentation be selectively blanking out parts of the image.

joaqo:
Hey @BradLarson! Finally got some time to make the changes needed. Removed SwiftCV and added line drawing.

PD: Really loved the design of the new standard training loop api, and the proposed api for model checkpointing shown on last week's meeting. The super powerful callbacks in the training loop, and using a function to map between each tensor's checkpoint name and its model name when loading checkpoints look particularly neat. Really happy with the decisions being made.

BradLarson:
To fix the merge conflicts, I believe you'll just need to remove Package.resolved within your branch and do a quick edit on Package.swift. I tried to do this on my end, but it looks like I don't have push access to this PR.

I had one small comment about being able to automatically download checkpoints, but if the conflict could be resolved and that little nicety added, I think we're good to go for pulling this in. Thanks for the line drawing implementation, that makes this simple to include.

PersonLab/main.swift
```diff
@@ -0,0 +1,67 @@
...
+struct Inference: ParsableCommand {
+  static var configuration = CommandConfiguration(
+    commandName: "personlab",
+    abstract: """
+      Runs human pose estimation on a local image file.
+      """
+  )
+
+  @Argument(help: "Path to checkpoint directory")
+  var checkpointPath: String
```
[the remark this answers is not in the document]
joaqo: Done.

Short note, I think I found a small bug in CheckpointReader when downloading checkpoints. I was looking for examples of CheckpointReader downloading checkpoints in swift-models and I tried this particular mobilenet [example](https://github.com/tensorflow/swift-models/blob/a6ad2bb523c8b79dcabfc564cd6d4ab895732879/Tests/CheckpointTests/CheckpointReaderTests.swift#L40) to see if it would trigger a download. It downloaded the file just fine, although obviously the checkpoint didn't correctly load into PersonLab as its not the correct checkpoint. When I tried using the url to the correct checkpoint, it seems that CheckpointReader loaded the incorrect cached initial checkpoint from my /tmp folder instead of re-downloading the new correct url. This got fixed after I deleted the initial checkpoint from /tmp/Personlab/ and ran PersonLab again.

joaqo:
Added the default checkpoint downloading to the checkpoint loader, and left a comment above regarding a possible bug.

Regarding the merge conflicts, I am not sure how to give you push access to the PR, so I just rebased to master, fixed some merge conflicts that popped up with CheckpointReader and force pushed, I hope thats fine.

BradLarson:
Thanks for getting everything ready to go. I tried it out locally, and it works great.

Why we've been holding off on pulling this in is unrelated to the code, which should be good to go. While testing this, I noticed that running a release build of this triggered a compiler crash. It's the same one as seen [in PR #1022 over on swift-apis](https://github.com/tensorflow/swift-apis/pull/1022), and we're currently working on a fix for that.

To make sure we don't cause problems with our CI runs, I'll hold off on pulling this in until we have that crasher fixed. There's nothing more to do on your end, this is on us to fix. Sorry for the further delays, we hopefully should have that resolved soon.

joaqo:
Got it, thank you for keeping me posted. Good luck with that fix!

PersonLab/Backbone.swift
```diff
@@ -0,0 +1,203 @@
...
+
+public struct DepthwiseSeparableConvBlock: Layer {
+  var dConv: DepthwiseConv2D<Float>
+  var conv: Conv2D<Float>
+  var depthWiseFilter: Tensor<Float>
+  var depthWiseBias: Tensor<Float>
+  var pointWiseFilter: Tensor<Float>
+  var pointWiseBias: Tensor<Float>
+
+  @noDerivative let strides: (Int, Int)
```
BradLarson, line 26:
```suggestion
```
If you delete these lines, and the ones below, that fixes the compiler crash. These filters are already being optimized within the two convolution layers below, so retaining them at the top level is unnecessary and may be causing the compiler some confusion here.

PersonLab/Backbone.swift
```diff
@@ -0,0 +1,203 @@
...
+    pointWiseFilter: Tensor<Float>,
+    pointWiseBias: Tensor<Float>,
+    strides: (Int, Int)
+  ) {
+    self.depthWiseFilter = depthWiseFilter
+    self.depthWiseBias = depthWiseBias
+    self.pointWiseFilter = pointWiseFilter
+    self.pointWiseBias = pointWiseBias
+    self.strides = strides
+
```
BradLarson, line 40:
```suggestion
```

BradLarson:
Sorry for the long delay as we've looked into this. The compiler crasher has proven to be difficult to address at the source, but I've created two suggestions above that work around the crash in this case. The convolution filters were being held on to in two places, when they really should have just been passed on to the internal convolution within the block.

This code can easily be removed without impacting the function of the network (and it simplifies the model). With these changes in place, optimized builds now work for this model and I'd finally be able to merge this in. I would have done this directly, but I don't have push access to this PR.

joaqo:
Hi Brad, thanks for the suggestion. Not sure why I had decided to define those tensors there in the first place, but your solution looks much better. Tested and pushed.

BradLarson:
Sorry again for the extremely long time it took to resolve this. Everything works great now, and it'll be awesome to have this reference model to work from. At some point, we might even get the components in place to add back the video functionality.

I appreciate the work you put into this and your patience with the process.

joaqo:
Thanks! No worries with the delays, S4TF is a very ambitious project with many moving parts and I completely understand having to put some things on hold while others are getting fixed.

At the moment I am particularly exited with the new Jetson support, and I plan on trying new things with pose estimation on there which may end up as PRs here.

PR #566, opened 2020-05-27
[WordSeg] Add inline documentation

texasmichelle:
Primarily adds inline documentation to `WordSeg` components. Also:

* Moves `CharacterErrors` into CharacterSequence.swift, where it is primarily used.
* Conforms naming to Swift conventions
* Renames `WordSegRecord` -> `Phrase`
* In `WordSegDataset`
  * Combines several functions
  * Removes a raw loop
  * Replaces optional members with empty arrays
  * Renames `DownloadDetails` -> `DownloadableArchive` and `convertDataset()` -> `numericalizeDataset()`
* Adds a test for running with training data only


Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -73,6 +102,10 @@ public struct WordSegDataset {
     return Alphabet(sorted, eos: eos, eow: eow, pad: pad)
   }
 
+  /// Creates a collection of records to be used with the WordSeg model.
```
xihui-wu, line 105: Same above start with "Returns"

Support/Text/WordSeg/Lexicon.swift
```diff
@@ -14,22 +14,39 @@
...
+/// In Python implementations, this is sometimes called the string vocabulary
+/// (in contrast to the character vocabulary or `Alphabet`, which maps
+/// characters to integers).
 public struct Lexicon {
+  /// A type whose instances represent a sequence of characters.
   public typealias Element = CharacterSequence
 
+  /// A one-to-one mapping between a sequence of characters and unique
+  /// integers.
   // TODO(marcrasi): if the value is not used to construct Tensor, switch to Int
```
xihui-wu, line 28: new line and ///

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Done

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,48 @@
...
+  ///
+  /// - Parameter data: text in UTF8 format.
   private static func load(data: Data) throws -> [String] {
     guard let contents: String = String(data: data, encoding: .utf8) else {
       throw CharacterErrors.nonUtf8Data
     }
     return load(contents: contents)
   }
 
+  /// Separates `contents` into a collection of strings by newlines, trimming
```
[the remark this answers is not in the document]
texasmichelle: Done

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,48 @@
...
     var testingFilePath = "br/br-text/te.txt"
+    /// The path to the training source.
     var trainingFilePath = "br/br-text/tr.txt"
+    /// The path to the validation source.
     var validationFilePath = "br/br-text/va.txt"
   }
 
+  /// Returns a list of records parsed from `data` in UTF8.
+  ///
+  /// - Parameter data: text in UTF8 format.
```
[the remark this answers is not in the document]
texasmichelle: Done, as well as other throwing functions

Support/Text/WordSeg/Lexicon.swift
texasmichelle: Are TODOs considered documentation? I didn't think we wanted to surface those.

xihui-wu: Yes...? 😄 It's listed as keyword of bullet item in "Write a documentation comment" https://swift.org/documentation/api-design-guidelines/

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
 import Foundation
 import ModelSupport
 
+/// A collection of raw and processed text used for training and validation
```
dabrahams, line 18: Does `Lexicon` actually conform to `Collection`?  If not, it's probably misleading to say it's “a collection.”  If so, it would be clearer to s/collection/`Collection`/

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
+/// of word segmentation models.
 public struct WordSegDataset {
```
dabrahams, line 20: There should be a blank line before each doc comment, so the doc comment is visually grouped with the thing it documents.

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
+  /// A collection of text used for training.
```
dabrahams, line 21:
```suggestion
  /// The text used for training.
```

Saying “a collection” is redunant here; we know it's an array by looking at the declaration.

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
   public let training: [WordSegRecord]
```
dabrahams, line 22:
```suggestion
  public let trainingText: [WordSegRecord]
```

or 

```suggestion
  public let trainingData: [WordSegRecord]
```

I realize this commit is about the comments, but one of the main functions of writing comments is that it informs naming and design.  In this case, it seems like saying what it is in the comment tells you what it should be called.  Calling this property "training" implies it is a representation of the process of training.

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
+  /// A collection of text used for testing.
   public private(set) var testing: [WordSegRecord]?
+  /// A collection of text used for validation.
   public private(set) var validation: [WordSegRecord]?
+  /// The set of characters found in all included texts.
```
dabrahams, line 27: This doc seems to imply that every member of the set is present in every included text.  That doesn't seem likely to me.

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
   public let alphabet: Alphabet
 
+  /// Details used for downloading source data.
   private struct DownloadDetails {
```
dabrahams, line 31: This looks like a case where a relatively weak name drove the doc comment.  I suggest trying to push the doc comment to really describe what's being represented here, and then using that to reconsider naming.

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Called this a dataset instead!

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Applied this change

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Done

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Renamed `WordSegRecord` to `Phrase` and `training` to `trainingPhrases`.

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Changed to `/// The union of all characters in the included dataset.`

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Renamed to `ReferenceArchive` and combined URL components into a single `location`.

Support/Text/WordSeg/CharacterSequence.swift
```diff
@@ -37,33 +47,65 @@ public struct CharacterSequence: Hashable {
...
   public var count: Int { return characters.count }
+
+  /// The last character in the sequence, if `characters` is not empty.
+  ///
+  /// - Note: This is usually the end marker.
   public var last: Int32? { return characters.last }
+
+  /// TODO: what's happening here?
   public var tensor: Tensor<Int32> {
     Tensor<Int32>([self.eos] + characters[0..<characters.count - 1])
```
texasmichelle, line 84: @compnerd Can you help me understand why the tensor is constructed this way, with `eos` explicitly added to a slice of the character array? Why is the last element not included? Related, why does one `init` append the end of sequence marker and others do not? Does that not cause inconsistencies with this property (is it assuming there is an eos marker or not)?

compnerd: Hmm, I don’t remember the full details, but this construction was definitely odd in the reference that was used, and this implementation was trying to preserve the original behavior since I couldn’t puzzle out why it was doing it this way.

Datasets/WordSeg/Phrase.swift
```diff
@@ -14,10 +14,17 @@
...
 
-public struct WordSegRecord {
+/// A sequence of text for use in word segmentation.
+public struct Phrase {
+
+  /// A raw, unprocessed sequence of text.
   public let plainText: String
+
+  /// A sequence of text in numeric form, derived from `plainText`.
   public let numericalizedText: CharacterSequence
```
dabrahams, line 24: I remember wanting to mention this before, but if “Character” doesn't mean what in Swift is called `Character`, we should look for other names, e.g. “glyph.”

Datasets/WordSeg/Phrase.swift
```diff
@@ -14,10 +14,17 @@
...
 
+  /// Creates an instance containing both raw (`plainText`) and processed
+  /// (`numericalizedText`) forms of a sequence of text.
```
dabrahams, line 27: I've always thought “numericalized” read terribly awkwardizedly.  I get the impression this is a term of art, but we should discuss whether it's the best choice.

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
+    /// The path to the training source.
     var trainingFilePath = "br/br-text/tr.txt"
+
+    /// The path to the validation source.
     var validationFilePath = "br/br-text/va.txt"
   }
 
-  private static func load(data: Data) throws -> [String] {
+  /// Returns the text of all phrases parsed from `data` in UTF8.
+  private static func load(data: Data) -> [String] {
```
dabrahams, line 55:
```swift
String(data: data, encoding: .utf8).map(load) ?? []
```

Would be much simpler. (GitHub won't let me use suggestions on this one for some reason).

This raises the question for me whether we really want to bail out when there are any utf8 encoding errors, which is what that initalizer does.  Typically, you want to repair malformed UTF8 and use the data you can, which would be done with: https://developer.apple.com/documentation/swift/string/2907004-init

It is [appalling](https://bugs.swift.org/browse/SR-12935) that it is not documented what happens with malformed UTF8 in that case!

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
     guard let contents: String = String(data: data, encoding: .utf8) else {
-      throw CharacterErrors.nonUtf8Data
+      return []
     }
     return load(contents: contents)
   }
 
+  /// Returns the text of all phrases from `contents`.
```
dabrahams, line 62: Here and above, I think we need to say something more about the expected format of contents.  How is it divided into phrases?

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
   private static func load(contents: String) -> [String] {
```
dabrahams, line 63: It's probably much more efficient to use swift-native string processing than to go through `Foundation` methods, and the best approach probably returns a `[Substring]` to avoid lots of copying and allocation.  You can start with contents.split(separator: "\n") and then consider whether more trimming needs to be done.

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -48,6 +71,11 @@ public struct WordSegDataset {
     return strings
   }
 
+  /// Returns the union of all characters in `training` and `otherSequences`.
+  ///
+  /// - Parameter eos: text to be used as the end of sequence marker.
+  /// - Parameter eow: text to be used as the end of word marker.
+  /// - Parameter pad: text to be used as the padding marker.
```
dabrahams, line 78: See notes above about "character"

```suggestion
  /// Returns the union of all characters in `training` and `otherSequences`.
  ///
  /// - Parameter eos: the end of sequence marker.
  /// - Parameter eow:the end of word marker.
  /// - Parameter pad: the padding marker.
```

We should talk about what these markers really mean; I think we can do better in the comment.

`otherSequences` needs to be described.  Generally the use of current Swift variadics, especially in a `private` method is questionable; we should talk about that. 

The implementation below looks like a terribly inefficient algorithm also, let's talk about that too.

Support/Text/WordSeg/Lexicon.swift
dabrahams: Cart before horse: A TODO should only be considered documentation *if* you think surfacing it to users is desirable.  I can't see the full context where I am writing this comment, but I am guessing this would not be useful to users.

Datasets/WordSeg/WordSegDataset.swift
dabrahams: Again, [No raw loops](https://developer.apple.com/videos/play/wwdc2018/223/)

Datasets/WordSeg/WordSegDataset.swift
xihui-wu: I was also looking for the repairing thing last week and expected init(decoding:as:) would do it but apparently not. Just FYI found this method does it https://developer.apple.com/documentation/swift/string/1641442-decodecstring which looks complicated.

dabrahams: No need to use that one; the one I referenced takes a `Collection` of the encoding's `CodeUnit`s, which `Data` already is.

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
+  public private(set) var validationPhrases: [Phrase]?
+
+  /// The union of all characters in the included dataset.
   public let alphabet: Alphabet
 
-  private struct DownloadDetails {
-    var archiveLocation = URL(string: "https://s3.eu-west-2.amazonaws.com/k-kawakami")!
-    var archiveFileName = "seg"
-    var archiveExtension = "zip"
+  /// A pointer to source data.
```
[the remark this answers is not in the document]
texasmichelle: Renamed to `DownloadableArchive` and changed members to `let`s.

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
+/// The reference archive was published in the paper "Learning to Discover,
+/// Ground, and Use Words with Segmental Neural Language Models" by Kazuya
+/// Kawakami, Chris Dyer, and Phil Blunsom:
+/// https://www.aclweb.org/anthology/P19-1645.pdf.
 public struct WordSegDataset {
-  public let training: [WordSegRecord]
-  public private(set) var testing: [WordSegRecord]?
-  public private(set) var validation: [WordSegRecord]?
+
+  /// The text used for training.
```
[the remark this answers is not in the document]
texasmichelle: Simplified these comments to remove implied "text"

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,28 +15,51 @@
...
+  public let trainingPhrases: [Phrase]
+
+  /// The text used for testing.
+  public private(set) var testingPhrases: [Phrase]?
+
+  /// The text used for validation.
+  public private(set) var validationPhrases: [Phrase]?
+
+  /// The union of all characters in the included dataset.
```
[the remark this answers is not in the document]
texasmichelle: Since `Alphabet` uses a `BijectiveDictionary`, the keys are represented as a set.

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Removed the `Foundation` string processing, raw loop, and collapsed the two functions into one.

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Morphed this comment into:
```
/// Returns phrases parsed from `data` in UTF8, separated by newlines.
```

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Updated to use `String(decoding:as:)`

Datasets/WordSeg/WordSegDataset.swift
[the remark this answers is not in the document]
dabrahams:
```suggestion
  /// A mapping between characters used in the dataset and densly-packed integers
```
?

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,54 +15,66 @@
...
+
+    /// The path to the validation data within the unpacked archive.
+    let validationFilePath = "br/br-text/va.txt"
+  }
+
+  /// Returns phrases parsed from `data` in UTF8, separated by newlines.
+  private static func load(data: Data) -> [String] {
+    let contents = String(decoding: data, as: Unicode.UTF8.self)
+    let splitContents = contents.split(separator: "\n", omittingEmptySubsequences: true)
+    return splitContents.map { String($0) }
```
dabrahams, line 59: Should return `spitContents` directly, which is of type `[Substring]`

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,54 +15,66 @@
...
   ) -> Alphabet {
     var letters: Set<Character> = []
 
-    for dataset in otherSequences + [training] {
-      guard let dataset = dataset else { continue }
-      for sentence in dataset {
-        for character in sentence {
-          if !character.isWhitespace { letters.insert(character) }
-        }
+    for phrase in phrases {
```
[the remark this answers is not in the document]
dabrahams: `letters = Set(phrases.joined().lazy.filter { !$0.isWhitepsace })`

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -73,52 +85,69 @@ public struct WordSegDataset {
...
+    return phrases
   }
 
+  /// Creates an instance containing phrases from the reference archive.
+  ///
+  /// - Throws: an error in the Cocoa domain, if the default training file
+  ///   cannot be read.
   public init() throws {
-    let downloadDetails = DownloadDetails()
+    let downloadableArchive = DownloadableArchive()
```
dabrahams, line 118: Don't name things after their types, generally speaking.  What role does this thing play in this code?  That's how you name it.

```
let person: Person = ...
```

vs.

```
let hero: Person = ...
```

texasmichelle: Renamed to `source`

Datasets/WordSeg/WordSegDataset.swift
texasmichelle: Done

Datasets/WordSeg/WordSegDataset.swift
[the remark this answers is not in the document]
texasmichelle: This is much better!

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -73,52 +85,69 @@ public struct WordSegDataset {
...
   {
-    return try dataset.map {
-      let trimmed = $0.components(separatedBy: .whitespaces).joined()
-      return try WordSegRecord(
-        plainText: $0,
-        numericalizedText: CharacterSequence(
-          alphabet: alphabet, appendingEoSTo: trimmed))
+    var phrases = [Phrase]()
+
+    for data in dataset {
```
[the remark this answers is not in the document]
texasmichelle: How could I use `compactMap` here if I need to include the original text? If I want to base my inclusion on whether `CharacterSequence()` results in nil, how can I include the original text in composing `Phrase`?

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -73,52 +101,69 @@ public struct WordSegDataset {
...
   }
 
-  private static func convertDataset(_ dataset: [String], alphabet: Alphabet) throws
-    -> [WordSegRecord]
+  /// Returns phrases from `dataset`, using `alphabet`, to be used with the
+  /// WordSeg model.
+  ///
+  /// - Note: Omits any part of the dataset that cannot be converted to
+  ///   `CharacterSequence`.
+  private static func convertDataset(_ dataset: [String], alphabet: Alphabet)
```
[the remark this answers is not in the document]
texasmichelle: Renamed to `numericalizeDataset`

Datasets/WordSeg/WordSegDataset.swift
```diff
@@ -15,110 +15,129 @@
...
+  /// - Throws: an error in the Cocoa domain, if the default training file
+  ///   cannot be read.
   public init() throws {
-    let downloadDetails = DownloadDetails()
+    let source = DownloadableArchive()
     let localStorageDirectory: URL = DatasetUtilities.defaultDirectory
       .appendingPathComponent("WordSeg", isDirectory: true)
 
-    WordSegDataset.downloadIfNotPresent(to: localStorageDirectory, downloadDetails: downloadDetails)
+    WordSegDataset.downloadIfNotPresent(
```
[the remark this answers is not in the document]
texasmichelle: I can never remember the difference between `self` and `Self`. Great suggestion :)

Datasets/WordSeg/WordSegDataset.swift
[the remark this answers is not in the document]
texasmichelle: This feels like a symptom of awkward design. I believe rethinking `CharacterSequence` will make these acrobatics unnecessary.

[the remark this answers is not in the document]
dabrahams:
```swift
    // untested of course
    return dataset.compactMap { data in
      let trimmed = data.split(separator: " ", omittingEmptySubsequences: true).joined()
      let numericalizedText = try? CharacterSequence(alphabet: alphabet, appendingEoSTo: trimmed)
      return numericalizedText.map { Phrase(plainText: String(data), numericalizedText: $0 }
    }
```
@texasmichelle

PR #571, opened 2020-05-28
Return replacement character instead of throwing error if bpe decode fails

xihui-wu:
See details in https://github.com/tensorflow/swift-models/issues/385


Support/Text/BytePairEncoder.swift
```diff
@@ -284,16 +280,15 @@ extension BytePairEncoder {
...
         var buffer = [UInt8]()
 
         for scalar in token.unicodeScalars {
             buffer.append(BytePairEncoder.unicodeToBytes[scalar]!)
         }
 
-        // TODO(#385): Support unicode strings with multibyte scalars.
         guard let decodedToken = String(bytes: buffer, encoding: .utf8) else {
-            throw BPEError.invalidGeneratedToken
+            return String("\u{FFFD}")
```
texasmichelle, line 291: How did you decide on this particular value?

dabrahams: It's "a thing": https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character

Issue #568, opened 2020-05-27
Flaky TextUnsupervisedTests


xihui-wu:
TextUnsupervisedTests.testCreateWikiText2WithBpe fails occasionally with: 
XCTAssertEqual failed: ("64") is not equal to ("12") 

https://source.cloud.google.com/results/invocations/e9a0dadd-a808-457d-a2f5-850fa9690960/targets/s4tf%2Fmodels%2Fubuntu%2Fmanual/log

I believe that's because it fails in loading the checkpoint and hits https://github.com/tensorflow/swift-models/blob/87552ffa194d257a7a98c5e5d970a343889bd3cb/Models/Text/GPT2/GPT2.swift#L94 while initializing GPT2. Then it creates an empty bpe with which encoding produces more data examples.

PR #573, opened 2020-05-29
Throw error if GPT2 checkpoint fails to load

xihui-wu:
This is the fix for https://github.com/tensorflow/swift-models/issues/568.


Tests/TextTests/GPT2Tests.swift
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
BradLarson, line 1: Looks like you might need to add GPT2Tests.swift to the CMakeLists and remove Inference.swift.

Issue #104, opened 2019-04-18
Inside `FeedForward`, input goes through `dense1`, `dropout`, `dense2`. Is it intentional?


leoxzhao:
```swift
    func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
        return input.sequenced(in: context, through: dense1, dropout, dense2)
    }
```
https://github.com/tensorflow/swift-models/blob/412a822a5c095c56504cae9681bd8ae1f8651edc/Transformer/Model.swift#L35

I did not find reference code in Python with dropout. I had the impression that dropout normally is added to the exit as the last operation before produces the final output. In this case, it would be dense1 -> dense2 -> dropout. This looks like a typo. But I don't possess the knowledge to judge. So I open an issue.

jekbradbury:
Good catch!

The original Transformer performs dropout after the first dense layer in each feedforward block (see e.g. T2T code [here](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/layers/common_layers.py#L1282-L1300)), but BERT performs it after the last dense layer in each feedforward block (see [code](https://github.com/google-research/bert/blob/master/modeling.py#L866-L879)).

This particular codebase is intended to replicate the architecture used in GPT-2, though, where we don't actually know which of these is the case since the training code isn't public.

leoxzhao:
I re-read GPT-2 today. Compared with Python implementing of GPT-2, our EncoderLayer contains additional three dropout layers, `selfAttentionDropout`, `feedForwardDropout` and one inside of `FeedForward`.  Oddly, GPT-2 does not contain any dropout layers. I guess strategic placement of dropout layers one of the things that OpenAI is holding back. I am going to close this issue for now. We can revisit once we have more information or more development in this field.

leoxzhao:
@xihui-wu I noticed that you recently did a fix for GPT2 training portion. If you are working on training or fine-tuning related aspect of it, this issue contains a conversation we had in the past about dropout which might be helpful.

HuggingFace's implementation also contains a dropout for embedding, https://github.com/huggingface/transformers/blob/2c1ebb8b507c19c75af5084b6c73e0b003c9eda6/src/transformers/modeling_gpt2.py#L483.

xihui-wu:
@jekbradbury  >>The original Transformer performs dropout after the first dense layer in each feedforward block.    -- Looks like the link is outdated. Can you paste again?

[...]

PR #574, opened 2020-05-29
Make GPT2 Dropout aligned with huggingface

xihui-wu:
This PR resolves issue https://github.com/tensorflow/swift-models/issues/104.

hugging face model for reference: https://github.com/huggingface/transformers/blob/2c1ebb8b507c19c75af5084b6c73e0b003c9eda6/src/transformers/modeling_gpt2.py

and dropout possibility settings:
https://github.com/huggingface/transformers/blob/2c1ebb8b507c19c75af5084b6c73e0b003c9eda6/src/transformers/configuration_gpt2.py#L122

### DIFF
--- a/Models/Text/GPT2/PythonCheckpointReader.swift
+++ b/Models/Text/GPT2/PythonCheckpointReader.swift
@@ -128,7 +128,7 @@ extension TransformerLM: InitializableFromPythonCheckpoint {
         positionalEmbeddings = reader.readTensor(
             name: scope + "/wpe",
             scalarType: Float.self)
-        embeddingDropout = Dropout<Float>(probability: 0.1)
+        embeddingDropout = Dropout(probability: 0.1)
         layers = (0..<config.layerCount).map { i in
             EncoderLayer(reader: reader, config: config, scope: scope + "/h\(i)")
         }
--- a/Models/Text/GPT2/PythonCheckpointReader.swift
+++ b/Models/Text/GPT2/PythonCheckpointReader.swift
@@ -90,7 +90,7 @@ extension MultiHeadAttentionGPT2: InitializableFromPythonCheckpoint {
         attention = Attention(
             size: config.embeddingSize / config.headCount,
             causal: true,
-            dropProbability: 0.2)
+            dropProbability: 0.1)
         wqkv = TimeDistributed(
             Dense<Float>(reader: reader, config: config, scope: scope + "/c_attn"))
         wo = TimeDistributed(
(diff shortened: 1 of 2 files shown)


Models/Text/GPT2/PythonCheckpointReader.swift
```diff
@@ -129,6 +128,7 @@ extension TransformerLM: InitializableFromPythonCheckpoint {
         positionalEmbeddings = reader.readTensor(
             name: scope + "/wpe",
             scalarType: Float.self)
+        embeddingDropout = Dropout<Float>(probability: 0.1)
```
texasmichelle, line 131: The type can be inferred here, no need to include it.

Models/Text/GPT2/TransformerLM.swift
```diff
@@ -319,6 +320,7 @@ public struct TransformerLM: Module {
         let positionsTensor = Tensor<Int32>(shape: [1, tokens.shape[1]], scalars: positions)
         var h = embedding(tokens)
         h = h + positionalEmbeddings.gathering(atIndices: positionsTensor)
+        h = embeddingDropout(h)
```
texasmichelle, line 323: This also belongs in the differentiable `callAsFunction()` below.

Models/Text/GPT2/TransformerLM.swift
```diff
@@ -298,15 +296,18 @@ public struct EncoderLayer: Layer {
 public struct TransformerLM: Module {
     var embedding: Embedding<Float>
     var positionalEmbeddings: Tensor<Float>
+    var embeddingDropout: Dropout<Float> 
     var layers: [EncoderLayer]
     var norm: LayerNorm<Float>
 
     public init(
         embedding: Embedding<Float>, positionalEmbeddings: Tensor<Float>,
+        dropProbability: Double,
```
texasmichelle, line 305: Why `Double` instead of `Float`?

Models/Text/GPT2/TransformerLM.swift
```diff
@@ -298,15 +296,18 @@ public struct EncoderLayer: Layer {
...
         layers: [EncoderLayer], norm: LayerNorm<Float>
     ) {
         self.embedding = embedding
         self.positionalEmbeddings = positionalEmbeddings
+        self.embeddingDropout = Dropout<Float>(probability: dropProbability)
```
texasmichelle, line 310: The type can be inferred here, too.

Models/Text/GPT2/TransformerLM.swift
xihui-wu: Ah! Thanks for catching this.

Models/Text/GPT2/TransformerLM.swift
xihui-wu: Dropout layer requires so: https://github.com/tensorflow/swift-apis/blob/5bc6e872f340095139d8b451281ec6aba182dec4/Sources/TensorFlow/Layers/Dropout.swift#L59

Issue #514, opened 2020-05-13
Convert image classification and generative examples to use X10 by default


BradLarson:
With the release of the 0.9 toolchain containing X10, we now have a higher-performance alternative to the default eager execution mode. In the interest of having swift-models demonstrate best practices, we should have our image classification and generative models use the highest performance pathway available. For image classification and image generative models, that's pretty much always X10.

Therefore, we should make the (small) modifications needed for the following models to use X10 by default:

- [ ] VGG-Imagewoof
- [ ] Custom-CIFAR10
- [ ] ResNet-CIFAR10
- [ ] LeNet-MNIST
- [ ] MobileNetV1-Imagenette
- [ ] MobileNetV2-Imagenette
- [ ] GAN
- [ ] CycleGan
- [ ] DCGAN
- [ ] FastStyleTransferDemo
- [ ] pix2pix

This is most likely a follow-on issue to issue #511, because the conversion of training loops to Epochs will conflict with updates to the training loops here. We also may need to check that X10's CPU compile times and execution speeds are acceptable for macOS.

Andrew-Pynch:
@BradLarson Hello, I am the guy who wanted to help with this from this morning's design meeting. I am still very new to this codebase. Where do these models live so I can start migrating them? :smile:

BradLarson:
@Andrew-Pynch - Great to hear about the interest. I've added links to each entry to describe where it lives in the repository. The only caution I'd have is that you'll want to coordinate with @xihui-wu who's working on issue #511, and not migrate any models she hasn't yet updated there. Any models not on that list, or ones that have already been updated to Epochs, should be free to update here.

Andrew-Pynch:
@xihui-wu I am going to use my weekend to work on these migrations. I was hoping I would have some time this week but that is not the case :-(. Is this a time-sensitive task? I don't have a gauge for how long it will take to do this but I will get as many done as I can this weekend :-)

xihui-wu:
Take your time @Andrew-Pynch #511 is up to date so you're good to work on anyone checked there.:)

BradLarson:
I'll note that the following models: 

- ResNet-CIFAR10
- LeNet-MNIST
- VGG-Imagewoof
- MobileNetV1-Imagenette
- MobileNetV2-Imagenette

will automatically be updated by PR #586, so we can hold off on this until that lands. The others could potentially be converted as a second stage using that training loop once that is available to the project.

PR #578, opened 2020-06-03
Convert LeNet-MNIST example to X10

Andr0id100:
First of many PRs for #514.

Been busy with a few things lately hence the hiatus. Played around with X10 the last couple of days and I am really exited about the future of the S4TF project. Great work

### DIFF
--- a/Examples/LeNet-MNIST/main.swift
+++ b/Examples/LeNet-MNIST/main.swift
@@ -17,7 +17,14 @@ import Datasets
 
 let epochCount = 12
 let batchSize = 128
-let device = Device.defaultXLA
+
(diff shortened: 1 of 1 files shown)


BradLarson:
As a quick update: I checked this out on Linux and it works great there, however, we're running into some issues with X10 on macOS that we're looking into now. Specifically, this example segfaults under X10 on macOS with both our 0.9 release toolchain and the current nightlies. It's not a problem with what you have here, but something internal.

Before we merge this in, I want to see if we can fix this. If not, we may need to add selective compilation checks to use the TF eager device by default in this example on macOS so that it still works out of the box. I'll let you know what we find.

rahulbhalley:
Yup, I also experienced segmentation fault when executing on X10 on my macOS earlier.

PR #584, opened 2020-06-04
Introduce new layer initialization APIs with automatic shape computation

shadaj:
to be addressed:
- [x] `AutoBatchNorm` is tied to 2 dimensions + channel inputs, we should handle inputs of any dimension (an `HList` equivalent (Automatic Requirement Satisfaction?) could help here)
- [ ] throw an error if the same key is used twice in the model and offer some alternative for attaching keys to layers in reused chunks

Model Porting Progress:
- [x] LeNet ([diff](https://github.com/shadaj/swift-models/compare/layer-init-prototyping...shadaj:layer-init-porting/lenet))
- [x] VGG ([diff](https://github.com/shadaj/swift-models/compare/layer-init-prototyping...shadaj:layer-init-porting/vgg))
- [x] ResNet ([diff](https://github.com/shadaj/swift-models/compare/layer-init-prototyping...shadaj:layer-init-porting/resnet))
- [ ] WordSeg
- [ ] BERT
- [ ] GPT-2

### DIFF
--- a/Models/LayerInit/AutoBatchNorm.swift
+++ b/Models/LayerInit/AutoBatchNorm.swift
@@ -1,0 +1,33 @@
+import TensorFlow
+
+public struct AutoBatchNorm<Shape, Scalar>: AutoLayer where Scalar: TensorFlowFloatingPoint {
+    let axis: Int
+    let momentum: Scalar
+    let epsilon: Scalar
+
+    public typealias InstanceType = BatchNorm<Scalar>
+    public typealias InputShape = Shape
+    public typealias OutputShape = Shape
+
+    public init(
+        axis: Int = -1,
+        momentum: Scalar = 0.99,
+        epsilon: Scalar = 0.001
+    ) {
+        self.axis = axis
+        self.momentum = momentum
+        self.epsilon = epsilon
+    }
+
+    public func buildModelWithOutputShape(inputShape: Shape) -> (InstanceType, Shape) {
+        let inputShapeArray: [Int]
+        if let inputShapeTuple = inputShape as? (Int, Int, Int) {
+            inputShapeArray = [inputShapeTuple.0, inputShapeTuple.1, inputShapeTuple.2]
+        } else {
+            fatalError("Could not extract out elements of shape")
+        }
+
+        let featureCount = inputShapeArray[(inputShapeArray.count + axis) % inputShapeArray.count]
+        return (BatchNorm<Scalar>(featureCount: featureCount, axis: axis, momentum: momentum, epsilon: epsilon), inputShape)
+    }
+}
--- a/Models/LayerInit/AutoBatchNorm.swift
+++ b/Models/LayerInit/AutoBatchNorm.swift
@@ -20,12 +20,7 @@ public struct AutoBatchNorm<Shape, Scalar>: AutoLayer where Scalar: TensorFlowFl
     }
 
     public func buildModelWithOutputShape<Prefix>(inputShape: Shape, keyPathSoFar: KeyPath<Prefix, InstanceType>, keyDict: inout [AnyAutoLayerKey: Any]) -> (InstanceType, Shape) {
-        let inputShapeArray: [Int]
-        if let inputShapeTuple = inputShape as? (Int, Int, Int) {
-            inputShapeArray = [inputShapeTuple.0, inputShapeTuple.1, inputShapeTuple.2]
-        } else {
-            fatalError("Could not extract out elements of shape")
-        }
+        let inputShapeArray: [Int] = intTupleToArray(tuple: inputShape)
 
         let featureCount = inputShapeArray[(inputShapeArray.count + axis) % inputShapeArray.count]
         return (BatchNorm<Scalar>(featureCount: featureCount, axis: axis, momentum: momentum, epsilon: epsilon), inputShape)
--- a/Models/LayerInit/AutoBatchNorm.swift
+++ b/Models/LayerInit/AutoBatchNorm.swift
@@ -19,7 +19,7 @@ public struct AutoBatchNorm<Shape, Scalar>: AutoLayer where Scalar: TensorFlowFl
         self.epsilon = epsilon
     }
 
-    public func buildModelWithOutputShape(inputShape: Shape) -> (InstanceType, Shape) {
+    public func buildModelWithOutputShape<Prefix>(inputShape: Shape, keyPathSoFar: KeyPath<Prefix, InstanceType>, keyDict: inout [AnyAutoLayerKey: Any]) -> (InstanceType, Shape) {
         let inputShapeArray: [Int]
         if let inputShapeTuple = inputShape as? (Int, Int, Int) {
             inputShapeArray = [inputShapeTuple.0, inputShapeTuple.1, inputShapeTuple.2]
--- a/Models/LayerInit/AutoConv.swift
+++ b/Models/LayerInit/AutoConv.swift
@@ -37,7 +37,7 @@ public struct AutoConv2D<Scalar>: AutoLayer where Scalar: TensorFlowFloatingPoin
         self.biasInitializer = biasInitializer
     }
 
-    public func buildModelWithOutputShape(inputShape: (Int, Int, Int)) -> (InstanceType, (Int, Int, Int)) {
+    public func buildModelWithOutputShape<Prefix>(inputShape: (Int, Int, Int), keyPathSoFar: KeyPath<Prefix, InstanceType>, keyDict: inout [AnyAutoLayerKey: Any]) -> (InstanceType, (Int, Int, Int)) {
         let outputShape: (Int, Int, Int)
         if (padding == .valid) {
             outputShape = (
--- a/Models/LayerInit/AutoConv.swift
+++ b/Models/LayerInit/AutoConv.swift
@@ -1,0 +1,63 @@
+import TensorFlow
+
+public struct AutoConv2D<Scalar>: AutoLayer where Scalar: TensorFlowFloatingPoint {
+    let filterShape: (Int, Int)
+    let outputChannels: Int
+    let strides: (Int, Int)
+    let padding: Padding
+    let dilations: (Int, Int)
+    let activation: Conv2D<Scalar>.Activation
+    let useBias: Bool
+    let filterInitializer: ParameterInitializer<Scalar>
+    let biasInitializer: ParameterInitializer<Scalar>
+
+    public typealias InstanceType = Conv2D<Scalar>
+    public typealias InputShape = (Int, Int, Int)
+    public typealias OutputShape = (Int, Int, Int)
+
+    public init(
+        filterShape: (Int, Int),
+        outputChannels: Int,
+        strides: (Int, Int) = (1, 1),
+        padding: Padding = .valid,
+        dilations: (Int, Int) = (1, 1),
+        activation: @escaping Conv2D<Scalar>.Activation = identity,
+        useBias: Bool = true,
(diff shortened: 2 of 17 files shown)


saeta:
CC @dan-zheng @dabrahams

shadaj:
Responding to @saeta's questions:
1. Passing the `input` at the beginning/end reduces to a decision about what the "root" node of a layer graph should be. If we pass `input` at the beginning, the root node for any graph would have to be a node that specifies the input shape. This makes it harder to break down complex graphs into independent chunks. For example, consider a situation where we we want to split out the middle two layers in a graph.

```swift
let myModelDefinition = InputNode(shape: ...).then(Dense(...)).then(Dense(...)).then(Dense(...)).then(Dense(...))
```

If we require input shapes to be specified at the beginning, we have to manually compute the input shape for the sub-graph, which introduces more overhead for the developer and also makes it impossible to re-use the subgraph in another model where the input shape is different.

```swift
let mySubModel = InputNode(shape: ...).then(Dense(...)).then(Dense(...))
let myModelDefinition = InputNode(shape: ...).then(Dense(...)).then(mySubModel).then(Dense(...))
```

If we have users specify the input shape at the end, however, this is solved more naturally because the entire model is composed without being fixed to a specific input shape. We can plug in any input shape at the end and produce a model instance with the appropriate propagated shapes

```swift
let mySubModel = Dense(...).then(Dense(...))
let myModelDefinition = Dense(...).then(mySubModel).then(Dense(...))

// elsewhere

var myModelInstance = myModelDefinition.buildModel(inputShape: ...)
```

2. Couple reasons why I picked `.then`
+ guides type inference to automatically pick the correct `Scalar` type by looking at the previous layer in the graph (in the example code, we only have to specify `Float` for the first layer)
+ as you mentioned, guides IDE completion
+ supports any length of sequential layers since `.then` can build up a linked-list of types (compared to `Sequential` which AFAIK is limited in the number of layers since it is just code generated)

Not sure I understood the extension method issue, could you explain further?

shadaj:
Also, VGG16 works now, but before I proceed to implement more models with this API I'd like to discuss the type complexity being introduced by the `AutoLayer` protocol.

In the VGG16 implementation, for example, when breaking out the model into a function we have to explicitly state the return type. This type contains the full model structure, which results in some very long types. However, it seems that we need this type information in order to return the appropriate `InstanceType` when `buildModel` is called. Any ideas on how we could simplify/eliminate these long types?

shadaj:
Some early performance benchmark results:
ResNetCIFAR10 (training) - old: 543.36 examples/sec, new: 435.76 examples/sec
ResNetCIFAR10 (inference) - old: 5286.49, new: 5527.64 examples/sec

8bitmp3:
@shadaj Awesome work—just saw this. Thanks for the efforts. Just wanted give my 2 cents - I know some of the stuff is marked as `TODO ... oh no` 🙂 e.g. `...<AutoSequenced<AutoSequenced<AutoSequenced....` in VGG16. So, hopefully the new stuff will be human friendly and readable in the spirit of Swiftlang. 🙌

saeta:
re:crazy types, I wonder if [associated type inference of opaque result types](https://github.com/apple/swift-evolution/blob/master/proposals/0244-opaque-result-types.md#associated-type-inference) might be able to help us out here.

shadaj:
The last commit (4a9719a) sets up the initial infrastructure for a `LayerModule` that properties as a way to get automatic type inference for abstracted layer chunks. Right now, explicit types are still needed when there are multiple lines in the layer creation logic, but with https://github.com/apple/swift/pull/32223 we should have full type inference.

Examples/LeNet-MNIST/main.swift
```diff
@@ -33,7 +33,7 @@ let denseDef = AutoConv2D<Float>(filterShape: (5, 5), outputChannels: 6, padding
   .then(AutoDense(outputSize: 84, activation: relu))
   .then(KeyedAutoLayer(AutoDense(outputSize: 10), key: lastDenseKey))
 
-var (classifier, keys) = denseDef.buildModelWithKeys(inputShape: (28, 28, 1))
+var classifier = denseDef.buildModel(inputShape: (28, 28, 1))
```
8bitmp3, line 36: Thanks for this @shadaj and other instances of simpler code. `var classifier = ...` appears friendlier for the folks who use Pythonic frameworks. Also +1 for the separate`BuiltAutoLayer` struct and I think the naming is in line with other struct - `KeyedAutoLayer` - so it makes sense.

shadaj:
Interestingly, running with `-cross-module-optimization` results in the linker crashing with `SIGBUS (Misaligned address error)`, but this crash doesn't happen on master. Need to investigate further...

Models/LayerInit/TupleToArray.swift
```diff
@@ -0,0 +1,61 @@
+public func intTupleToArray(tuple: Any) -> [Int] {
```
saeta, line 1: One potential idea: consider making a dedicated type that is expressible by array literal (https://developer.apple.com/documentation/swift/expressiblebyarrayliteral)

shadaj:
@saeta hmm, not sure I completely understand how the shape-first approach simplifies layer definitions. Right now, the reason why we have `AutoModule` instead of just regular functions is for type inference, but functions do work as well. The only issue with functions is that the return type (which is very long) would have to be explicitly included in the code.

I _do_ really like the chained API style to reduce `.then` calls for simple sequential models. I'll work on getting that added and using that in the examples, which should give us a better idea of the complexity of the end-user experience.

Models/LayerInit/AutoBatchNorm.swift
```diff
@@ -0,0 +1,28 @@
...
+
+    public init(
+        axis: Int = -1,
+        momentum: Scalar = 0.99,
+        epsilon: Scalar = 0.001
+    ) {
+        self.axis = axis
+        self.momentum = momentum
+        self.epsilon = epsilon
+    }
```
saeta, line 20: This to me looks just like a closure capture.

Models/LayerInit/AutoConv.swift
```diff
@@ -0,0 +1,63 @@
...
+        self.filterShape = filterShape
+        self.outputChannels = outputChannels
+        self.strides = strides
+        self.padding = padding
+        self.dilations = dilations
+        self.activation = activation
+        self.useBias = useBias
+        self.filterInitializer = filterInitializer
+        self.biasInitializer = biasInitializer
+    }
```
saeta, line 38: This also just looks like a closure capture.

Models/LayerInit/AutoDense.swift
```diff
@@ -0,0 +1,19 @@
...
+    let activation: Dense<Scalar>.Activation
+
+    public typealias InstanceType = Dense<Scalar>
+    public typealias InputShape = Int
+    public typealias OutputShape = Int
+
+    public init(outputSize: Int, activation: @escaping Dense<Scalar>.Activation = identity) {
+        self.outputSize = outputSize
+        self.activation = activation
+    }
```
saeta, line 14: This also looks like a closure capture.

Models/LayerInit/AutoFunction.swift
```diff
@@ -0,0 +1,19 @@
...
+
+    public typealias InstanceType = Function<Input, Output>
+    public typealias InputShape = InputShape
+    public typealias OutputShape = OutputShape
+
+    public init(fnShape: @escaping (InputShape) -> OutputShape, fn: @escaping @differentiable (Input) -> Output) {
+        self.fnShape = fnShape
+        self.fn = fn
+    }
+
```
saeta, line 15: This also appears to be a hand-written closure capture.

Models/LayerInit/AutoLayerKey.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    public typealias InstanceType = Underlying.InstanceType
+    public typealias InputShape = Underlying.InputShape
+    public typealias OutputShape = Underlying.OutputShape
+
+    public init(_ underlying: Underlying, key: AutoLayerKey<InstanceType>) {
+        self.underlying = underlying
+        self.key = key
+    }
+
```
saeta, line 29: This also looks a bit like a closure capture.

Models/LayerInit/AutoPool.swift
```diff
@@ -0,0 +1,100 @@
...
+    public init(
+        poolSize: (Int, Int),
+        strides: (Int, Int) = (1, 1),
+        padding: Padding = .valid
+    ) {
+        self.poolSize = poolSize
+        self.strides = strides
+        self.padding = padding
+    }
+
```
saeta, line 21: This also looks like a closure capture.

(Also below.)

shadaj:
Ah, that makes a lot more sense now! I think I have an idea that can get the best of both worlds by implementing these layers as functions that return something similar to `AutoFunction` but with a function that returns a layer instance instead. Will work on implementing this!

Models/LayerInit/AutoFunction.swift
```diff
@@ -8,6 +9,13 @@ public struct AutoFunction<Input: Differentiable, Output: Differentiable, InputS
     public typealias InputShape = InputShape
     public typealias OutputShape = OutputShape
 
+    /**
```
dabrahams, line 12: nit: use `///` style.  I can get you a chat from May 28 that covers some reasons

rahulbhalley:
If this PR gets merged, there will be multiple ways to do the same thing i.e. both high and low level. But why is it important to have high-level abstract APIs? I'll answer myself: low-level is good for experienced ML people whereas high-level is good for newcomers to ML, and practitioners. But I think **one learns better with low-level programming**. So, why not leave high-level API implementations up to some new projects (for instance, similar to [PyTorch's projects](https://pytorch.org/ecosystem/) such as CrypTen, PyTorchLightning, Ignite, Kornia, BoTorch, Optuna, and so on)  and **keep S4TF low-level, simple, clean, and beautiful**. Similarly, `TrainingLoop` proposal https://github.com/tensorflow/swift-models/pull/586 seems like an another attempt to allow high-level APIs in S4TF. It seems like imitating Keras' APIs. Seriously,  this doesn't seem important. But just keep making new, principled APIs like this https://github.com/tensorflow/swift-apis/pull/693 where we made a radical decision about renaming `RNN` (very commonly used, misnomer) to `RecurrentLayer`.

Btw I saw your presentation @shadaj, why would you refer `Layer` conforming APIs as **legacy**?

And using `Auto` prefix to `Layer` APIs makes it seem like flooding new alternative version of original APIs. I believe there should be only one way to write a program (writing neural networks, here). Just take a look at iOS and Android development. iOS apps are easier to build because there's principled one way to design, and program apps, whereas in Android development things get messy because there are multiple ways to do the same thing.

Swift is a damn beautiful language, so I'd request to not flood S4TF with complicated APIs.

cc @dan-zheng @saeta @BradLarson @dabrahams

rahulbhalley:
**Update**

I realized I was wrong about suggesting to focus on only low-level APIs because high-level APIs seem equally important given that they don't replace low-level APIs. Secondly, I think it'll be a good idea if the proposed `Auto` prefixed APIs (`AutoConv2D`, for instance) replace the traditional `Conv2D`, but without `Auto` prefix, making this a standard approach to building models.

P.S. Please ignore my last suggestion comment.

saeta:
I recommend we convert this PR to a draft, as we're not ready to actually merge in the approach based on this direction, and instead continue to refine our thinking per https://github.com/tensorflow/swift/pull/515

shabalind:
Closing for now. Feel free to reopen if there is more progress!

Issue #582, opened 2020-06-04
Move pix2pix and CycleGAN examples from Batcher to Epochs


BradLarson:
As part of the process of migrating over to use Epochs instead of the older Batcher API, we need to migrate the following models over from Batcher:

- [x] [CycleGAN](https://github.com/tensorflow/swift-models/tree/master/CycleGAN)
- [x] [pix2pix](https://github.com/tensorflow/swift-models/tree/master/pix2pix)

Additionally, all dependencies to Batcher should be removed from Datasets and other modules in Package.swift.

PR #589, opened 2020-06-09
Switch off Batcher from CycleGAN

xihui-wu:
This PR resolves item 1 in https://github.com/tensorflow/swift-models/issues/582

### DIFF
--- a/CycleGAN/main.swift
+++ b/CycleGAN/main.swift
@@ -174,7 +174,6 @@ let bResultsFolder = try createDirectoryIfNeeded(path: FileManager.default
 
 var testStep = 0
 for testBatch in dataset.testing {
-// for testBatch in zippedTest {
     let realX = testBatch.domainA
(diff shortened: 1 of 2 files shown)


CycleGAN/main.swift
```diff
@@ -204,20 +167,16 @@ for epoch in 0 ..< epochs {
...
                                                                   .currentDirectoryPath + "/testA_results")
 let bResultsFolder = try createDirectoryIfNeeded(path: FileManager.default
                                                                   .currentDirectoryPath + "/testB_results")
 
 var testStep = 0
-for testBatch in zippedTest {
-    let realX = testBatch.0
-    let realY = testBatch.1
+for testBatch in dataset.testing {
+// for testBatch in zippedTest {
```
BradLarson, line 177: This looks like a comment that slipped through.

xihui-wu: Ha yes..

PR #593, opened 2020-06-09
Make WordSeg benchmarks support x10 as an option

asuhan:
Also convert WordSeg example to x10.


asuhan:
@BradLarson How are you running on your machine? I can look into it as well, for my runs it didn't recompile.

BradLarson:
@asuhan - I used the following to run the benchmarks:

```
swift run -c release Benchmarks measure --benchmark WordSegScore --batchSize 4 --batches 20 --warmupBatches 10 --training --x10
```

and 

```
swift run -c release Benchmarks measure --benchmark WordSegScoreAndGradient --batchSize 4 --batches 20 --warmupBatches 10 --training --x10
```

You can swap out `--x10` for `--eager` to compare eager / X10 timings. Saw a lot of ptxas in `top` during the X10 benchmarks. This was with CUDA 10.1 and the current nightly toolchain.

asuhan:
@BradLarson I've tried and there are no recompilations, at least on our side, with the `CUDA_CACHE_MAXSIZE` environment variable set to `2147483647`.

The first benchmark, `WordSegScore`, is indeed 4.4x slower with x10 on my GPU. The second one, `WordSegScoreAndGradient`, is 2.7x faster, which bodes well for training workloads. It's still worth looking into the slowdown, of course. Is this what you're seeing as well?

Issue #592, opened 2020-06-09
Move remaining examples from Batcher to Epochs


BradLarson:
As a follow-on to issue #582, we have a number of other models and datasets that should be migrated from the deprecated Batcher API to Epochs:

**Datasets:**
- [x] [ImageSegmentationDataset](https://github.com/tensorflow/swift-models/blob/13cd9724c41880793e509c7fd8388b516bee44a8/Datasets/ImageSegmentationDataset.swift)
- [x] [COCODataset](https://github.com/tensorflow/swift-models/blob/9b404f3035108ec2fce2ff5a3a48e07659633ede/Datasets/COCO/COCODataset.swift)
- [x] [OxfordIIITPets](https://github.com/tensorflow/swift-models/blob/13cd9724c41880793e509c7fd8388b516bee44a8/Datasets/OxfordIIITPets/OxfordIIITPets.swift)
- [x] [Imagenette](https://github.com/tensorflow/swift-models/blob/b31852f0f21de8c63a5900824d7658a3a30a1c80/Datasets/Imagenette/Imagenette.swift)
- [x] [Imagewoof](https://github.com/tensorflow/swift-models/blob/b31852f0f21de8c63a5900824d7658a3a30a1c80/Datasets/Imagenette/Imagewoof.swift) (ImageClassificationDataset can then be removed after these last two are done.)
- [x] [TensorPair](https://github.com/tensorflow/swift-models/blob/master/Datasets/TensorPair.swift)

**Models**

- [x] [LanguageModel notebook](https://github.com/tensorflow/swift-models/blob/23b09c0c4bcf727ec640ad7a684f4e94d56fc966/Examples/Notebooks/LanguageModel.ipynb)
- [x] [NeuMF-MovieLens](https://github.com/tensorflow/swift-models/tree/78dfc0520dbf8aceff872dc8b06b5561dd35c738/Examples/NeuMF-MovieLens)

BradLarson:
...and with PR #615, this is officially done.

PR #603, opened 2020-06-12
Converting Imagenette and Imagewoof datasets to Epochs

BradLarson:
As part of migrating off of the old Batcher API and onto Epochs, this converts the Imagenette and Imagewoof datasets over to the new API. The three examples that used these datasets have also been converted.

A Jupyter notebook that demonstrated the use of Batcher for setting up Imagenette has been removed, as it is no longer relevant with the new API.

This addresses two items in issue #592.


Examples/VGG-Imagewoof/main.swift
```diff
@@ -24,15 +25,15 @@ let optimizer = SGD(for: model, learningRate: 0.02, momentum: 0.9, decay: 0.0005
 
 print("Starting training...")
 
-for epoch in 1...90 {
+for (epoch, epochBatches) in dataset.training.prefix(epochCount).enumerated() {
```
8bitmp3, line 28: @BradLarson 👍 on migration to Epochs. In case I missed it in the discussions here and during the design meetings, does Epoch lib affect training ⏳ significantly vs Batcher? Thanks

BradLarson: @8bitmp3 - I don't believe I've seen a noticeable performance difference, Epochs and Batcher largely perform the same operations under the hood. The primary benefit is a cleaner, more extensible interface that went through a lot of review before migrating up to swift-apis. For example, in these datasets it's easier to see that batches are lazily constructed from files on disk by loading from their URL.

It does become a lot easier to move the latter parts of the processing operations onto X10, where we see some benefits for having normalization, etc. be performed as part of the X10 trace.

This also made it possible to do on-the-fly synthetic data generation for use in benchmarks, which impacted performance there (as opposed to generating a block of random data upfront and shuffling that).

8bitmp3: Thanks for your detailed answer @BradLarson. Cleaner interface is 👍 especially if users want to look under the hood (which is all Swift).

Fyi, I'm putting together a tutorial, and I received a strange little error which may only apply to my setup in Colab. I'm testing VGG right now with Epochs and Imagewoof after your last merge: 
- Using "master" branch, freshly updated for `Imagewoof` in `Datasets` and the training loop in [`VGG.swift`](https://github.com/tensorflow/swift-models/blob/master/Models/ImageClassification/VGG.swift ).
- I manually defined the `VGGBlock` struct (`Layer` protocol), then ran the cell. 
- Used `VGGBlock` to manually define the `VGG16` struct to set up the neural net architecture. Ran the cell.
- ```
  ...error: type of expression is ambiguous without more context
          let backbone = input.sequenced(through: layer1, layer2, layer3, layer4, layer5)
  ```
- Highlighting it here because yesterday I ran the old code with the `tensorflow-0.9` branch and it was OK. 
- This doesn't happen and all runs smooths if I define both the `VGGBlock` and `VGG16` structs in one cell.
- There is a chance I have a bug that's only applicable to my setup 🙂.

Issue #528, opened 2020-05-14
Determine cause of XLA compiler crash when using the GPU and synthetic datasets


BradLarson:
In the process of completing PR #507, I discovered that running image classification benchmarks using a synthetic dataset and the X10 backend triggers an XLA compiler crash:

> F tensorflow/compiler/xla/xla_client/xla_util.cc:90] Internal: From /job:localservice/replica:0/task:0:
> Unable to launch convolution with type forward and algorithm (1, none)
> [[{{node XRTExecute}}]]
> Aborted (core dumped)

To reproduce, run the command with a GPU attached and a CUDA-capable toolchain:

```bash
swift run -c release Benchmarks measure --benchmark ResNetCIFAR10 --training --x10 --synthetic
```

This only happens with GPU usage, not TPU or CPU.

We should first identify what is different about the use of synthetic data that triggers this, then see this can be prevented via a dataset change or a fix to the underlying implementation.

BradLarson:
The workaround in PR #609 addresses this in practice, and the fundamental issue is that traces (or at least GPU-bound ones) can execute slightly past the lifetime of the referenced model and optimizer. Further work will be needed to determine if this can be prevented in the general case, but for now this handles the near-term issues here.

PR #609, opened 2020-06-16
Workaround to prevent X10 crashes during benchmarks

BradLarson:
Larger X10 models occasionally cause crashes when run in our benchmarks, but these crashes were only occurring at the very end of the benchmark. Tracking this down, it appears to be due to X10 (at least on GPU) executing traces asynchronously, leading to them running slightly after the point that the model and optimizer have gone out of scope. Inserting a synchronous barrier that blocks until execution is finished prevents these crashes.

This is inserted immediately before the returns of the timing values, because I found that this is the only location that reliably prevents crashes. Placing this barrier in a `defer` doesn't achieve the desired results, and the barrier must be within the benchmark function where the model and optimizer are allocated.

This addresses issue #528, but a more robust fix may be needed at the API level.

### DIFF
--- a/Benchmarks/Models/ImageClassificationInference.swift
+++ b/Benchmarks/Models/ImageClassificationInference.swift
@@ -52,6 +52,11 @@ where
 
         for (epoch, epochBatches) in dataset.training.enumerated() {
             if case let .epochs(epochs) = duration, epoch >= epochs {
+                if backend == .x10 {
+                    // A synchronous barrier is needed for X10 to ensure all execution completes
+                    // before tearing down the model.
+                    LazyTensorBarrier(wait: true)
+                }
                 return batchTimings
             }
 
--- a/Benchmarks/Models/ImageClassificationInference.swift
+++ b/Benchmarks/Models/ImageClassificationInference.swift
@@ -65,9 +65,19 @@ where
                 batchTimings.append(batchTime)
                 currentBatch += 1
                 if case let .batches(batches) = duration, currentBatch >= batches {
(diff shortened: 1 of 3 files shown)


Benchmarks/Models/ImageClassificationInference.swift
```diff
@@ -65,10 +65,20 @@ where
                 batchTimings.append(batchTime)
                 currentBatch += 1
                 if case let .batches(batches) = duration, currentBatch >= batches {
+                    if backend == .x10 {
```
xihui-wu, line 68: Should we add this to line 55 as well ?

Benchmarks/Models/ImageClassificationTraining.swift
```diff
@@ -74,11 +74,21 @@ where
                 batchTimings.append(durationInMilliseconds(since: beforeBatch))
                 currentBatch += 1
                 if case let .batches(batches) = duration, currentBatch >= batches {
+                    if backend == .x10 {
```
xihui-wu, line 77: Same, should add before line 62?

Benchmarks/Models/ImageClassificationInference.swift
BradLarson: Good catch. I've added the same above.

Benchmarks/Models/ImageClassificationTraining.swift
BradLarson: Likewise, this has been added above. Thanks for pointing it out.

PR #236, opened 2019-12-04
Add support for minimal perfzero integration

shabalind:
This PR adds support for minimal integration with [perfzero](https://github.com/tensorflow/benchmarks) benchmark runner.

Integration works by exposing a perfzero-compatible benchmark interface in python that instantiates underlying Swift benchmarks in a subprocess. Python integration can automatically pick up all benchmarks listed in `Models.swift` without any changes to the python code. 

After this PR one could use unmodified perfzero in combination with a custom [dockerfile](https://github.com/tensorflow/benchmarks/pull/440) to collect median benchmark running time (`wall_time` in perfzero terms). Currently we don't expose any other metrics but this can be added later on.

PR #612, opened 2020-06-19
Port benchmarks to swift-benchmark

shabalind:
This PR introduces new target called `NewBenchmarks` that closely mirrors current `Benchmarks`, only based on [swift-benchmark](https://github.com/google/swift-benchmark). Some noteworthy differences compared to the old infra:

* Some of the flags have changed: no more `--batches` or `--epochs`, but rather `--iterations` for both training and inference.

* Default number of iterations is adaptive, and doesn't have to be statically specified in each benchmark. You can use `--min-time` to decide for how long a particular benchmark should be run (rather than static number of iterations).

* No more subcommands such as `measure` or `list-defaults`. There is only one mode that runs the benchmarks. To list all benchmarks one can run benchmarks with 0 iterations.

* No more split between `training` and `inference` benchmarks. Benchmark suites can define arbitrarily named benchmarks.

* Metrics are now computed in Swift, and never in python code when integrating with perfzero (see BenchmarkColumns.swift for the implementation).

* Settings are much more flexible now, and it's easy to add custom settings that are tailored to a particular suite. For example: `Length` for WordSeg.

* It's easy to specify global settings across all benchmarks, rather having to have duplicate defaults in multiple benchmark suites.


shabalind:
Latest changes bring PerfZero integration to feature parity. Overview:

* Everything is done in Swift now, Python bridge is not supposed to change any more. Previously it has grown quite a bit of custom and hard to maintain logic (i.e., different treatment of wordseg vs rest of benchmarks).

* Benchmark columns map to perfzero metrics, one-to-one. A few metrics are special in perfzero output (i.e., `wall_time`, `iterations`, `startup_time`) and should always be present for integration to work well.

* Benchmarks can be defined multiple times per suite to have various variants (i.e. x10 vs non-x10). It's quite easy to express this thanks to flexibility of settings in swift-benchmark.

* Each benchmark suite can now have a custom list of columns to show. So it's possible to define unique per-suite metrics (wordseg needed this). When you run everything from a console this outputs a wide column of all possible metrics vs all benchmarks, and blank if a pair is not defined. Filtering down benchmarks, automatically filters redundant columns.

Package.swift
```diff
@@ -23,6 +23,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.9.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.1")),
+        .package(url: "https://github.com/shabalind/swift-benchmark", .branch("topic/mixed-columns")),
```
shabalind, line 26: Temporarily we use my fork, until `topic/mixed-columns` is merged to `google/swift-benchmark`.

Benchmarks/BenchmarkColumns.swift
```diff
@@ -0,0 +1,67 @@
...
+  BenchmarkColumn.register(
+    BenchmarkColumn.registry["median"]!.renamed("time_median"))
+  BenchmarkColumn.register(
+    BenchmarkColumn.registry["min"]!.renamed("step_time_min"))
+  BenchmarkColumn.register(
+    BenchmarkColumn.registry["min"]!.renamed("time_min"))
+  BenchmarkColumn.register(
+    BenchmarkColumn.registry["max"]!.renamed("step_time_max"))
+  BenchmarkColumn.register(
+    BenchmarkColumn.registry["max"]!.renamed("time_max"))
```
shabalind, line 61: It's unfortunate that wordseg uses slightly different names for the same metrics. We can't change them because it's going to break historical performance charts.

Package.swift
shabalind: PR: https://github.com/google/swift-benchmark/pull/59

shabalind:
Another minor thing: I run manual tests as [described here](https://github.com/tensorflow/swift-models/pull/236#issuecomment-561791807) to make sure that new perfzero integration is indeed working end-to-end.

Package.swift
shabalind: This has been resolved in de1de67

Benchmarks/__init__.py
```diff
@@ -245,16 +128,24 @@ def discover_swift_benchmarks():
   and add them as new top-level types in the current module.
   """
 
-  g = globals()
-  defaults = subp.check_output([
-      'swift', 'run', '-c', 'release', 'Benchmarks', 'list-defaults', '--json'
+  output = subp.check_output([
+      'swift', 'run', '-c', 'release', 'NewBenchmarks',
```
BradLarson, line 132: Does this need to be changed back to `Benchmarks`?

Benchmarks/__init__.py
```diff
@@ -44,198 +44,81 @@ def _setup(self):
...
-  Returns a json-style object or None for additional
-  benchmark metadata such as flags that can be useful for debugging.
-  """
-
-  return
-
-
-def extract_metrics(result, variety, backend):
-  """Extract PerfZero metrics based on the measurements.
+cwd = '/workspace/perfzero/workspace/site-packages/swift-models'
```
BradLarson, line 51: Will changing the path here from `/workspace/benchmarks/perfzero` to `/workspace/perfzero` change the way this operates within the Docker image?

Benchmarks/Models/ImageClassificationTraining.swift
```diff
@@ -12,88 +12,64 @@
...
-    func run(backend: Backend) -> [Double] {
-        // Include model and optimizer initialization time in first batch, to be part of warmup.
-        // Also include time for following workaround to allocate memory for eager runtime.
-        var beforeBatch = timestampInMilliseconds()
+  // Note: The this initial eager-mode tensor computation is needed, or all GPU memory
+  // will be exhausted on initial allocation of the model.
+  // TODO: Remove the following tensor workaround when above is fixed.
+  let testTensor = Tensor<Float>([1.0, 2.0, 3.0])
+  let testTensor2 = Tensor<Float>([1.0, 2.0, 3.0])
+  let _ = testTensor + testTensor2
```
BradLarson, line 38: This workaround has to be placed before the first allocation of a device. Let's remove this small-tensor creation workaround here, and instead add the one line 
```swift
let _ = _ExecutionContext.global
```
right before `switch backend` in `var device`. That'll make it work for all benchmarks, in a little cleaner manner than I had done here..

Benchmarks/Models/ResNetImageNet.swift
```diff
@@ -12,60 +12,67 @@
...
-        switch(variety) {
-        case .inferenceThroughput:
-            return BenchmarkSettings(duration: .batches(1000), batchSize: 128, iterations: 10,
-                warmupBatches: 1, synthetic: true, backend: .eager)
-        case .trainingThroughput:
-            return BenchmarkSettings(duration: .batches(110), batchSize: 128, iterations: 1,
-                warmupBatches: 1, synthetic: true, backend: .eager)
-        }
+  func inference(state: inout BenchmarkState) throws {
+    print(String(reflecting: state.settings))
```
BradLarson, line 26: Do we need the print statements here and in the training case? They dump out quite a bit of information when running the benchmarks at the command line.

BradLarson:
I'm not quite sure where this is specified, but when I run the benchmarks, `startup_time`, `step_time_median`, `step_time_min`, and `step_time_max` all seem to be reported in nanoseconds, despite the `s` label. `wall_time` and other measurements seem to be in the right units.

Is there a way to limit the digits of precision in the command-line readout to keep it legible within a narrower terminal window?

Benchmarks/Models/ResNetImageNet.swift
shabalind: We don't need, that's an accidental debug print.

Benchmarks/__init__.py
shabalind: Yep, good call!

Benchmarks/__init__.py
shabalind: This seemed to be necessary while I was testing locally, but it's possible my env is slightly different from the one we use for benchmark dashboard. I'll change it back to the old path to be on the safe side.

shabalind:
Thanks for spotting the unit conversion issues! It's going to be fixed in https://github.com/google/swift-benchmark/pull/60

shabalind:
Default number formatter is going to be updated to use less space in https://github.com/google/swift-benchmark/pull/61

Benchmarks/__init__.py
shabalind: Fixed in 6d5fdc2

Benchmarks/__init__.py
shabalind: Reverted in 0f3a0cc

Benchmarks/Models/ResNetImageNet.swift
shabalind: Removed in ea74d15

Benchmarks/Models/ImageClassificationTraining.swift
shabalind: Workaround moved to `var device` in a60d185

shabalind:
PRs linked above are merged now!

Benchmarks/__init__.py
```diff
@@ -44,198 +44,81 @@ def _setup(self):
...
-def run_swift_benchmark(name, variety, backend):
-  print('running swift benchmark {} ({})'.format(name, variety))
+def run_swift_benchmark(name):
+  print('running swift benchmark {}'.format(name))
   # TODO: Remove the need for 2 warmup batches when we have better-shaped zero tangent vectors.
   output = subp.check_output([
-      'swift', 'run', '-c', 'release', 'Benchmarks', 'measure', '--benchmark',
-      name, '--' + variety, '--' + backend, '--warmupBatches', '2', '--json'
+      'swift', 'run', '-c', 'release', 'Benchmarks', 
+      '--filter', name, '--warmup-iterations', '2', '--format', 'json'
```
shabalind, line 59: @BradLarson It might be a good idea to specify larger `--min-time` (default is just 1s) here to gather more samples for better accuracy of the results while collecting results for internal dashboards. How much time can we afford to spend per one benchmark?

shabalind: **Note:** this time limit is not going to include initial 2 iterations for warmup.

BradLarson: So far, I've used the hardcoded iteration counts used for other benchmarks that we're comparing against (110 iterations for most, 20 iterations with 10 for warmup for WordSeg) to provide an exact comparison. Some of our benchmarks can run into tens of minutes, largely due to compile time, so I don't think we have a set limit for how long they can run.

When we start adding training accuracy benchmarks, I imagine those will take a while to run through the number of epochs used internally to compare accuracies.

shabalind: How about we set it to 5 minutes of consecutive measurements while running under perfzero? This shouldn't affect everyday development of swift-models because the actual default stays to 1s when running things manually.

BradLarson: Sure, we can do that. I don't think the number of overall measurements will cause any changes in the overall results (as long as the warmup iterations remain the same). You'd think that more would increase reliability, but our largest source of variability right now is the cloud instance-to-instance performance difference. Shouldn't hurt, though.

shabalind: Time limit increased to 5 minutes in 78c4227

Benchmarks/BenchmarkColumns.swift
texasmichelle: WordSeg does not have a very long history yet. If it's the right thing to do going forward, I'm not opposed to changing them as part of this transition.

Benchmarks/Models/WordSeg.swift
```diff
@@ -12,171 +12,137 @@
...
+    "iterations",
+    "time_median",
+    "time_min",
+    "time_max",
+  ])
+  return BenchmarkSuite(name: name, settings: WarmupIterations(10), columns) { suite in
+    suite.benchmark(
+      "sentence_4", settings: Length(4), Backend(.eager), function: function)
+    suite.benchmark(
+      "sentence_4_x10", settings: Length(4), Backend(.x10), function: function)
```
texasmichelle, line 52: Hm, is there a better way to set this up so that running defaults will test both eager and backend, but running an individual test does not run the same test twice? As an example, if I expect to run a single test with eager:
```
$ swift run -c release Benchmarks --filter WordSegScore.sentence_4 --eager
running WordSegScore: sentence_4... done! (6120.60 ms)
running WordSegScore: sentence_4_x10... done! (6071.35 ms)

name                        wall_time startup_time iterations time_median time_min time_max
-------------------------------------------------------------------------------------------
WordSegScore.sentence_4       4.644 s      3.544 s          3     0.337 s  0.334 s  0.429 s
WordSegScore.sentence_4_x10   4.595 s      3.500 s          3     0.334 s  0.333 s  0.428 s
$
```
It appears that the same test is run twice, both with eager.

shabalind: I think it's best to use more descriptive names, with good defaults in each configuration. In that case you never have to pass flags to override defaults, you just specify which variant to run via filter. 

Problem is that due to compat with old infra "_eager" is not included in the benchmark name so it's hard to filter on it, but you can use `swift run -c release Benchmarks --filter WordSegScore.sentence_4 --filter-not x10` to run just eager variant (I recently added support for `--filter-not` just for this purpose).

Benchmarks/BenchmarkColumns.swift
shabalind: I think we can discuss if it's worth changing to be more uniform in a follow-up discussion outside of this PR (it's a minor change).

PR #613, opened 2020-06-23
Exploring structural generic programming and layer APIs

saeta:
@dabrahams (in discussions with @shadaj and others) earlier today (yesterday) mentioned that leveraging [structural generic programming](https://github.com/google/swift-structural) might yield some interesting points in API design space as an alternative to the progress so far in https://github.com/tensorflow/swift-models/pull/584

I quickly hacked together this draft to see what it might look like. I initially focused on using structural generic programming to avoid needing to spell out the `callAsFunction` implementation. To see what this looks like in action, check out `SequentialExample.swift`.

Subsequent work can define the shape propagation information as appropriate.


saeta:
CC @shadaj @dabrahams @shabalind @BradLarson @pschuh @compnerd for state.

shabalind:
I am *really* excited to see the idea of building Differerentiable on top of Structural panning out quite well in practice!

StructuralModelBuilding/SequentialExample.swift
```diff
@@ -0,0 +1,151 @@
...
+    public var flatten = Flatten<Float>()
+    @SequentialSkip(passing: Type<Tensor<Float>>()) var denseSkipped = Dense<Float>(inputSize: 1, outputSize: 2)
+    public var dense = Dense<Float>(inputSize: 36 * 6, outputSize: 10)
+}
+
+public struct MyResidualModel: Module, Layer, SequentialLayer {
+    public var conv = Conv2D<Float>(filterShape: (5, 5, 3, 6))
+    public var pool = MaxPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    public var flatten = Flatten<Float>()
+    @ResidualConnection var denseSkipped = Dense<Float>(inputSize: 36 * 6, outputSize: 36 * 6)
```
shadaj, line 25: How would we express models where two different transformations need to be applied to the same data? As far as I understand, `ResidualConnection` assumes that one of the transformations will always be the identity function, but for example in ResNet this may not be the case if a projection is needed (https://github.com/tensorflow/swift-models/blob/master/Models/ImageClassification/ResNet.swift#L92).

StructuralModelBuilding/SequentialExample.swift
```diff
@@ -0,0 +1,151 @@
...
+    public var pool = MaxPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    public var flatten = Flatten<Float>()
+    public var dense = Dense<Float>(inputSize: 36 * 6, outputSize: 10)
+}
+
+public struct MyModelSkipping: Module, Layer, SequentialLayer {
+    public var conv = Conv2D<Float>(filterShape: (5, 5, 3, 6))
+    public var pool = MaxPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    public var flatten = Flatten<Float>()
+    @SequentialSkip(passing: Type<Tensor<Float>>()) var denseSkipped = Dense<Float>(inputSize: 1, outputSize: 2)
```
shadaj, line 17: Hmm, so `passing` is the type of the previous and following layer? I guess it can't be inferred since we're using properties.

StructuralModelBuilding/SequentialExample.swift
saeta: Yeah, there's definitely other ways to spell this. One thing we could do is instead of using property wrappers, we could instead have something like:

```swift
struct ParallelLayers<Lhs: Layer, Rhs: Layer>: Layer where Lhs.Input == Rhs.Input, ... {
  var lhs: Lhs
  var rhs: Rhs
  // TODO: make merge func configurable.
  @differentiable
  public func callAsFunction(_ input: Lhs.Input) -> Lhs.Output {
    return lhs(input) + rhs(input)
  }
}
```

which could represent the parallelism.

shadaj: That could work really well! It's interesting to see this combination of using both structural and regular functions/structs to compose layers.

Another (slightly crazier) idea is to offer a structural `ParallelLayer` where all the properties are executed on the same input. From what I see, the biggest advantage of the structural approach is you replace the keyed access to layers with regular property accesses since every element in the sequence has a label (the property name). Though I'm not sure how the user would define the merge function in this case since there aren't a fixed number of parallel layers.

saeta: Yup, that's definitely another potentially interesting point in design space. 

Perhaps you just require users to define a pair-wise reduction function. Alternatively, perhaps a `fold` function and initial value could be provided...

BradLarson:
Closing for now, pending future evaluation.

PR #617, opened 2020-06-29
Implement Deep Q-Network

seungjaeryanlee:
This is a mostly finished implementation of Deep Q-Network on the CartPole environment that needs some improvements. Following [the original paper (Mnih et al., 2015)](https://storage.googleapis.com/deepmind-media/dqn/DQNNaturePaper.pdf), the code has both experience replay and target network implemented.

I have identified three places of improvements:

- [x] Parallelize training
  - ~~The Q-network is given a batch of states and outputs a batch of Q-values for all actions (batch size x # actions). This should be transformed to a 1D vector (batch size x 1) that selects the Q-value of the action that was actually selected. In Python, `tf.gather_nd` seems to be the best option, but it does not seem to exist in `_Raw`. Is there a way to use this function?~~
  - `_Raw.gatherNd` exists
- [x] Refactor code
  - ~~There are a lot of commented out `print()` statements. These are left out to aid me as I work through the improvements, but they should be removed once I am finished.~~
- [x] Remove uses of `_Raw`
  - ~~As per @BradLarson 's suggestion, most `_Raw` commands can replaced by S4TF functions. Replace them accordingly.~~
- [x] Use Huber loss
   - ~~Huber loss is mentioned in the original paper and can help with training stability.~~
- [x] Use `:on` for Tensors to support X10 devices
- [x] Improve reproducibility
  - ~~The hyperparameters were chosen via a rough random search, and the results vary heavily.~~
- [x] Separate `gatherNd` into a separate file
  - For a more streamlined improvement to `swift-api`, separate the `gatherNd` redefinition to a new file.


Gym/DQN/main.swift
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    func train(batchSize: Int) {
+        // Don't train if replay buffer is too small
+        if replayBuffer.count >= batchSize {
+            // print("train | Start training")
+            let (tfStateBatch, tfActionBatch, tfRewardBatch, tfNextStateBatch) = replayBuffer.sample(batchSize: batchSize)
+
+            // TODO: Find equivalent function of tf.gather_nd in S4TF to parallelize Q-value computation (_Raw.gather_nd does not exist)
+            // Gradient are accumulated since we calculate every element in the batch individually
```
dan-zheng, line 148: > * In Python, `tf.gather_nd` seems to be the best option, but it does not seem to exist in `_Raw`. Is there a way to use this function?

`_Raw.gatherND` does exist, you can find it in [this source file](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Bindings/RawOpsGenerated.swift):

<details>

```swift
  /// Gather slices from `params` into a Tensor with shape specified by `indices`.
  ///
  /// `indices` is a K-dimensional integer tensor, best thought of as a
  /// (K-1)-dimensional tensor of indices into `params`, where each element defines a
  /// slice of `params`:
  ///
  ///     output[\\(i_0, ..., i_{K-2}\\)] = params[indices[\\(i_0, ..., i_{K-2}\\)]]
  ///
  /// Whereas in `tf.gather` `indices` defines slices into the `axis`
  /// dimension of `params`, in `tf.gather_nd`, `indices` defines slices into the
  /// first `N` dimensions of `params`, where `N = indices.shape[-1]`.
  ///
  /// The last dimension of `indices` can be at most the rank of
  /// `params`:
  ///
  ///     indices.shape[-1] <= params.rank
  ///
  /// The last dimension of `indices` corresponds to elements
  /// (if `indices.shape[-1] == params.rank`) or slices
  /// (if `indices.shape[-1] < params.rank`) along dimension `indices.shape[-1]`
  /// of `params`.  The output tensor has shape
  ///
  ///     indices.shape[:-1] + params.shape[indices.shape[-1]:]
  ///
  /// Note that on CPU, if an out of bound index is found, an error is returned.
  /// On GPU, if an out of bound index is found, a 0 is stored in the
  /// corresponding output value.
  ///
  /// Some examples below.
  ///
  /// Simple indexing into a matrix:
  ///
  /// ```python
  ///     indices = [[0, 0], [1, 1]]
  ///     params = [['a', 'b'], ['c', 'd']]
  ///     output = ['a', 'd']
  /// ```
  ///
  /// Slice indexing into a matrix:
  ///
  /// ```python
  ///     indices = [[1], [0]]
  ///     params = [['a', 'b'], ['c', 'd']]
  ///     output = [['c', 'd'], ['a', 'b']]
  /// ```
  ///
  /// Indexing into a 3-tensor:
  ///
  /// ```python
  ///     indices = [[1]]
  ///     params = [[['a0', 'b0'], ['c0', 'd0']],
  ///               [['a1', 'b1'], ['c1', 'd1']]]
  ///     output = [[['a1', 'b1'], ['c1', 'd1']]]
  ///
  ///
  ///     indices = [[0, 1], [1, 0]]
  ///     params = [[['a0', 'b0'], ['c0', 'd0']],
  ///               [['a1', 'b1'], ['c1', 'd1']]]
  ///     output = [['c0', 'd0'], ['a1', 'b1']]
  ///
  ///
  ///     indices = [[0, 0, 1], [1, 0, 1]]
  ///     params = [[['a0', 'b0'], ['c0', 'd0']],
  ///               [['a1', 'b1'], ['c1', 'd1']]]
  ///     output = ['b0', 'b1']
  /// ```
  ///
  /// Batched indexing into a matrix:
  ///
  /// ```python
  ///     indices = [[[0, 0]], [[0, 1]]]
  ///     params = [['a', 'b'], ['c', 'd']]
  ///     output = [['a'], ['b']]
  /// ```
  ///
  /// Batched slice indexing into a matrix:
  ///
  /// ```python
  ///     indices = [[[1]], [[0]]]
  ///     params = [['a', 'b'], ['c', 'd']]
  ///     output = [[['c', 'd']], [['a', 'b']]]
  /// ```
  ///
  /// Batched indexing into a 3-tensor:
  ///
  /// ```python
  ///     indices = [[[1]], [[0]]]
  ///     params = [[['a0', 'b0'], ['c0', 'd0']],
  ///               [['a1', 'b1'], ['c1', 'd1']]]
  ///     output = [[[['a1', 'b1'], ['c1', 'd1']]],
  ///               [[['a0', 'b0'], ['c0', 'd0']]]]
  ///
  ///     indices = [[[0, 1], [1, 0]], [[0, 0], [1, 1]]]
  ///     params = [[['a0', 'b0'], ['c0', 'd0']],
  ///               [['a1', 'b1'], ['c1', 'd1']]]
  ///     output = [[['c0', 'd0'], ['a1', 'b1']],
  ///               [['a0', 'b0'], ['c1', 'd1']]]
  ///
  ///
  ///     indices = [[[0, 0, 1], [1, 0, 1]], [[0, 1, 1], [1, 1, 0]]]
  ///     params = [[['a0', 'b0'], ['c0', 'd0']],
  ///               [['a1', 'b1'], ['c1', 'd1']]]
  ///     output = [['b0', 'b1'], ['d0', 'c1']]
  /// ```
  ///
  /// See also `tf.gather` and `tf.batch_gather`.
  ///
  /// - Parameters:
  ///     - params: The tensor from which to gather values.
  ///     - indices: Index tensor.
  ///
  /// - Output output: Values from `params` gathered from indices given by `indices`, with
  ///     shape `indices.shape[:-1] + params.shape[indices.shape[-1]:]`.
  @inlinable @inline(__always)
  public static func gatherNd<
    Tparams: TensorFlowScalar,
    Tindices: TensorFlowIndex
  >(
    params: Tensor<Tparams>,
    indices: Tensor<Tindices>
  ) -> Tensor<Tparams> {
    let nOutputs = Int(1)
    let op = makeOp("GatherNd", nOutputs)
    op.updateAttribute("Tparams", Tparams.tensorFlowDataType)
    op.updateAttribute("Tindices", Tindices.tensorFlowDataType)
    op.addInput(params)
    op.addInput(indices)
    return op.execute(Int(1))
  }
```

</details>

Is this helpful?

Gym/DQN/main.swift
```diff
@@ -0,0 +1,278 @@
...
+typealias Reward = Tensor<Float>
+
+class ReplayBuffer {
+    var states: Tensor<Float>
+    var actions: Tensor<Int32>
+    var rewards: Tensor<Float>
+    var nextStates: Tensor<Float>
+    let capacity: Int
+    var count: Int
+    var index: Int
```
dan-zheng, line 49: You can [set initial values for stored properties](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html) `var count` and `var index` here, which is a bit clearer than initializing them in the initializer:

```suggestion
    var count: Int = 0
    var index: Int = 0
```

Gym/DQN/main.swift
```diff
@@ -0,0 +1,278 @@
...
+
+            // TODO: Find equivalent function of tf.gather_nd in S4TF to parallelize Q-value computation (_Raw.gather_nd does not exist)
+            // Gradient are accumulated since we calculate every element in the batch individually
+            var totalGrad = qNet.zeroTangentVector
+            for i in 0..<batchSize {
+                let 𝛁qNet = gradient(at: qNet) { qNet -> Tensor<Float> in
+
+                    let stateQValueBatch = qNet(tfStateBatch)
+                    let tfAction: Tensor<Int32> = tfActionBatch[i][0]
+                    let action = Int(tfAction.makeNumpyArray()).unwrapped()
```
dan-zheng, line 155: You can use [`Tensor.scalarized()`](https://www.tensorflow.org/swift/api_docs/Structs/Tensor#scalarized) to concisely convert a scalar `Tensor<Scalar>` into a `Scalar`.
```suggestion
                    let action = Int(tfAction.scalarized())
```

Gym/DQN/main.swift
```diff
@@ -0,0 +1,278 @@
...
+                    let prediction: Tensor<Float> = stateQValueBatch[i][action]
+
+                    let nextStateQValueBatch = self.targetQNet(tfNextStateBatch)
+                    let tfReward: Tensor<Float> = tfRewardBatch[i][0]
+                    let leftQValue = Float(nextStateQValueBatch[i][0].makeNumpyArray()).unwrapped()
+                    let rightQValue = Float(nextStateQValueBatch[i][1].makeNumpyArray()).unwrapped()
```
dan-zheng, line 161:
```suggestion
                    let leftQValue = nextStateQValueBatch[i][0].scalarized()
                    let rightQValue = nextStateQValueBatch[i][1].scalarized()
```

Gym/DQN/main.swift
```diff
@@ -0,0 +1,278 @@
...
+
+    // Interact with environment
+    let action = agent.getAction(state: state, epsilon: startEpsilon * Float(maxEpisode - episodeIndex))
+    // print("action: \(action)")
+    var (nextState, reward, isDone, _) = env.step(action)
+    // print("state: \(state)")
+    // print("nextState: \(nextState)")
+    // print("reward: \(reward)")
+    // print("isDone: \(isDone)")
+    episodeReturn += Int(reward.makeNumpyArray().item()).unwrapped()
```
dan-zheng, line 246:
```suggestion
    episodeReturn += Int(reward.scalarized())
```

Gym/DQN/main.swift
```diff
@@ -0,0 +1,278 @@
...
+        self.action_space = env.action_space
+        self.observation_space = env.observation_space
+    }
+
+    func reset() -> Tensor<Float> {
+      let state = self.originalEnv.reset()
+      return Tensor<Float>(numpy: np.array(state, dtype: np.float32))!
+    }
+
+    func step(_ action: Tensor<Int32>) -> (Tensor<Float>, Tensor<Float>, PythonObject, PythonObject) {
```
dan-zheng, line 197: Please add labels to the tuple result type! From the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#label-closure-parameters):

> **Label tuple members and name closure parameters** where they appear in your API.
>
> These names have explanatory power, can be referenced from documentation comments, and provide expressive access to tuple members.

Since `TensorFlowEnvironmentWrapper` is a Swift wrapper around Python objects, I also think it makes sense to change the type of the "isDone" tuple element from `PythonObject` to `Bool`.

```suggestion
    func step(_ action: Tensor<Int32>) -> (state: Tensor<Float>, reward: Tensor<Float>, isDone: Bool, info: PythonObject) {
```

Gym/README.md
```diff
@@ -31,4 +31,5 @@ To build and run the models, run:
 swift run Gym-CartPole
 swift run Gym-FrozenLake
 swift run Gym-Blackjack
+swift run Gym-DQN
```
dan-zheng, line 34: Here's the current output of `swift run Gym-DQN`:

```console
$ swift run Gym-DQN
Episode 1 Return 36
Episode 2 Return 16
Episode 3 Return 23
Episode 4 Return 17
Episode 5 Return 20
Episode 6 Return 22
Episode 7 Return 17
Episode 8 Return 20
Episode 9 Return 10
Episode 10 Return 10
Episode 11 Return 16
Episode 12 Return 22
Episode 13 Return 20
Episode 14 Return 14
Episode 15 Return 14
Episode 16 Return 9
Episode 17 Return 19
Episode 18 Return 29
Episode 19 Return 49
Episode 20 Return 20
...
```

I wonder if the model is currently learning and improving?

Do you have ideas for improving the output format? It would be cool if we could visualize the CartPole environment.

Gym/DQN/main.swift
seungjaeryanlee: Interesting! I have been getting this error: `error: type '_Raw' has no member 'gather_nd'` so I thought it didn't exist.

Maybe it is a version issue. This will make my life a lot easier. Thanks!

Gym/README.md
seungjaeryanlee: The learning is quite unstable indeed, although RL runs tend to be like that in general. For now, I think a simple matplotlib plot with a sliding window average could suffice.

```swift
let plt = Python.import("matplotlib.pyplot")

plt.plot(episodeReturns)
plt.savefig("episodeReturns.png")
```

Gym/DQN/main.swift
seungjaeryanlee: Oh... I just realized it's not `gather_nd` but `gatherNd` since Swift uses lowerCamelCase. I guess that makes sense!

Gym/DQN/main.swift
seungjaeryanlee: Good point! Thank you for the suggestion.

Gym/DQN/main.swift
seungjaeryanlee: Amazing! This makes everything so much cleaner.

Gym/DQN/main.swift
seungjaeryanlee: Didn't know that I could add labels! Will do: they definitely make it more readable.

I realized that I did not used `isDone` in the target calculation. The correct target calculation is

reward + isDone * discount * max_a Q(s', a)

So `isDone` should probably be a `Tensor<Bool>` or `Tensor<Float>`. Thanks for pointing it out!

Gym/DQN/main.swift
seungjaeryanlee: I am getting the following error message: 

```
error: <Cell 103>:53:44: error: expression is not differentiable
                let predictionBatch = _Raw.gatherNd(params: stateQValueBatch, indices: tfFullIndices)
                                           ^

<Cell 103>:53:44: note: cannot differentiate functions that have not been marked '@differentiable' and that are defined in other files
                let predictionBatch = _Raw.gatherNd(params: stateQValueBatch, indices: tfFullIndices)
```

Is `gatherNd` not differentiable in S4TF? I believe it was differentiable in Python TF.

dan-zheng: `_Raw.gatherNd` doesn't yet have a registered derivative, so you can temporarily register one yourself.

[Here's some (somewhat obtuse) documentation](https://github.com/apple/swift/blob/master/docs/DifferentiableProgramming.md#derivative-functions) on derivative registration using the `@derivative(of:)` attribute. [This custom differentiation tutorial](https://www.tensorflow.org/swift/tutorials/custom_differentiation) has another simple example.

Here's an untested derivative implementation I ported from [Python TensorFlow's `_GatherNdGrad`](https://github.com/tensorflow/tensorflow/blob/r2.2/tensorflow/python/ops/array_grad.py#L691-L701):
```swift
import TensorFlow

extension _Raw {
  /// Derivative of `_Raw.gatherNd`.
  @inlinable
  @derivative(of: gatherNd)
  public static func vjpGatherNd<
    Scalar: TensorFlowFloatingPoint,
    Index: TensorFlowIndex
  >(
    params: Tensor<Scalar>,
    indices: Tensor<Index>
  ) -> (
    value: Tensor<Scalar>,
    pullback: (Tensor<Scalar>) -> Tensor<Scalar>
  ) {
    // Ported from TensorFlow Python reference implementation:
    // https://github.com/tensorflow/tensorflow/blob/r2.2/tensorflow/python/ops/array_grad.py#L691-L701
    let shapeTensor = Tensor<Index>(params.shapeTensor)
    let value = gatherNd(params: params, indices: indices)
    return (value, { v in
      let dparams = scatterNd(indices: indices, updates: v, shape: shapeTensor)
      return dparams
    })
  }
}
```

After you add this to `Gym/DQN/main.swift`, `error: expression is not differentiable` should go away!

Gym/DQN/main.swift
```diff
@@ -0,0 +1,305 @@
...
+        nextStateBatch: Tensor<Float>,
+        isDoneBatch: Tensor<Bool>
+    ) {
+        let randomIndices = Tensor<Int32>(numpy: np.random.randint(count, size: batchSize, dtype: np.int32))!
+
+        let stateBatch = _Raw.gather(params: states, indices: randomIndices)
+        let actionBatch = _Raw.gather(params: actions, indices: randomIndices)
+        let rewardBatch = _Raw.gather(params: rewards, indices: randomIndices)
+        let nextStateBatch = _Raw.gather(params: nextStates, indices: randomIndices)
+        let isDoneBatch = _Raw.gather(params: isDones, indices: randomIndices)
```
BradLarson, line 117: I'd prefer if we could avoid using _Raw operations directly in public models, because for layering purposes we've discouraged people accessing operations at that level (the underbar is an indication of this). We recently [purged all uses of _Raw from swift-models](https://github.com/tensorflow/swift-models/issues/225) and I'm reluctant to start adding them back in.

In this case, I believe [Tensor.gathering()](https://github.com/tensorflow/swift-apis/blob/e9df616445bad8e66b48ae47674dacbc34949466/Sources/TensorFlow/Operators/Basic.swift#L528) might be what you could use here.

Gym/DQN/main.swift
```diff
@@ -0,0 +1,305 @@
...
+            let (loss, gradients) = valueWithGradient(at: qNet) { qNet -> Tensor<Float> in
+                // Compute prediction batch
+                let npActionBatch = tfActionBatch.makeNumpyArray()
+                let npFullIndices = np.stack([np.arange(batchSize, dtype: np.int32), npActionBatch], axis: 1)
+                let tfFullIndices = Tensor<Int32>(numpy: npFullIndices)!
+                let stateQValueBatch = qNet(tfStateBatch)
+                let predictionBatch = _Raw.gatherNd(params: stateQValueBatch, indices: tfFullIndices)
+
+                // Compute target batch
+                let targetBatch: Tensor<Float> = tfRewardBatch + Tensor<Float>(tfIsDoneBatch) * self.discount * _Raw.max(self.targetQNet(tfNextStateBatch), reductionIndices: Tensor<Int32>(1))
```
BradLarson, line 182: Instead of `_Raw.max()` here, can you just use `max()`?

Gym/DQN/main.swift
```diff
@@ -0,0 +1,305 @@
...
+fileprivate extension Optional {
+    func unwrapped(file: StaticString = #filePath, line: UInt = #line) -> Wrapped {
+        guard let unwrapped = self else {
+            fatalError("Value is nil", file: (file), line: line)
+        }
+        return unwrapped
+    }
+}
+
+extension _Raw {
```
BradLarson, line 33: For `_Raw.gatherNd,` could we split this off into a separate file and make it an extension on Tensor, like with `Tensor.gathering()`? We could follow the template linked below for the `Tensor.gathering()` method, and instead of defining a derivative on `_Raw.gatherNd()`, do so for the new Tensor method. We could give it a more descriptive, Swift-like name in the process.

Once we've done that, and prototyped that the new method works in the context of this model, you'd then be able to add it to the core swift-apis themselves for use by others. This staged process would allow us to iterate on the method, naming, parameters, etc. here and then migrate into the core when ready.

Gym/DQN/main.swift
seungjaeryanlee: I was not aware! Thank you for the note. I will convert them to native operations.

Gym/DQN/main.swift
```diff
@@ -0,0 +1,305 @@
...
+    var index: Int = 0
+
+    init(capacity: Int) {
+        self.capacity = capacity
+
+        states = Tensor<Float>(numpy: np.zeros([capacity, 4], dtype: np.float32))!
+        actions = Tensor<Int32>(numpy: np.zeros([capacity], dtype: np.int32))!
+        rewards = Tensor<Float>(numpy: np.zeros([capacity], dtype: np.float32))!
+        nextStates = Tensor<Float>(numpy: np.zeros([capacity, 4], dtype: np.float32))!
+        isDones = Tensor<Bool>(numpy: np.zeros([capacity], dtype: np.bool))!
```
BradLarson, line 82: Rather than relying on Numpy to initialize these tensors, it would probably be more efficient to use `Tensor(zeros:on:)`. The `on:` parameter might be useful to specify, or you could leave that off for now, but we're trying to get all of our models to support the X10 device in addition to the default TF eager device and to do so you have to make sure Tensors are created on the right device. That's something we can worry about later, as a refinement to the model.

Gym/DQN/main.swift
seungjaeryanlee: Good point! It has been fixed.

Gym/DQN/main.swift
seungjaeryanlee: That makes sense! I just have a few related questions:

Q1. Is there a similar method for boolean tensors? I get the following error for `Tensor<Bool>(on: [capacity])`:

```
/home/seungjl/git/swift-models/Gym/DQN/main.swift:82:19: error: referencing initializer 'init(zeros:on:)' on 'Tensor' requires that 'Bool' conform to 'Numeric'
        isDones = Tensor<Bool>(zeros: [capacity])
                  ^
TensorFlow.Tensor:1:11: note: where 'Scalar' = 'Bool'
extension Tensor where Scalar : Numeric {
          ^
```

Q2. For the device, I plan to define `let device: Device = Device.default` and pass it to relevant classes and functions. Is there a way to test that it works for X10 devices?

Thanks!

dan-zheng: > Q1. Is there a similar method for boolean tensors? I get the following error for `Tensor<Bool>(on: [capacity])`:

You can use the general repeated value initializer [`Tensor.init(repeating:shape: on:)`](https://www.tensorflow.org/swift/api_docs/Structs/Tensor#/s:10TensorFlow0A0V9repeating5shape2onACyxGx_AA0A5ShapeVAA6DeviceVtcfc):

```swift
Tensor<Bool>(repeating: false, shape: [capacity], on: ...)
```

seungjaeryanlee:
I have tried various hyperparameter settings from existing DQN implementations, including [eaplatanios/swift-rl](https://github.com/eaplatanios/swift-rl), [higgsfield/RL-Adventure](https://github.com/higgsfield/RL-Adventure), and [TF-Agents](https://www.tensorflow.org/agents/tutorials/1_dqn_tutorial). None of them consistently got a score of 200, so I improved the model with three additional algorithms:

- Soft Target Network Update from [Lillicrap et al. (2015)](https://arxiv.org/abs/1509.02971)
- Double DQN from [Hasselt, Guez, and Silver (2015)](https://arxiv.org/abs/1509.06461)
- Combined Experience Replay from [Zhang and Sutton (2017)](https://arxiv.org/abs/1712.01275)

With these three improvements and the TF-Agents hyperparameters, I am now getting consistent 200s!

BradLarson:
I just pushed a commit that does a few things:

- Fixes segmentation faults on GPU eager mode execution.
- Creates an extension on Tensor for `gatherNd` (which I'm calling `dimensionalGathering` until we can figure out a better name or better way of expressing the difference in parameter names).
- Splits out the types from main.swift into a few files to better organize the components here.
- Replaces a few Numpy uses with native (and faster) Swift equivalents.
- Reformats the files per swift-format's style that we use for this repository.

The segmentation fault I was seeing was due to trying to set a Tensor at an index within a larger Tensor during `append()` in the ReplayBuffer. I've changed this to a simpler (and probably more performant) approach using arrays of tensors that are stacked when needed during the batch extraction. For a replay buffer, we might want to consider using pure scalars there, because they'll be faster to work with when interacting with environments that take and provide scalars. I've done this with other environments and gotten very fast speeds by converting them to Tensors only at the point where you want to use the accelerator (model training and agent logic).

There are a few places where Numpy functions are being used when there are Swift-native equivalents, and that can impact performance on tight loops around an environment. I've replaced a few of these, but I think that all the rest should also be able to be replaced with Swift functions.

This isn't always solving the environment (it seems to solve ~half of the time), so if that's unexpected, you may want to check over my replay buffer logic and other changes to make sure I didn't introduce a subtle error somewhere.

For organization, I thought there was a lot of code in main.swift, so I broke it out into separate files for the core types. I think that makes it a little easier to follow what's going on here with the various pieces.

seungjaeryanlee:
Thank you for the wonderful commit @BradLarson !

1. I agree that the replay buffer should store numbers as Swift native types instead of Tensors and only convert them in the `sample()` function. I forgot Swift was fast! I will make that change.
2. Converting NumPy functions to Swift is definitely a good idea. I could not find some of the functions that I wanted which was why I have been using NumPy, but using Swift functions makes much more sense.
3. The division of code into Gathering / Agent / ReplayBuffer / main makes sense to me. I am considering moving the `eval()` function to the agent later.
4. The issue with the degraded performance stems from the ReplayBuffer being changed from `class` to `struct`. As far as I know, that means `replayBuffer` is passed by value, so the `agent.replayBuffer` is never updated and is empty the entire time, and thus the training never occurs. I changed `ReplayBuffer` to be a class again and it solves CartPole in <200 episodes or <10000 steps.

BradLarson:
@seungjaeryanlee - Oh, right, I accidentally left the class->struct conversion in there from my testing. I had been trying to see if ReplayBuffer as a class was interacting badly with the autodiff system, thus the attempt at converting it to a struct. I totally forgot to convert that back, and that makes complete sense as to why that wasn't being mutated properly.

There might be an argument in the future for converting that to a struct and then using appropriate inout semantics throughout to utilize some advantages of value types while preserving appropriate mutation, but we're good for now in having this be a class. Good catch, and glad that was that simple to fix.

Do you feel ready to convert this from a draft to a full PR for final review and submission?

seungjaeryanlee:
Sounds good!

Gym/DQN/main.swift
```diff
@@ -0,0 +1,190 @@
...
+  ) {
+    let (state, reward, isDone, info) = originalEnv.step(action.scalarized()).tuple4
+    let tfState = Tensor<Float>(numpy: np.array(state, dtype: np.float32))!
+    let tfReward = Tensor<Float>(numpy: np.array(reward, dtype: np.float32))!
+    let tfIsDone = Tensor<Bool>(numpy: np.array(isDone, dtype: np.bool))!
+    return (tfState, tfReward, tfIsDone, info)
+  }
+}
+
+func eval(agent: Agent) -> Float {
```
dan-zheng, line 48: Minor naming suggestions, based on the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines):

```suggestion
func evaluate(_ agent: Agent) -> Float {
```

Idiomatic Swift tends to [avoid abbreviations](https://swift.org/documentation/api-design-guidelines/#avoid-abbreviations) in favor of clarity. The `agent:` argument label can be omitted because it repeats type information.

Gym/DQN/main.swift
```diff
@@ -0,0 +1,190 @@
...
+  // End-of-step
+  state = nextState
+}
+
+// Save learning curve
+plt.plot(episodeReturns)
+plt.title("Deep Q-Network on CartPole-v0")
+plt.xlabel("Episode")
+plt.ylabel("Episode Return")
+plt.savefig("dqnEpisodeReturns.png")
```
dan-zheng, line 168: Right now, `plt.savefig` saves images in the directory wherever `swift run Gym-DQN` is invoked, causing them to appear as untracked files in git.

It would be nice to save them in some standard `.gitignore`'d output directory. `/tmp` could be good for now, though it may not work on Windows.

Gym/DQN/Agent.swift
```diff
@@ -0,0 +1,145 @@
...
+extension Optional {
+  fileprivate func unwrapped(file: StaticString = #filePath, line: UInt = #line) -> Wrapped {
+    guard let unwrapped = self else {
+      fatalError("Value is nil", file: (file), line: line)
+    }
+    return unwrapped
+  }
+}
+
+struct Net: Layer {
```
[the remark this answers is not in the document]
seungjaeryanlee: I renamed in to `DeepQNetwork` and added some documentation! What do you think?

Gym/DQN/Agent.swift
```diff
@@ -0,0 +1,145 @@
...
+    l2 = Dense<Float>(inputSize: hiddenSize, outputSize: actionCount, activation: identity)
+  }
+
+  @differentiable
+  func callAsFunction(_ input: Input) -> Output {
+    return input.sequenced(through: l1, l2)
+  }
+}
+
+class Agent {
```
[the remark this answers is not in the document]
seungjaeryanlee: Good idea! I renamed it to `DeepQNetworkAgent`.

Gym/DQN/main.swift
seungjaeryanlee: Didn't know that! Thank you for the link. I have made the changes.

Gym/DQN/main.swift
seungjaeryanlee: `/tmp/` sounds like a good idea for now. Thank you for the suggestion!

Gym/DQN/ReplayBuffer.swift
```diff
@@ -0,0 +1,81 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import TensorFlow
+
+class ReplayBuffer {
```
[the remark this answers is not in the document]
seungjaeryanlee: I have added them!

Gym/DQN/Agent.swift
```diff
@@ -0,0 +1,145 @@
...
+  var qNet: Net
+  var targetQNet: Net
+  let optimizer: Adam<Net>
+  let replayBuffer: ReplayBuffer
+  let discount: Float
+  let minBufferSize: Int
+  let doubleDQN: Bool
+  let device: Device
```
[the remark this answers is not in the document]
seungjaeryanlee: I have added the comments with references to the original DQN paper and the Double DQN paper.

seungjaeryanlee:
I have added documentation comments, focusing mainly on the `ReplayBuffer` and `Agent` classes and the list of hyperparameters. I would love to hear your feedback on them!

Gym/DQN/Agent.swift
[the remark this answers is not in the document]
dan-zheng: Looks great, thanks!

BradLarson:
Thanks for working hard on getting this ready.

To follow on, I've created [an issue in swift-apis](https://github.com/tensorflow/swift-apis/issues/1058) to track lifting up the `dimensionGathering()` (or whatever we end up naming it) into swift-apis and [an issue](https://github.com/tensorflow/swift-models/issues/657) in swift-models to add a little extra documentation to the Gym Readme describing the Deep Q-Network model in the same way as the other Gym models as well as indicating the additional dependencies of Matplotlib and Numpy.

PR #618, opened 2020-07-01
Restructuring checkpoint files, adding an initial filesystem abstraction

BradLarson:
As a first step in preparing the checkpoint reading and writing functionality to be upgraded from swift-models to swift-apis, this reorganizes the checkpointing functionality into a new `Checkpoints` module. No other changes have been made to the interfaces or functionality, these files have just been reorganized.

Along with this reorganization, a prototype of a filesystem abstraction has been added and used within the checkpoint reader and writer. This attempts to place a layer above raw file reading and writing, to allow us to use alternative file handling mechanisms. The default is a Foundation-based filesystem, with the same functionality as currently exists.


brettkoonce:
🎁

PR #619, opened 2020-07-03
Remove Package.resolved and add to .gitignore

texasmichelle:


texasmichelle:
Happy to add this back if there's some reason we need this, but it seems like a hassle to maintain over time and it's unclear what the benefits are of having it.

BradLarson:
I know that Apple has advocated for [checking in the Package.resolved file](https://developer.apple.com/videos/play/wwdc2019/408/?time=1454) in certain circumstances, but I think for the way that we're using packages here it makes sense to rebuild it on demand.

PR #622, opened 2020-07-03
Add fine-grain layer benchmarks

shabalind:
This PR builds upon upon #612 and adds infra for fine-grain benchmarks for arbitrary `Layer<Float>`:

* Performance of the forward pass
* Performance of the gradient + forward pass 

Primary objective of these benchmarks is to evaluate performance of autodiff code generation. The difference between two benchmarks is the time taken by backwards pass that we want to improve upon.

The PR includes all of our image classification models, and all of their variants on common input sizes (Mnist, Cifar, ImageNet).

I also hand-picked a common largest batch size which works for all of the image classification models in eager mode on V100. it's relatively small but it allows us to compare benchmarks at the same batch size.

### DIFF
--- a/Benchmarks/Layers/LayerSuite.swift
+++ b/Benchmarks/Layers/LayerSuite.swift
@@ -17,50 +17,50 @@ import Datasets
 import ImageClassificationModels
 import TensorFlow
 
-func makeSyntheticBatch<Model>(
-  model: Model.Type,
+func makeRandomTensor(
   batchSize: Int,
+  dimensions: [Int],
   device: Device
-) -> (Tensor<Float>, Tensor<Int32>)
-where
-  Model: ImageClassificationModel, Model.TangentVector.VectorSpaceScalar == Float
-{
-  let dataset = SyntheticImageDataset<SystemRandomNumberGenerator>(
-    batchSize: batchSize,
-    labels: model.outputLabels,
-    dimensions: model.preferredInputDimensions,
-    entropy: SystemRandomNumberGenerator(),
-    device: device)
-
-  for epochBatches in dataset.training {
-    for batch in epochBatches {
-      return (batch.data, batch.label)
-    }
-  }
-
-  fatalError("unreachable")
+) -> Tensor<Float> {
+  var allDimensions = [batchSize]
+  allDimensions.append(contentsOf: dimensions)
+  let tensor = Tensor<Float>(
+    randomNormal: TensorShape(allDimensions), mean: Tensor<Float>(0.5, on: device),
+    standardDeviation: Tensor<Float>(0.1, on: device), seed: (0xffeffe, 0xfffe),
+    on: device)
+  return tensor
 }
 
-func forwardBenchmark<Model>(
-  model modelType: Model.Type
+func makeForwardBenchmark<CustomLayer>(
+  layer: CustomLayer.Type,
+  inputDimensions: [Int],
+  outputDimensions: [Int]
 ) -> ((inout BenchmarkState) throws -> Void)
 where
-  Model: ImageClassificationModel, Model.TangentVector.VectorSpaceScalar == Float
+  CustomLayer: Layer,
+  CustomLayer: DefaultInit,
+  CustomLayer.Input == Tensor<Float>,
+  CustomLayer.Output == Tensor<Float>,
+  CustomLayer.TangentVector.VectorSpaceScalar == Float
 {
   return { state in
     let settings = state.settings
     let device = settings.device
     let batchSize = settings.batchSize!
-    var model = Model()
-    model.move(to: device)
-    let (images, _) = makeSyntheticBatch(model: modelType, batchSize: batchSize, device: device)
+    var layer = CustomLayer()
+    layer.move(to: device)
 
-    var sink = TensorShape([])
+    let input = makeRandomTensor(
+      batchSize: batchSize,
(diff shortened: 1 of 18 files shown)


Benchmarks/Layers/LayerSuite.swift
```diff
@@ -0,0 +1,178 @@
...
+  inputDimensions inp: [Int],
+  outputDimensions outp: [Int],
+  batchSizes: [Int] = [4],
+  layer: @escaping () -> CustomLayer
+) -> BenchmarkSuite
+where
+  CustomLayer: Layer,
+  CustomLayer.Input == Tensor<Float>,
+  CustomLayer.Output == Tensor<Float>,
+  CustomLayer.TangentVector.VectorSpaceScalar == Float
```
shabalind, line 153: @saeta @dabrahams Initially I tried abstracting over `Float` but it seemed to require insurmountable amount of effort and search for new abstractions. If we can't write dtype polymorphic code that ourselves, can we really expect our users to be able to?

Benchmarks/Layers/LayerSuite.swift
```diff
@@ -0,0 +1,178 @@
...
+      device: device)
+
+    var sink: TensorShape = TensorShape([])
+
+    while true {
+      do {
+        try state.measure {
+          let result = layer(input)
+          // Force materialization of the lazy results.
+          sink = result.shape
```
shabalind, line 64: @pschuh Is this the best way to force materialization of the lazy tensor result, or would you do it differently?

Benchmarks/Layers/LayerSuite.swift
```diff
@@ -0,0 +1,178 @@
...
+
+    while true {
+      do {
+        try state.measure {
+          let result = TensorFlow.gradient(at: layer) { layer -> Tensor<Float> in
+            let predicted = layer(input)
+            return meanAbsoluteError(predicted: predicted, expected: output)
+          }
+          // Force materialization of the lazy results.
+          sink += result
```
shabalind, line 121: @pschuh Same question as above only in slightly different context.

shabalind:
A few more findings from trying to run this benchmarks in all configurations on v100:

* EfficientNet L2 variant fails on batch size 4 under x10, while succeeds under eager. Usually x10 takes less memory, but this model variant is an exception.
* MobileNet V3 crashes on x10 on pretty much on all variants, so I had to exclude it from being run with x10.

Benchmarks/Layers/LayerSuite.swift
```diff
@@ -0,0 +1,177 @@
...
+  }
+}
+
+func makeLayerSuite<CustomLayer>(
+  name: String,
+  inputDimensions inp: [Int],
+  outputDimensions outp: [Int],
+  batchSizes: [Int] = [4],
+  backends: [Backend.Value] = [.eager, .x10],
+  layer: @escaping () -> CustomLayer
```
saeta, line 146: You could consider making this an `@autoclosure`, which might tighten up the syntax at the use site (but alternatively, it might be more confusing, as `@autoclosure` parameters are typically not invoked more than once).

Benchmarks/Layers/LayerSuite.swift
saeta: Hmm, is there a substantial performance difference between this and just calling `LazyTensorBarier(wait: true)`? (The advantage of a LTB approach is that you would be sure to not transfer anything from the accelerator to the host (I don't think you do in this case either, because there are no shape-dynamic operations in XLA today, but it could change in the future), and it would simplify the end-of-benchmark codepath.)

Benchmarks/Layers/LayerSuite.swift
saeta: Agreed that the APIs we have today are not easy to use in a generic fashion. I suspect that as we revisit how we build layer APIs we should also reconsider aspects of our `Tensor` API. CC @shadaj

Benchmarks/Layers/LayerSuite.swift
dabrahams: Using @autoclosure is usually a bad idea outside of special circumstances like `&&` and `assert`.  It turns code that looks eager into lazy code.  Swift is generally explicit about laziness.

shabalind: I side with @dabrahams. Changing to autoclosure here would make it a bit too magical, especially considering that closure is going to be invoked multiple times.

Benchmarks/Layers/LayerSuite.swift
BradLarson: @saeta - I believe the problem is in how to mark the result as active, so that it is not ignored and optimized away. Something needs to be materialized at some point so that the calculation is marked as needing to be run. Or maybe there's another way of stating the result is needed without triggering a copy to the host.

pschuh: .shape does not materialize the lazy tensor result. You want a materialization that does not wait. some sort of: LazyTensorBarier(wait: false)

shabalind: Based on empirical experiments barrier type doesn't really affect performance, as long as it's called. 

I changed forward benchmark to add the result to a `sink` tensor instead of using .shape to make sure results are indeed computed in some form (in case x10 gets more lazy in the future), but it doesn't really affect results compared to the previous one that used .shape.

PR #623, opened 2020-07-04
Implement Multi GPU UNet X10 training example

### LINKED ISSUE #35: UNet Example
I'd love to contribute an example of using UNet for image segmentation - would this be useful?

tanmayb123:
This implements data parallelism for UNet on the OxfordIIITPets dataset.

Current implementation gets 3.25x performance when scaling from 1x to 4x RTX 2080 Tis.

2x Tesla P100 takes just over 2 minutes per epoch with a batch size of 16 (per GPU).

There's still a couple TODOs, unnecessary ops, etc. but I'd appreciate any feedback as I'm sure this can be made better.

Closes #35

### DIFF
--- a/UNet/main.swift
+++ b/UNet/main.swift
@@ -153,12 +153,6 @@ func unetWorkerLoss(model: UNet, input: Tensor<Float>, output: Tensor<Float>) ->
     tbSigmoidCrossEntropy(logits: model(input).reshaped(to: [-1, 256 * 256]), labels: output.reshaped(to: [-1, 256 * 256]))
 }
 
-func constantPtr<T>(_ value: T) -> UnsafeMutablePointer<T> {
-    let ptr = UnsafeMutablePointer<T>.allocate(capacity: 1)
-    ptr.pointee = value
-    return ptr
-}
-
 let gpus = 1
 let batchSize = 16
 let cpuDevice = Device(kind: .CPU, ordinal: 0, backend: .TF_EAGER)
--- a/UNet/main.swift
+++ b/UNet/main.swift
@@ -1,0 +1,214 @@
+import Dispatch
+import TensorFlow
+import Datasets
+import PythonKit
+
+let tqdm = Python.import("tqdm")
+
+// TODO: remove this terrible way of saving/loading weights
+let np = Python.import("numpy")
+extension Layer {
+    func weights() -> PythonObject {
+        var tensors: [PythonObject] = []
+        for kp in self.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self) {
+            tensors.append(self[keyPath: kp].makeNumpyArray())
+        }
+        return np.array(tensors)
+    }
+
+    mutating func load(weights: PythonObject) {
+        for (kpIdx, kp) in self.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self).enumerated() {
+            self[keyPath: kp] = Tensor<Float>(numpy: weights[kpIdx].astype("float32"))!
+        }
+    }
+}
+
+struct UNet: Layer {
+    struct ConvBlock: Layer {
+        var conv: Conv2D<Float>
+        var bn: BatchNorm<Float>
+
+        init(filterShape: (Int, Int, Int, Int)) {
+            self.conv = Conv2D(filterShape: filterShape, padding: .same)
+            self.bn = BatchNorm(featureCount: filterShape.3)
+        }
+
+        func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+            return relu(input.sequenced(through: conv, bn))
+        }
+    }
+
+    @noDerivative let encoderPooling = MaxPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+    var encoderConv1 = ConvBlock(filterShape: (3, 3, 3, 64))
+    var encoderConv2 = ConvBlock(filterShape: (3, 3, 64, 128))
+    var encoderConv3 = ConvBlock(filterShape: (3, 3, 128, 256))
+    var encoderConv4 = ConvBlock(filterShape: (3, 3, 256, 512))
+
+    @noDerivative let decoderUpsampling = UpSampling2D<Float>(size: 2)
+    var decoderConv1 = ConvBlock(filterShape: (3, 3, 768, 256))
+    var decoderConv2 = ConvBlock(filterShape: (3, 3, 384, 128))
+    var decoderConv3 = ConvBlock(filterShape: (3, 3, 192, 1))
+
+    func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+        let e1 = encoderConv1(input)
+        let e2 = e1.sequenced(through: encoderConv2, encoderPooling)
+        let e3 = e2.sequenced(through: encoderConv3, encoderPooling)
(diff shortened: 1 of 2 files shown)


UNet/main.swift
```diff
@@ -0,0 +1,214 @@
...
+        self.finished = finished
+        self.x = x
+        self.y = y
+        self.output = output
+        self.loss = loss
+    }
+}
+
+func gpuWorker<Model: Layer>(_ workerInput: GpuWorkerInput<Model>, loss: @differentiable (Model, Model.Input, Model.Output) -> (Tensor<Float>)) {
+    withDevice(.gpu, workerInput.device) {
```
BradLarson, line 100: `withDevice()` doesn't really do anything, so you should be fine in removing this and relying on your manual specification of Devices.

UNet/main.swift
```diff
@@ -0,0 +1,214 @@
...
+        let gpu = Device(kind: .GPU, ordinal: 0, backend: .XLA)
```
BradLarson, line 101: You've hardcoded an ordinal of 0 here, is that intentional, or should it use the value from your GPUWorkerInput?

UNet/main.swift
```diff
@@ -0,0 +1,214 @@
+import Dispatch
+import TensorFlow
+import Datasets
+import PythonKit
+
+let tqdm = Python.import("tqdm")
+
+// TODO: remove this terrible way of saving/loading weights
```
BradLarson, line 8: You should be able to use CheckpointReader and CheckpointWriter to read and write these without needing numpy. I'm working on a little cleaner / easier interface for those when serializing your own weights.

UNet/main.swift
```diff
@@ -0,0 +1,214 @@
...
+
+func constantPtr<T>(_ value: T) -> UnsafeMutablePointer<T> {
+    let ptr = UnsafeMutablePointer<T>.allocate(capacity: 1)
+    ptr.pointee = value
+    return ptr
+}
+
+let gpus = 1
+let batchSize = 16
+let cpuDevice = Device(kind: .CPU, ordinal: 0, backend: .TF_EAGER)
```
BradLarson, line 164: I'll note that even when .CPU is specified for an eager-mode device, if a GPU is available the eager mode will still automatically use the GPU. I don't know if this can be overridden at present.

UNet/main.swift
tanmayb123: A question about this and the ordinal issue — whenever I try to specify an ordinal other than `0` on a multi GPU machine, it errors out saying the device doesn't exist. However, when I used `withDevice` with the device index, it did seem to allocate memory on the GPU I specify.

UNet/main.swift
```diff
@@ -0,0 +1,214 @@
...
+            LazyTensorBarrier(on: gpu, wait: true)
+            workerInput.output = grad
+            workerInput.loss = loss.scalarized()
+
+            workerInput.outSemaphore.signal()
+        }
+    }
+}
+
+func averageTangentVectors(_ tangentVectors: [UNet.TangentVector]) -> UNet.TangentVector {
```
BradLarson, line 122: Pulling values to the eager-mode device, combining them there, updating the model via the eager-mode optimizer, then copying the model back over to the various devices is an interesting solution to the multi-GPU problem, but it might not be the most efficient. It's a little complicated to read, but an alternative approach that avoids a lot of this overhead is present in [the X10 training loop within swift-apis](https://github.com/tensorflow/swift-apis/blob/master/Sources/x10/swift_bindings/training_loop.swift). We're working to encapsulate that in the new training loop we're building in swift-models to make it easy to do this kind of multi-device training.

UNet/main.swift
BradLarson: Before I forget, thanks for exploring this and providing a motivating example. We haven't gotten far into our multi-GPU tests yet, so this helps to find new issues.

tanmayb123: Of course, I can't wait to start using Swift for TensorFlow for more demanding tasks once we can leverage multi GPU training.

I will run that test now, but the only thing I'm wondering is how exactly do we get the 3.25x speedup in training them with 4 GPUs vs 1? 1 GPU doesn't have enough memory for 4x the batch size so it couldn't just be that.

tanmayb123: Also I ran your test, and when I just print `Devices.allDevices`, I only see ordinal 0, but when I set the `XRT_DEVICE_MAP` environment variable as you mentioned, I only see ordinal 1.

mikowals:
Going through this code I don't think it is using X10 but is instead only training on the available GPUs using eager execution.   None of the tensors are ever copied to an XLA device (a `LazyTensorBarrier` the refers to GPU on XLA device but I don't think that is sufficient to force all executions to XLA).  

Could I suggest an initial example of UNet that is trained using the `TrainingLoop` structure?  I think having that to benchmark the models results vs other implementations.  It could also be useful for comparison to improvements with distributed training.

UNet/main.swift
BradLarson: I haven't tested this, but I think this is the right syntax to map two GPU devices for X10 (and if this works you can extend this to others):

```
export XRT_DEVICE_MAP='GPU:0;/job:localservice/replica:0/task:0/device:XLA_GPU:0|GPU:1;/job:localservice/replica:0/task:0/device:XLA_GPU:1'
```

I think the pipe character separator is needed to split the assignments. This environment variable is [mapping Swift for TensorFlow device IDs to the internal TF device format](https://github.com/tensorflow/swift-apis/blob/master/Sources/x10/xla_client/xrt_computation_client.h#L139).

Regarding work on the GPUs, I think @mikowals is correct in that these aren't being run on the XLA device (the Tensors and model aren't being copied over to an XLA device, from what I can see). I'm not entirely sure where the eager mode is allocating the Tensors and running operations here. You might potentially be seeing performance improvements just from running operations in parallel, which could hide overheads and better use even a single GPU.

UNet/main.swift
```diff
@@ -0,0 +1,208 @@
...
+        let e2 = e1.sequenced(through: encoderConv2, encoderPooling)
+        let e3 = e2.sequenced(through: encoderConv3, encoderPooling)
+        let e4 = e3.sequenced(through: encoderConv4, encoderPooling)
+        let d1i = decoderUpsampling(e4).concatenated(with: e3, alongAxis: 3)
+        let d1 = decoderConv1(d1i)
+        let d2i = decoderUpsampling(d1).concatenated(with: e2, alongAxis: 3)
+        let d2 = decoderConv2(d2i)
+        let d3i = decoderUpsampling(d2).concatenated(with: e1, alongAxis: 3)
+        let d3 = decoderConv3(d3i)
+        return d3
```
marcrasi, line 63: I noticed a problem with the last layer. It has a relu nonlinearity but there should be a last layer without a nonlinearity.

BradLarson:
With the work that @xihui-wu is doing right now on adding multi-device training support to our training loop, and the concerns we've expressed about the ways that parallel operation is performed here, I think this probably isn't the direction we want to pursue for multi-device training. I'll close out the draft PR for now, but the UNet model and its training on the OxfordIIITPets dataset might be useful by itself, if you wanted to separate that out into a training example using our current APIs.

Thank you for experimenting with this, and for motivating us to explore how this should work, as well as surfacing issues with how the X10 devices are exposed. This was a very useful test case.

PR #631, opened 2020-07-09
A higher-level interface to model checkpointing

BradLarson:
CheckpointReader and CheckpointWriter provide the building blocks for reading and writing model tensors to TensorFlow v2 checkpoint files, but to use them you have had to write model-specific code for extracting Tensors and their names, as well as placing parameters back into the model. This introduces a higher-level API that makes it easy to checkpoint and restore simpler models, and provides customization options for more complex models or for matching the naming schemes of external frameworks.

Writing a checkpoint becomes as easy as

```swift
extension LeNet: Checkpointable {}

let model = LeNet()

try model.writeCheckpoint(to: directory, name: "LeNet")
```

and reading

```swift
var newModel = LeNet()

try newModel.readCheckpoint(from: directory, name: "LeNet")
```

This is handled via the new `Checkpointable` protocol, which builds on `KeyPathIterable`. If you extend a model to be `Checkpointable` you will gain the `writeCheckpoint(to:...)` and `readCheckpoint(from:...)` methods for writing and reading checkpoints. By default, this will generate path names for each `Tensor<Float>` within your model using a naming scheme where each level in the model is named by its property name and separated by "/". For example: `conv1/filter`.

This can be overridden in a number of ways. The internal separator string can be changed by providing a different `checkpointSeparator` in the protocol conformance. Specific tensors within the model can be ignored for checkpointing by providing them in `ignoredTensorPaths` in the protocol conformance.

For the greatest flexibility, customized String -> String mapping between internal tensor path names and those in the on-disk checkpoint can be provided as a default for the model via `tensorNameMap` in the protocol conformance, or as a parameter when reading or writing a specific checkpoint. This allows for interoperability between a wide variety of frameworks and model naming conventions.

Documentation and tests have been provided for checkpointing overall, and this new functionality in particular. This is now ready for review.


BradLarson:
I'm merging this in for now, to allow for the documentation to be accessible and to unblock work on improvements that depend on this.

PR #632, opened 2020-07-09
Split benchmarks into lib and executable

texasmichelle:
To reproduce these benchmarks in other systems, it would be useful to have the ability to import them as a library. This PR splits the existing benchmarks into a library and executable.


texasmichelle:
The most obvious question I have is around naming. Can we come up with something better than `BenchmarksLib` and `Benchmarks`?

BradLarson:
`BenchmarkCore` for the library? I still like `Benchmarks` for the executable target, because `swift run Benchmarks` has a nice flow to it.

shabalind:
Another alternative for naming: `Benchmarks` for core/lib and `BenchmarksCli` or `BenchmarksMain` for the runnable target.

texasmichelle:
I agree about the utility of `swift run Benchmarks`.  All three of `Benchmarks|BenchmarksLib|BenchmarksCore` for swift-models are confusing wrt `Benchmark` for `swift-benchmark`, so let's go with `BenchmarksCore` for now and I'm happy to modify when a better suggestion comes up.

PR #634, opened 2020-07-15
Add README to explain how to add new settings

shabalind:

PR #635, opened 2020-07-15
Add `datasetFilePath` argument for benchmarks

texasmichelle:
Adds `datasetFilePath` argument and passes the value to WordSeg benchmarks. This is useful for specifying file locations other than the default download location.

### DIFF
--- a/BenchmarksCore/BenchmarkArguments.swift
+++ b/BenchmarksCore/BenchmarkArguments.swift
@@ -34,16 +34,20 @@ public struct BenchmarkArguments: ParsableArguments {
   @Flag(help: "Use real data.")
   var real: Bool
 
+  @Option(name: .customLong("datasetFilePath"), help: "File path for dataset loading.")
+  var datasetFilePath: String?
+
   public init() {}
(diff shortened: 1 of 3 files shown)


texasmichelle:
This uses CamelCase for local consistency with `batchSize`. #636 filed for global consistency with defaults.

BenchmarksCore/BenchmarkArguments.swift
```diff
@@ -34,16 +34,20 @@ public struct BenchmarkArguments: ParsableArguments {
   @Flag(help: "Use real data.")
   var real: Bool
 
+  @Option(name: .customLong("datasetFilePath"), help: "File path for dataset loading.")
```
shabalind, line 37: I have a slight preference to migrate away from camelCase if possible as mentioned in https://github.com/tensorflow/swift-models/pull/634#discussion_r455688551

texasmichelle: Agreed - I can tackle #636 next week

PR #637, opened 2020-07-16
Added UNet for Image Segmentation

Andr0id100:
I need some help with this model implementation.
It works currently and provides decent results but I want to improve it a bit.

Is it possible for a Layer to have multiple inputs and/or outputs as that would vastly improve the model readability and it would better translate the representation of the model from the paper.

Any ideas on variable naming specifically for the "Up" and "Down" modules. Referencing some implementations they seem to be frequently used but I would love something more descriptive.
Also suggestions for naming of the outputs of layers would also be greatly appreciated. I seem to have hit a naming block.

I will add the model to build manifest and add tests soon. Also will modify the model to take image shape as input during the initialization.

### DIFF
--- a/Models/ImageSegmentation/UNet.swift
+++ b/Models/ImageSegmentation/UNet.swift
@@ -1,0 +1,106 @@
+import TensorFlow
+
+public struct DoubleConv: Layer {
+
+    var conv1: Conv2D<Float>
+    var conv2: Conv2D<Float>
+
+    public init(inputFilters: Int, outputFilters: Int) {
+        conv1 = Conv2D<Float>(
+            filterShape: (3, 3, inputFilters, outputFilters),
+            strides: (1, 1),
+            padding: .same,
+            activation: relu
+        )
+        conv2 = Conv2D<Float>(
+            filterShape: (3, 3, outputFilters, outputFilters),
+            strides: (1, 1),
+            padding: .same,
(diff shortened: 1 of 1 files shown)


BradLarson:
The [pix2pix](https://github.com/tensorflow/swift-models/tree/master/pix2pix) example, PR #623, and this PR all have different implementations of the U-Net model. Would it be possible to create a single canonical model that could be used in each of these cases and at least rework the pix2pix example to use that shared model?

We may also need a matching image segmentation dataset and example to truly judge the training of the segmentation model by itself, so it would be great to have those.

Would Upsampling and Downsampling (or some variant) be more descriptive? That's what those layers do in the architecture.

Where do you need to have multiple inputs or outputs? The pix2pix implementation just uses one Tensor in, one Tensor out, for the various layers. The model as a whole should be able to take in a Tensor for the input image and return a Tensor for the segmented image.

Andr0id100:
Since opening this PR nearly 3 weeks ago, I have been accepted in a university research program. Consequently, it has been very chaotic trying to move across the country during a pandemic.
Because of this I haven't been able to devote any time to working on the PR.
If this still requires work when things settles down for me, I will be happy to open a new PR.

BradLarson:
That's totally understandable, and we appreciate the time you've put into this so far. Good luck with the new position.

PR #640, opened 2020-07-20
Pin swift-benchmark to a specific revision.

saeta:
Swift-Benchmark currently has an unstable API. This change pins
Swift-Benchmark to an explicit revision to ensure stable builds.


shabalind:
Is this a duplicate of #641 ?

shabalind:
*Edit:* It's not, two PRs are for different branches (master vs tensorflow-0.11).

BradLarson:
@shabalind - Just checking on the status, did we need to modify this to bring it in?

shabalind:
@BradLarson Changes here look good to me! We should probably do a 0.1 release on swift-benchmark some time soon to make it a bit more user friendly than a particular commit hash.

saeta:
+1 to doing a 0.1 release. Things seem reasonably stable presently. Thanks!

PR #641, opened 2020-07-20
Pin Swift-Benchmark to a specific revision for a consistent release.

saeta:


saeta:
Related PR: https://github.com/tensorflow/swift-models/pull/640

shabalind:
Alternative idea: maybe we should just tag 0.1 for swift-benchmark instead of using hashes?

saeta:
I think for going forward, that makes sense. for the S4TF v0.11 release, I think pinning to a specific commit is probably the best, as we need a commit that's not the latest.

PR #642, opened 2020-07-20
Remove camelCase override

### LINKED ISSUE #636: Use default word separator in CLI arguments
Currently, there is a mixture of word separator usage in CLI arguments, e.g. [`batchSize`](https://github.com/tensorflow/swift-models/blob/master/BenchmarksCore/BenchmarkArguments.swift#L23), [`min-time`](https://github.com/tensorflow/swift-models/blob/master/Benchmarks/__init__.py#L60).

Both of these override the default underscores provided by `ArgumentParser`. It would be good to standardize on the default in all cases.

texasmichelle:
Fixes #636 

Removes camelCase overrides for benchmark arguments, using the [swift-argument-parser](https://github.com/apple/swift-argument-parser) default provided by [swift-benchmark](https://github.com/google/swift-benchmark). This changes the `batchSize` argument to `batch-size` and `datasetFilePath` to `dataset-file-path`.

### DIFF
--- a/BenchmarksCore/BenchmarkArguments.swift
+++ b/BenchmarksCore/BenchmarkArguments.swift
@@ -19,7 +19,7 @@ public struct BenchmarkArguments: ParsableArguments {
   @OptionGroup()
   var arguments: Benchmark.BenchmarkArguments
 
-  @Option(name: .customLong("batchSize"), help: "Size of a single batch.")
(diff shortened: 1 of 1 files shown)


texasmichelle:
Is there anywhere else we need to test this change?

texasmichelle:
Excellent! That was less hairy than I expected. Thanks for verifying!

PR #644, opened 2020-07-21
Add support for electra

Shashi456:
Clone #534 

The build issues have been fixed.

The `electra-small` model is not supported due to a different hidden size than the formula permits, I mentioned this issue in #534 as well.

> Although the hidden_size = embedding_size generally for Electra, but in the case of electra_small, the embedding size is 128 while the hidden size is 256, any idea how I could solve in that particular case?

Other than that, build passes.


Shashi456:
@BradLarson @dabrahams

Shashi456:
@BradLarson i've been wondering about this for a while but what do you think is a decent way for testing these pretrained language models we have, even though i was careful while writing the tensor values, I still dont know if that is robust. Do you have any thoughts on this?

PR #646, opened 2020-07-21
Shift x10 transformations from the BERT-CoLA example into the CoLA dataset

shadaj:
Eliminates device copying in the example code by allowing users to specify a device to load CoLA onto.

### DIFF
--- a/Datasets/CoLA/CoLA.swift
+++ b/Datasets/CoLA/CoLA.swift
@@ -114,7 +114,7 @@ extension CoLA {
     maxSequenceLength: Int,
     batchSize: Int,
     entropy: Entropy,
-    on: Device,
+    on device: Device,
     exampleMap: @escaping (CoLAExample) -> LabeledTextBatch
   ) throws {
     self.directoryURL = taskDirectoryURL.appendingPathComponent("CoLA")
@@ -159,8 +159,8 @@ extension CoLA {
     ).lazy.map { (batches: Batches) -> LazyMapSequence<Batches, LabeledTextBatch> in
       batches.lazy.map{ 
         (
-          data: $0.map(\.data).paddedAndCollated(on: on, to: maxSequenceLength),
-          label: Tensor(copying: Tensor($0.map(\.label)), to: on)
+          data: $0.map(\.data).paddedAndCollated(on: device, to: maxSequenceLength),
+          label: Tensor(copying: Tensor($0.map(\.label)), to: device)
         )
       }
     }
@@ -168,8 +168,8 @@ extension CoLA {
     // Create the validation collection of batches.
     validationBatches = validationExamples.inBatches(of: batchSize / maxSequenceLength).lazy.map{ 
       (
-        data: $0.map(\.data).paddedAndCollated(on: on, to: maxSequenceLength),
-        label: Tensor(copying: Tensor($0.map(\.label)), to: on)
+        data: $0.map(\.data).paddedAndCollated(on: device, to: maxSequenceLength),
+        label: Tensor(copying: Tensor($0.map(\.label)), to: device)
       )
     }
   }
--- a/Datasets/CoLA/CoLA.swift
+++ b/Datasets/CoLA/CoLA.swift
@@ -114,7 +114,7 @@ extension CoLA {
     maxSequenceLength: Int,
     batchSize: Int,
     entropy: Entropy,
-    on device: Device,
+    on device: Device = .default,
     exampleMap: @escaping (CoLAExample) -> LabeledTextBatch
   ) throws {
     self.directoryURL = taskDirectoryURL.appendingPathComponent("CoLA")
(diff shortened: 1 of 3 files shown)


Examples/BERT-CoLA/main.swift
```diff
@@ -19,7 +19,13 @@ import TensorFlow
...
 import x10_optimizers_optimizer
 
-let device = Device.defaultXLA
+// Until https://github.com/tensorflow/swift-apis/issues/993 is fixed, default to the eager-mode
+// device on macOS instead of X10.
+#if os(macOS)
+  let device = Device.defaultTFEager
+#else
+  let device = Device.defaultXLA
+#endif
```
BradLarson, line 28: I think this model may actually be one of the few that runs well on macOS (and may in fact be faster using X10 there). Do you experience a crash with X10 and this model on macOS? The image classification models I'd tagged like this each crashed when I tried to run them on macOS, but I thought the Transformer-based models all worked with X10 there.

shadaj: I've been testing on Ubuntu, so wasn't sure if the model runs fine with x10 on macOS. I'll switch this back to just using x10 on all platforms.

Datasets/CoLA/CoLA.swift
```diff
@@ -114,6 +114,7 @@ extension CoLA {
     maxSequenceLength: Int,
     batchSize: Int,
     entropy: Entropy,
+    on: Device,
```
BradLarson, line 117: The pattern we've followed elsewhere is to have this be specified by `on device: Device` so that we're referring to the easier-to-understand `device` within the code, rather than `on`. It might make sense to replace the arguments here and elsewhere throughout the changes below with that pattern.

Support/Text/TextBatch.swift
```diff
@@ -55,20 +55,20 @@ extension TextBatch: Collatable {
 extension Collection where Element == TextBatch {
   /// Returns the elements of `self`, padded to `maxLength` if specified
   /// or the maximum length of the elements in `self` otherwise.
-  public func paddedAndCollated(to maxLength: Int? = nil) -> TextBatch {
+  public func paddedAndCollated(on: Device, to maxLength: Int? = nil) -> TextBatch {
```
texasmichelle: How about making this optional with a default of default device? The general pattern is for `device` to be the last parameter, so I'd like to see that here. +1 to @BradLarson's suggestion of `on device: Device`.

Datasets/CoLA/CoLA.swift
```diff
@@ -114,6 +114,7 @@ extension CoLA {
     maxSequenceLength: Int,
     batchSize: Int,
     entropy: Entropy,
+    on device: Device,
```
texasmichelle, line 117: How about making this optional with a default of defaultDevice?

Datasets/CoLA/CoLA.swift
```diff
@@ -183,13 +184,15 @@ extension CoLA where Entropy == SystemRandomNumberGenerator {
     taskDirectoryURL: URL,
     maxSequenceLength: Int,
     batchSize: Int,
+    on: Device,
```
texasmichelle, line 187: Same suggestion about optionality.

Examples/BERT-CoLA/main.swift
```diff
@@ -62,8 +62,10 @@ var cola = try CoLA(
   taskDirectoryURL: workspaceURL,
   maxSequenceLength: maxSequenceLength,
   batchSize: batchSize,
-  entropy: SystemRandomNumberGenerator()
+  entropy: SystemRandomNumberGenerator(),
+  on: device
 ) { (example: CoLAExample) -> CoLA.LabeledTextBatch in
+  // must be eager, since the text has not been padded yet
```
texasmichelle, line 68: Can you clarify this comment a bit?

Support/Text/TextBatch.swift
```diff
@@ -55,20 +55,20 @@ extension TextBatch: Collatable {
 extension Collection where Element == TextBatch {
   /// Returns the elements of `self`, padded to `maxLength` if specified
   /// or the maximum length of the elements in `self` otherwise.
-  public func paddedAndCollated(to maxLength: Int? = nil) -> TextBatch {
+  public func paddedAndCollated(on device: Device, to maxLength: Int? = nil) -> TextBatch {
```
texasmichelle, line 58: Same comment about device optionality, as well as moving it to the end to follow the general pattern.

Examples/BERT-CoLA/main.swift
```diff
@@ -62,8 +62,11 @@ var cola = try CoLA(
   taskDirectoryURL: workspaceURL,
   maxSequenceLength: maxSequenceLength,
   batchSize: batchSize,
-  entropy: SystemRandomNumberGenerator()
+  entropy: SystemRandomNumberGenerator(),
+  on: device
 ) { (example: CoLAExample) -> CoLA.LabeledTextBatch in
+  // in this closure, both the input and output text batches must be eager
```
texasmichelle, line 68: Nit: capitalization & punctuation at the end of the sentence.

PR #649, opened 2020-07-25
feat: Added ImageLoader

ayushi2103:
Integrated JPEGTurbo with swift-models using CMake.


ayushi2103:
@googlebot I signed it!

CMakeLists.txt
```diff
@@ -79,13 +79,39 @@ ExternalProject_Add(swift-protobuf
...
+    git://github.com/libjpeg-turbo/libjpeg-turbo
+  GIT_TAG
+    master
+  UPDATE_DISCONNECTED
+    TRUE
+  CMAKE_ARGS
+    -D CMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+    -D ENABLE_SHARED=NO
+  INSTALL_COMMAND
+    ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_SOURCE_DIR}/ImageLoader/libjpeg-turbo/module.modulemap <SOURCE_DIR>
```
compnerd, line 93: This isn't exactly the best option IMO.  This modifies the source tree for JPEG Turbo and that repository is hidden away in the build tree.  It would be nicer if we could copy out the headers into a staging directory instead.

CMakeLists.txt
```diff
@@ -79,13 +79,39 @@ ExternalProject_Add(swift-protobuf
...
+  BUILD_BYPRODUCTS
+    <BINARY_DIR>/${CMAKE_STATIC_LIBRARY_PREFIX}turbojpeg${CMAKE_STATIC_LIBRARY_SUFFIX}
+  STEP_TARGETS
+    build)
+
+ExternalProject_Get_Property(libjpeg-turbo BINARY_DIR)
+ExternalProject_Get_Property(libjpeg-turbo SOURCE_DIR)
```
compnerd, line 100: Because these overwrite the variables (`BINARY_DIR` and `SOURCE_DIR`), could you please re-order the rules such that the `SwiftProtobuf` library is completely added before handling the JPEG Turbo library?

ImageLoader/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+
+
```
compnerd, line 2: Nit: Extraneous whitespace

ImageLoader/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
...
+add_subdirectory(JPEG)
+add_subdirectory(ImageOps)
+
```
compnerd, line 5: Nit: Extraneous whitespace

ImageLoader/ImageOps/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+   ${BINARY_DIR})
+   
+target_link_libraries(ImageOps 
+turbojpeg)
+
+install(TARGETS ImageOps
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+  )
```
compnerd, line 21: Nit: missing newline

ImageLoader/ImageOps/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+
+add_library(ImageOps 
+  SHARED
```
compnerd, line 3: The rest of the project places `SHARED` on the previous line, could you please follow the same style?

ImageLoader/ImageOps/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  ImageOps.swift)
+
+set_target_properties(ImageOps PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
+
+target_include_directories(ImageOps PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}
+  ${SOURCE_DIR}
+   ${BINARY_DIR})
```
compnerd, line 12: Nit: indentation is off.

ImageLoader/ImageOps/ImageOps.swift
```diff
@@ -0,0 +1,88 @@
+
+import Foundation
+import TurboJPEG
+
+// Loads image
+public func tjJPEGLoadCompressedImage( filename: UnsafePointer<Int8>?, width: inout Int32, align: inout Int32,  height: inout Int32, pixelFormat: inout Int32, inSubsamp: Int32, flags: Int) -> UnsafeMutablePointer<UInt8>? {
```
compnerd, line 6: Does this need to be exposed?  If the library is the image operations, they should be high level enough that this shouldn't matter IMO.  Alternatively, we could call this `TurboJPEGOperations`.

ImageLoader/ImageOps/ImageOps.swift
```diff
@@ -0,0 +1,88 @@
+
```
compnerd, line 1: Could you please `swift-format` this file?

ImageLoader/ImageOps/ImageOps.swift
```diff
@@ -0,0 +1,88 @@
...
+import Foundation
+import TurboJPEG
+
+// Loads image
+public func tjJPEGLoadCompressedImage( filename: UnsafePointer<Int8>?, width: inout Int32, align: inout Int32,  height: inout Int32, pixelFormat: inout Int32, inSubsamp: Int32, flags: Int) -> UnsafeMutablePointer<UInt8>? {
+  
+    pixelFormat = -1
+ 
+    /* Read the JPEG file into memory. */
+    var jpegFile = fopen(filename, "rb")
```
compnerd, line 11: `fopen` is not available.  Please add this to the beginning of the file:

```swift
#if os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
import Darwin
#elseif os(Android) || os(Linux)
import Glibc
#elseif os(Windows)
import ucrt
#else
#error "C library not known for the target OS"
#endif
```

ImageLoader/ImageOps/ImageOps.swift
```diff
@@ -0,0 +1,88 @@
...
+    fseek(jpegFile, 0, SEEK_END)
+    let size = ftell(jpegFile)
+    fseek(jpegFile, 0, SEEK_SET)
+    let jpegSize = CUnsignedLongLong(size)
```
compnerd, line 15: What is the purpose of this variable?  You immediately cast it away to `Int32`.

ImageLoader/ImageOps/ImageOps.swift
```diff
@@ -0,0 +1,88 @@
...
+    var jpegBuf = (tjAlloc(Int32(jpegSize)))
+    fread(jpegBuf, Int(jpegSize), 1, jpegFile)
+    fclose(jpegFile)
+    jpegFile = nil
```
compnerd, line 19: If this should be in the C style, this definitely needs error handling throughout.  However, this all seems overly complicated for what it seems like - reading the contents of the file.  Why does this need to be done manually rather than something like:

```swift
guard let data = FileManager.default.contents(atPath: filename) else {
  // log.error("unable to load contents of \(filename)")
  return nil
}
let decompressor: ? = tjInitDecompress()
data.withUnsafeBytes {
  tjDecompressHeader(decompressor, $0.baseAddress!, UInt(size), &width, &height)
}
```

ImageLoader/ImageOps/ImageOps.swift
compnerd, line 6: Nit: Extraneous space after `(`.

The first formal parameter (`filename`) is ambiguous.  What is `UnsafePointer<Int8>?`?  Is it a UTF-8 encoding string?  Perhaps UTF-16 so that it is compatible with Cocoa and Win32.  It is probably best to pass that as a Swift style `String` to allow for the encoding to be dealt with properly.

I don't really like the myriad of `inout` parameters here.  It seems that you are using them to return the image data.  Why not redesign the API as:

```
enum PixelFormat: Int32 {
case ...
}
struct ImageData {
  let height: Int32
  let width: Int32
  let alignment: Int32
  let pixelFormat: PixelFormat
  let data: [UInt8]
}
private func LoadJPEG(atPath path: String, flags: Int = 0) -> ImageData? {
}
```

This allows for the data associated with the image content to returned as a single entity and be easily tracked.  It also means that future loaders can easily have a similar API, making them easy to swap out.  Lastly, it avoids the `UnsafeMutablePointer` which makes memory management a trickier issue.

Is the cost of the additional copy here the concern?  Is it the unnecessary zero'ing?

Additionally, the `subsample` parameter to the load function seems unused.  Is there a reason for that parameter?

Are the pixel formats specific to JPEG?  I assume that the pixel format is referring more to something like ARGB8888 vs RGB565, etc.

ImageLoader/ImageOps/ImageOps.swift
```diff
@@ -0,0 +1,88 @@
...
+    tjFree(jpegBuf)
+    jpegBuf = nil
+    tjDestroy(tjInstance)
+    tjInstance = nil
+    
+    return imgBuf
+   
+}
+ //Saves image
+public func tjJPEGSaveImage(filename: UnsafePointer<Int8>?, buffer: UnsafePointer<UInt8>?, width: Int32, pitch: Int32, height: Int32, pixelFormat: Int32, outSubsamp: Int32, flags: Int32) -> Int32 {
```
compnerd, line 49: Similar issues with the `filename` parameter.  The image details seem better encapsulated as `ImageData` IMO.  The `outSubsamp` parameter is unused, what is the purpose of that?

ImageLoader/ImageOps/ImageOps.swift
```diff
@@ -0,0 +1,88 @@
...
+        - `pixelFormat` = format to which image is being compressed.
+        - Image gets compressed into `jpegBuf` buffer
+        - initializes `jpegSize` as size of image in bytes
+        - `outSubsamp` = the level of chrominance subsampling to be used when generating the JPEG image
+        - `outQual` = the image quality of the generated JPEG image (1 = worst, 100 = best), 95 taken as default
+        - `flags` = the bitwise OR of one or more of the flags
+    */
+    tjCompress2(tjInstance, buffer, width, 0, height, pixelFormat, &jpegBuf, &jpegSize, outSubsamp, outQual, flags)
+    
+    if (fwrite(jpegBuf, Int(jpegSize), 1, jpegFile) == 1){
```
compnerd, line 74: Again why not use `FileManager` from `Foundation` to write the file?

ImageLoader/JPEG/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
+target_link_libraries(JPEG PRIVATE
+ ImageOps)
+
+set_target_properties(JPEG PROPERTIES
+  Swift_LANGUAGE_VERSION 4.2)
+
+install(TARGETS JPEG
+  DESTINATION bin
+  COMPONENT exe)
```
compnerd, line 12: This component is poorly named - I imagine that we can have more than one executable.  More importantly, you are installing a library but naming the component `executable` which is very confusing.

ImageLoader/JPEG/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_library(JPEG
```
compnerd, line 1: I think that we should merge `ImageOps` and `JPEG`.  These don't make sense as separate libraries.  The `ImageOps` are not generic image operations currently, they are JPEG specific.  Why not make that `JPEGImageOperations`?  Why is that library separate?  Do you foresee other uses of it beyond just the `JPEG` library?

CMakeLists.txt
ayushi2103: Okay. I will copy it.

ImageLoader/ImageOps/CMakeLists.txt
ayushi2103: Will do.

ImageLoader/JPEG/CMakeLists.txt
ayushi2103: I will merge the files.

CMakeLists.txt
ayushi2103: Will do.

ImageLoader/ImageOps/ImageOps.swift
ayushi2103: Yes, passing the filename as a Swift style `String` can be done. 
 
So what I had in mind while writing this function signature was to give the user a similar experience as to the functions declared in libjpeg.
 
But actually we only need/store `width`, `height`, `pixelFormat` and `data` for our operations, so as you said, we can redesign the API but maybe just to only include `width`, `height`, `pixelFormat` and `data`, thereby removing the `alignment`, `subsample` and `pitch` parameters.
 
In the current implementation, I'm taking `flags` as an input but it can cause an error on user's side, hence I'm thinking of maybe removing that too from the function signature, giving it a 0 value by default, and change everything to - 
 
 
```
enum PixelFormat: Int32 {
case ...
}
struct ImageData {
  let height: Int32
  let width: Int32
  let pixelFormat: PixelFormat
  let data: [UInt8]
}
private func LoadJPEG(atPath path: String) -> ImageData? {
}
```
 
Yes, the Pixel Formats are specific to JPEG I believe. You can take a look at the formats [here](http://jpeg-turbo.dpldocs.info/libjpeg.turbojpeg.TJPF.html).
 
What are your views on this? Should we give the user the ability to choose the alignment, pitch and flags, or hardcode them to default value as 0 inside function?

compnerd: I don't think that there is much value to providing the same API surface as TurboJPEG as this is a helper function.  If there is some reason that I am not seeing, that might be important.

The pixel formats are specific to TurboJPEG only in their return'ed value.  The enumeration is generic, and I think that we should have our enumeration as the same encodings are prevalent in other image formats.

```swift
enum PixelFormat: Int8 {
case RGB888,      // TJPF_RGB
case BGR888,      // TJPF_BGR
case RGBA8888, // TJPF_RGBA
case BGRA8888, // TJPF_BGRA
case ARGB8888, // TJPF_ARGB
case ABGR8888, // TJPF_ABGR
case RGBA8880, // TJPF_RGBX
case BGRA8880, // TJPF_BGRX
case ARGB0888, // TJPF_XRGB
case ABGR0888, // TJPF_XBGR
case YUV400,     // TJPF_GREY
}
```

The encodings from that link are explicitly YUV 4:0:0 for the greyscale, RBG/BGR 888, and ARGB/ABGR/RGBA/BGRA 8888.  The reason for actually explicitly special casing the ARGB/ABGR 0888 RGBA/BGRA 8880 is to actually see if we can take advantage of that when constructing the embedding and using a tensor with a lower rank (@BradLarson should know if that is ever useful).  In the case that the reduced rank tensor is not useful, we should drop the 0888/8880 formats in favour of the 8888 equivalents.

BradLarson: Most image models are built around the assumption of a 3-channel input tensor (ignoring the alpha channel), but there are some that take advantage of that extra channel. I think it would make sense to have the ability to generate 3- or 4-channel image tensors based on whether the image on disk has a specified alpha channel.

I agree with Saleem's suggestion that we should have a Swift-like interface to the loading and saving functions, rather than mirroring a C interface. Taking in a String filename and returning a tuple or struct of parameters would make this easier to use for people on the Swift side. I'd even go so far as to suggest that instead of a name like `tjJPEGLoadCompressedImage` we might prefer `loadCompressedJPEGImage`.

We probably also want these to `throw` errors if there are failures on reading or writing, which will be common for image loading and saving functions.

ImageLoader/JPEG/JPEG.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import ImageOps
+
+// this function loads and saves image
+func imageLoadSave(imagePath: String,  savedImagePath: String) -> Int32 {
```
BradLarson, line 5: Is this intended as a unit test for saving and loading a file? If so, it might belong in a dedicated test target. Usually what I do for those is save a known set of bytes, then load them back in and verify they're the same. That might not work so well for a lossy JPEG compression, so we'd have to figure out how to test that things are approximately the same.

ayushi2103:
@compnerd @BradLarson can you please review them now. I've made the changes requested.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+            return 1
+        }
+    }
+}
+ 
+public struct ImageData {
+    var height: Int32
+    var width: Int32
+    var data: UnsafeMutablePointer<UInt8>
+    var formatProperties: pixelFormats
```
compnerd, line 59: Can these not be `let` bindings?

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+    var height: Int32 = 0
+    
+    let data: Data
+    
+    if FileManager.default.fileExists(atPath: path) {
+        data = FileManager.default.contents(atPath: path)!
+    } else {
+        // File not present
+        fatalError("File does not exist at: \(path).")
+    }
```
compnerd, line 81: This is better to write as:

```swift
guard FileManager.default.fileExists(atPath: path) else {
  fatalError("File does not exist at \(path).")
}
let data: Data = FileManager.default.contents(atPath: path)!
```

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+    var width: Int32 = 0
+    var height: Int32 = 0
+    
+    let data: Data
+    
+    if FileManager.default.fileExists(atPath: path) {
+        data = FileManager.default.contents(atPath: path)!
+    } else {
+        // File not present
+        fatalError("File does not exist at: \(path).")
```
compnerd, line 80: Is this not a bit extreme?  Why not `throw`?

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+    }
+    
+    let jpegSize: UInt = UInt(data.count)
+    let baseAddress: UnsafeRawPointer = data.withUnsafeBytes { return $0.baseAddress! }
```
compnerd, line 84: This is not valid.  You cannot escape the `baseAddress`.  You need to do all the work inside the closure.

```swift
return data.withUnsafeBytes {
  let data: UnsafeMutablePointer<UInt8> = UnsafeMutablePointer<UInt8>(mutating: $0.baseAddress!.assumingMemoryBound(to: UInt8.self))

  var decompressor = tjInitDecompress()
  defer { tjDestroy(decompressor) }

  tjDecompressHeader(decompressor, data, jpegSize, &width, &height)

  let buffer = tjAlloc(imageFormat.channelCount * width * height)
  tjDecompress2(decompressor, data, UInt(jpegSize), buffer, width, 0, height, 0, imageFormat.rawValue, 0)

  return ImageData(height: height, width: width, data: buffer!, imageFormat: imageFormat)
}
```

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+    do {
+        FileManager.default.createFile(atPath: path, contents: jpegData)
+        retVal = 0
+    } catch {
+        // File not created
+        fatalError("Error during image saving: \(error)")
+    }
+    
+    /* Free/Destroy instances and buffers */
+    tjDestroy(compressor)
```
compnerd, line 155: Use a `defer` block for this.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+    compressor = nil
+    tjFree(jpegBuf)
```
compnerd, line 157: And this.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+        FileManager.default.createFile(atPath: path, contents: jpegData)
+        retVal = 0
+    } catch {
+        // File not created
+        fatalError("Error during image saving: \(error)")
+    }
+    
+    /* Free/Destroy instances and buffers */
+    tjDestroy(compressor)
+    compressor = nil
```
compnerd, line 156: This is unnecessary after the `defer` block, because you know its valid for the duration of the function and cannot accidentally use it after free.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+    tjFree(jpegBuf)
+    jpegBuf = nil
```
compnerd, line 158: Similar.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,161 @@
...
+ 
+    let bufferPointer  = UnsafeMutableBufferPointer.init(start: jpegBuf, count: Int(jpegSize))
+    let jpegData = Data.init(buffer: bufferPointer)
+    
+    do {
+        FileManager.default.createFile(atPath: path, contents: jpegData)
+        retVal = 0
+    } catch {
+        // File not created
+        fatalError("Error during image saving: \(error)")
```
compnerd, line 151: Is this not a bit extreme?  Why not throw an error?

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: Okay. Will do

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: Okay.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: Yes, I will change that.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: Okay yes.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -120,42 +115,40 @@ func SaveJPEG(atPath path : String, image: ImageData) -> Int32 {
...
+     - `pixelFormat` = image.formatProperties.rawValue which denotes Pixel Format to which image is being compressed.
      - Image gets compressed into `jpegBuf` buffer
      - initializes `jpegSize` as size of image in bytes
      - `outSubsamp` = 0
      - `outQual` = the image quality of the generated JPEG image (1 = worst, 100 = best), 95 taken as default
      - `flags` = 0
      */
     tjCompress2(compressor, image.data, image.width, 0, image.height, image.formatProperties.rawValue, &jpegBuf, &jpegSize, 0, outQual, 0)
- 
+
```
compnerd, line 140: Bleeding white space

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -66,52 +66,47 @@ public struct ImageData {
...
+         */
+        tjDecompress2(decompressor, finPointer, UInt(jpegSize), buffer, width, 0, height, imageFormat.rawValue, 0)
+        
+        return ImageData(height: height, width: width, data: buffer!, imageFormat: imageFormat)
+    }
     
 }
  
-func SaveJPEG(atPath path : String, image: ImageData) -> Int32 {
+func SaveJPEG(atPath path : String, image: ImageData) throws -> Int32 {
```
compnerd, line 109: Extra white space after `path`

CMakeLists.txt
```diff
@@ -79,13 +79,39 @@ ExternalProject_Add(swift-protobuf
...
+    -D CMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+    -D ENABLE_SHARED=NO
+  INSTALL_COMMAND
+    ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_SOURCE_DIR}/ImageLoader/libjpeg-turbo/module.modulemap ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+  BUILD_BYPRODUCTS
+    <BINARY_DIR>/${CMAKE_STATIC_LIBRARY_PREFIX}turbojpeg${CMAKE_STATIC_LIBRARY_SUFFIX}
+  STEP_TARGETS
+    build)
+
+add_library(turbojpeg STATIC IMPORTED)
```
BradLarson, line 99: Presubmits are currently failing with

```
/swift-models/ImageLoader/JPEGImageOperations/JPEGImageOperations.swift:2:8: error: no such module 'TurboJPEG'
import TurboJPEG
       ^
```

Could that be because the name here is `turbojpeg` instead of `TurboJPEG`? Not as familiar with CMake, but the other Swift module names use the same capitalization as when they are imported in code.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,154 @@
...
+    
+    guard FileManager.default.fileExists(atPath: path) else {
+        throw ("File does not exist at \(path).")
+    }
+    
+    let data: Data = FileManager.default.contents(atPath: path)!
+    let jpegSize: UInt = UInt(data.count)
+    
+    return data.withUnsafeBytes {
+        let finPointer : UnsafeMutablePointer<UInt8> = UnsafeMutablePointer<UInt8>(mutating: $0.baseAddress!.assumingMemoryBound(to: UInt8.self))
```
BradLarson, line 82: As part of the Google Swift style, we generally let type inference do the work for us in defining types, so you can safely leave off the `: UnsafeMutablePointer<UInt8>` at the left here. The line itself is a little long (the Google Style Guide specifies 100 character max widths), but if you have trouble wrapping it we can run this through swift-format to get things to the right length.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,154 @@
...
+        case .YUV400:
+            return 1
+        }
+    }
+}
+ 
+public struct ImageData {
+    let height: Int32
+    let width: Int32
+    let data: UnsafeMutablePointer<UInt8>
```
BradLarson, line 58: I'm concerned about the use of the `UnsafeMutablePointer` here and how the memory for this will be managed. When will `data` be deallocated, and who will be responsible for doing so? It looks like `tjAlloc()` creates the buffer within `LoadJPEG()`, but it is never deallocated in any of the code below. This could be a recipe for large memory leaks.

Instead, would it make sense to treat ImageData as a class and contain `tjAlloc()` within its initializer, reference its buffer to copy data in, then use `tjDealloc()` in a `deinit` to abstract away memory management from the consumer of this API?

ayushi2103: Yes that makes sense, I had multiple approaches in mind for the interface, one of them was the one you mentioned. And it does makes sense to do better memory management.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: Okay I will swift-format the file.

CMakeLists.txt
ayushi2103: It was working earlier and I guess due to my cache I didn't get the error earlier. I am not able to figure out any difference in the cmakefiles in the working copy and this one. I don't think it's because of the naming. @compnerd if you could help me with this, it would be great.

ayushi2103: @BradLarson @compnerd It's working but with using source directory, as you mentioned that it's better to use staging directory instead of that, I am not able to make it work with that.
https://github.com/tensorflow/swift-models/blob/5674cadfc7acfe0df6ac400611f275e5647fe1e4/CMakeLists.txt#L93

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,155 @@
...
+
+    let bufferPointer = UnsafeMutableBufferPointer(start: jpegBuf, count: Int(jpegSize))
+    let jpegData = Data(buffer: bufferPointer)
+
+    guard FileManager.default.createFile(atPath: path, contents: jpegData) else {
+        retVal = -1
+        throw ("Error during image saving")
+    }
+
+    return retVal
```
BradLarson, line 150: We can probably drop the return value (and retVal), the throwable error should serve the same purpose.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,155 @@
...
+        formatProperties = imageFormat
+        buffer = tjAlloc(imageFormat.channelCount * width * height)
+    }
+
+    deinit {
+        tjFree(self.buffer)
+    }
+}
+
+func LoadJPEG(atPath path: String, imageFormat: pixelFormats) throws -> ImageData? {
```
BradLarson, line 73: Per [the Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#general-conventions), functions should start with a lowercase letter. Uppercase initial letters are only used for names of types and protocols. This comment can also apply to the `SaveJPEG()` function below.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: I have made the changes.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,151 @@
...
+    import Darwin
+#elseif os(Android) || os(Linux)
+    import Glibc
+#elseif os(Windows)
+    import ucrt
+#else
+    #error("C library not known for the target OS")
+#endif
+
+public enum pixelFormats: Int32 {
```
compnerd, line 14: This should be `PixelFormat` not `pixelFormats`.  Enumerations represent a single entity and should not be pluralized.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,151 @@
...
+        formatProperties = imageFormat
+        buffer = tjAlloc(imageFormat.channelCount * width * height)
+    }
+
+    deinit {
+        tjFree(self.buffer)
+    }
+}
+
+func loadJPEG(atPath path: String, imageFormat: pixelFormats) throws -> ImageData? {
```
compnerd, line 73: Hmm, can this not be:

```
func loadJPEG(atPath path: String, pixelFormat: PixelFormat) throws -> ImageData
```

The function currently either returns an `ImageData` or will throw.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: Okay I will change that.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: Yes, thats better.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,151 @@
...
+    import Darwin
+#elseif os(Android) || os(Linux)
+    import Glibc
+#elseif os(Windows)
+    import ucrt
+#else
+    #error("C library not known for the target OS")
+#endif
+
+public enum pixelFormat: Int32 {
```
compnerd, line 14:
```suggestion
public enum PixelFormat: Int {
```

In general, with Swift, you should use `Int` unless you explicitly need it to be 32-bit in width for some reason.  Enumerated types should use PascalCase.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,151 @@
...
+    case BGRA8888 = 3 // TJPF_BGRA
+    case ARGB8888 = 4 // TJPF_ARGB
+    case ABGR8888 = 5 // TJPF_ABGR
+    case RGBA8880 = 6 // TJPF_RGBX
+    case BGRA8880 = 7 // TJPF_BGRX
+    case ARGB0888 = 8 // TJPF_XRGB
+    case ABGR0888 = 9 // TJPF_XBGR
+    case YUV400 = 10 // TJPF_GREY
+
+    var channelCount: Int32 {
```
compnerd, line 27:
```suggestion
    var channelCount: Int {
```

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,151 @@
...
+            return 3
+        case .YUV400:
+            return 1
+        }
+    }
+}
+
+public class ImageData {
+    let height: Int32
+    let width: Int32
```
compnerd, line 57:
```suggestion
    let height: CInt
    let width: CInt
```

I assume that these were `Int32` because they come from the C API?  If so, it is probably better to sugar this as `CInt`.  Otherwise, prefer `Int`.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    deinit {
+        tjFree(self.buffer)
+    }
+}
+
+func loadJPEG(atPath path: String, imageFormat: pixelFormat) throws -> ImageData {
+    var width: Int32 = 0
+    var height: Int32 = 0
```
compnerd, line 75:
```suggestion
    var width: CInt = 0
    var height: CInt = 0
```

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,151 @@
...
+    do {
+        try FileManager.default.removeItem(atPath: path)
+    } catch {
+        // File not present
+    }
+
+    var jpegBuf: UnsafeMutablePointer<UInt8>?
+    defer { tjFree(jpegBuf) }
+
+    let outQual: Int32 = 95
```
compnerd, line 121:
```suggestion
    let outQual: CInt = 95
```

This seems like an inout parameter, what do you think of renaming it to `quality`?

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
```diff
@@ -0,0 +1,151 @@
...
+    var jpegSize: UInt = 0
```
compnerd, line 122:
```suggestion
    var jpegSize: CUInt = 0
```

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
ayushi2103: I renamed it.

ImageLoader/JPEGImageOperations/JPEGImageOperations.swift
Ayush517: @compnerd, [`tjCompress2`](https://github.com/tensorflow/swift-models/blob/3af7ef2893fdb3d6e97cbd4fd47e95693557975e/ImageLoader/JPEGImageOperations/JPEGImageOperations.swift#L139) needs the pointer of `jpegSize`to be of type `UnsafeMutablePointer<UInt>` , whereas `CUnsignedInt` essentially is just `UInt32`. Hence, maybe here jpegSize should remain as UInt.

PR #656, opened 2020-08-10
[Models/ImageClassification] For transfer learning, apply @_Freezable property wrapper on all Layer conforming properties

rahulbhalley:
This PR proposes to apply `@_Freezable` property wrapper on every property of image classification models that conforms to `Layer` protocol. 

One thing I'd like to point out is that in some models like ResNet, VGG, EfficientNet, and others, the property wrapper is applied on the neural blocks instead of the layers contained in those neural blocks because, according to my best knowledge (please correct me if I'm wrong here), it is rare to freeze specific layers in neural blocks during transfer learning. 

cc @BradLarson

### DIFF
--- a/Models/ImageClassification/DenseNet121.swift
+++ b/Models/ImageClassification/DenseNet121.swift
@@ -20,7 +20,7 @@ import TensorFlow
 // https://arxiv.org/pdf/1608.06993.pdf
 
 public struct DenseNet121: Layer {
-    public var conv = Conv(
+    @_Freezable public var conv = Conv(
         filterSize: 7,
         stride: 2,
         inputFilterCount: 3,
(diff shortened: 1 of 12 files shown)


rahulbhalley:
@googlebot I fixed it.

rahulbhalley:
@googlebot I fixed it.

rahulbhalley:
@googlebot I fixed it.

saeta:
Hi @rahulbhalley Thank you very much for putting this together. Unfortunately, I think we want to hold off on merging this in. Transfer learning is a _really_ important use case to support well, and something we're actively thinking about carefully. @shadaj will be presenting some of our latest ideas at this Friday's open design meeting. I hope you're able to join in!

rahulbhalley:
Ok.

PR #658, opened 2020-08-14
GPT2 Checkpoint Additions

inukshukdeveloper:
This PR proposes to add some functionality to checkpoint writing for GPT2.  One of the layers appeared to have the wrong mapping ("weight" vs "embedding") and another layer in the model was not being written (MultiheadAttention vs MultiheadAttentionGPT2).  Output from the modified checkpoint file differed slightly from the checkpoint stored remotely by a few bytes so it is possible some additional information is being written by this change versus the remotely stored version.

All checkpoint writing tests passed successfully.  Another test could be added to exercise the writing of auxiliary files if desired.

It's known that new protocols are in process (PR #631) for checkpoint writing so this change is more of a historical completion PR for the pattern first established by the GPT2 model.  Hopefully, it will be useful for anyone looking at previous examples of checkpoint writing done in the past.

The work was performed on the 0.10 branch due to some compilation or configuration issues on master.  A fallback to 0.10 allowed the work to proceed but hopefully it can be merged down if the change is acceptable.


Package.swift
```diff
@@ -24,7 +24,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.10.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "0.0.1")),
-        .package(url: "https://github.com/google/swift-benchmark", .branch("master")),
+        .package(url: "https://github.com/google/swift-benchmark", .revision("e2bb5e3e74702b2a764f62007a79bc9feb667ee3")),
```
BradLarson, line 27: This is conflicting with the current head, and I think you should be able to merge in master in order to resolve this. The currently tagged revision in the repository should handle any compilation issues you might have seen.

BradLarson: Sorry, was able to fix this myself.

inukshukdeveloper:
Thanks for the review.  Deleted the branch.  I'll delete the fork shortly.  I believe I have some outstanding work still on it that I need to move to another branch.

PR #662, opened 2020-09-01
Make accuracy and loss values available for upstream use-cases

vkuznet:
Quite often we need accuracy and loss values in upstream code, e.g. to plot them. This PR address this issue.

### DIFF
--- a/TrainingLoop/TrainingProgress.swift
+++ b/TrainingLoop/TrainingProgress.swift
@@ -19,8 +19,8 @@ let progressBarLength = 30
 /// A progress bar that displays to the console as a model trains, and as validation is performed.
 /// It hooks into a TrainingLoop via a callback method.
 public class TrainingProgress {
-  public var accuracies: [Float]
-  public var losses: [Float]
+  public var accuracies: [Float] // accessible list of accuracies values
+  public var losses: [Float]     // acceessible list of loss values
   var statistics: TrainingStatistics?
   let metrics: Set<TrainingMetrics>
   let liveStatistics: Bool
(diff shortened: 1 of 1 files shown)


BradLarson:
Thanks for the suggestion, but I don't think that adding logging functionality within the visual progress bar is the right place for this. That progress bar should only be a visual element, and is only there if you need to have a display of how training is going.

Logging (to CSV, text, TensorBoard, etc.) should happen within other callbacks built for that purpose. That would make it easy for the end user to know what functionality to add. If they just wanted logging, we don't want to have them need to use a visual display if they don't want it.

The idea is that these mechanisms would focus around a central TrainingStatistics callback that would extract the statistics once (to avoid the overhead of repeatedly materializing loss and other tensors) and then provide them to all consumers of stats. The initial architecture I used for the progress bar (it hosting an instance of TrainingStatistics and calling it directly) isn't going to achieve this, so we need to find a way to share a common TrainingStatistics callback from multiple logging or display callbacks while not exposing that implementation detail to the end user.

Again, thanks for suggesting this approach, but I'd like to find a more extensible solution for the problem.

vkuznet:
Thanks for feedback. I agree, I just found that this use-case is not yet covered and I used a a place which had this info. But I'm totally fine to put it in more generic block. Feel free to close/reject PR , but it would be nice to have an open issue on this subject of having access to this info from upstream code.

BradLarson:
I just added issue #663 to track this, hopefully that includes all the relevant information and needs. I'll close this out in favor of whatever we can do to address the needs of that issue.

Again, thanks for bringing this up and reminding us about the need.

PR #666, opened 2020-09-11
Differentiable Shallow Water PDE Solver Example

vojtamolda:
Dear S4TF team,

To follow up on my post on [the mailing list](https://groups.google.com/a/tensorflow.org/g/swift/c/UadyIYNEBoM/m/1vvzHeRcAgAJ) I've turned my [differentiable shallow water PDE solver gist](https://gist.github.com/vojtamolda/bd85033cf62877e6f8ada68b8bbb32a0) into an example. I did some polishing of the source code, added a CLI interface and adapted the readme file.

As usual, naming is difficult ;) I did my best to look around and try to pick a name that's consistent with the other folders, files and code in the repo. Nevertheless, if anyone has a better location or name I'll be happy to amend the commit.

All of the dependencies that were not already needed by something else here (and therefore already listed in `Package.swift`) are gone. I figured it's not worth the hassle to introduce new ones required just by this example.

Besides the above, I have a few points I'd like to discuss:

1. The example contains a simple benchmark of the 3 solver implementations in forward mode. The results are not very informative yet because I had to navigate around a few bugs. So I'm not sure whether you are interested in incorporating them into `BenchmarksCore` target or if it's better to wait.

2. I'm having trouble with CMake builds. I've tried my best to have the example built with this monster, but no luck so far. I'd really appreciate any help with taming the beast. The problematic part is the dependency on [`Benchmark`](https://github.com/google/swift-benchmark). When I comment out the benchmarking functionality in `main.swift, everything works well.
  
   I've noticed that `Benchmark` is also used by parts of the codebase. Interestingly enough none of them can be currently built with CMake. So one idea is that if someone adds CMake support to, for instance, _BenchmarksCore_, I could then cowardly piggyback on this work and copy-paste the relevant parts of `CMakeLists.txt` file.

Cheers,

  Vojta


BradLarson:
I did a quick pass where I added copyright headers and ran the files through swift-format just to get the formatting right. The `--benchmark` command was triggering problems due to the nested argument parsing with the benchmarks and the surrounding application argument parsing, so I changed this to manually use a BenchmarkRunner. I also formatted the benchmark results using milliseconds, which makes them easier to read.

I'm currently waiting on [an upstream fix to swift-benchmark](https://github.com/google/swift-benchmark/pull/77) and then I can add the proper CMake files to get this to build with swift-benchmark. That'll also enable it for other use in the project.

Interestingly, [a recent workaround](https://github.com/tensorflow/swift-apis/pull/1081) for the materialization of extra zeroes during our backwards passes appears to hugely impact your benchmarks. Here are your benchmarks on my GTX 1080-equipped Ubuntu 18.04 machine using the 0.11 release toolchain:

```
name                                        time         std        iterations warmup      
-------------------------------------------------------------------------------------------
Shallow Water PDE Solver.Array Loop         19338.070 ms ±   1.41 %         10 39143.724 ms
Shallow Water PDE Solver.Tensor Slice         452.951 ms ±   1.47 %         10  4386.044 ms
Shallow Water PDE Solver.Tensor Slice (XLA)  1092.033 ms ±   2.31 %         10  2586.686 ms
```

and those  same benchmarks using the latest nightly toolchain:

```
name                                        time        std        iterations warmup     
-----------------------------------------------------------------------------------------
Shallow Water PDE Solver.Array Loop          795.390 ms ±   0.16 %         10 1598.568 ms
Shallow Water PDE Solver.Tensor Slice        453.472 ms ±   1.35 %         10 2139.348 ms
Shallow Water PDE Solver.Tensor Slice (XLA) 1099.160 ms ±   0.91 %         10 2576.604 ms
```

The Array Loop is ~24X faster after that workaround got into the toolchain.

I also was experimenting with simplifying your discrete Laplacian operator as a 3x3 convolution, rather than a series of slices. You can replace the Delta function in TensorSliceSolution.swift with the following:

```swift
    @differentiable
    private func Δ(_ u: Tensor<Float>) -> Tensor<Float> {
        assert(u.shape.allSatisfy{ $0 > 2 })
        assert(u.rank == 2)
        let finiteDifference = depthwiseConv2D(
            u.expandingShape(at: [0, 3]), filter: laplaceKernel, strides: (1, 1, 1, 1), padding: .valid)
        let Δu = finiteDifference / Δx / Δx
        return Δu.squeezingShape(at: [0, 3])
    }
```

where `laplaceKernel` has been defined in your inits as

```swift
        self.laplaceKernel = Tensor<Float>(
            shape: [3, 3, 1, 1], scalars: [0.0, 1.0, 0.0, 1.0, -4.0, 1.0, 0.0, 1.0, 0.0], on: u0.device)
```

That produces the same results as your slicing, and is ~10% faster in eager mode. However, it's 60% slower than the slicing for X10, and I don't know why. It should be a simpler operation. I hesitated on replacing your implementation with that for this reason.

If I can get the CMake build working, I'd be glad to pull this in after that. We can update the image writing with direct animated GIFs once my other PR is in, and see what we can learn from trying to optimize this.

BradLarson:
Thinking about it, the performance difference between the 3x3 convolution and your separate slices does make sense. The discrete Laplacian calculation only requires sampling from and performing math involving the center pixel and four surrounding pixels (up, down, left, right). You do this explicitly in your slicing. In my 3x3 convolution, I'm also sampling from the four corner pixels and just multiplying them by 0. This is 9 pixels per convolution, vs. the 5 pixels needed, an 80% increase.

Thus, if XLA was optimizing your calculation well, a 60% increase in calculation time when switching to a 3x3 convolution would make sense. Our current eager mode implementation is very sensitive to the number of operations dispatched, so it also makes sense that the benefit of compressing this into a single operation could outweigh the increase in calculations performed.

Ideally, a custom kernel that only operated on the center pixels and the four cardinal directions in a single dispatch would be the fastest path here. That might be worth exploring in the future.

vojtamolda:
Thanks for the code review and the benchmarks, Brad.

With the nightly build I reproduced your results on macOS, so everything is good. I thought it would be nice to have a convolution based solution. The readme mentions it anyway so unless you have any objections, I just pushed your convolution based solver to the PR. It's located in `TensorConvSolution.swift`.

Besides that I also noticed that some of the implementation used a different α damping constant. This lead to slightly different results so I fixed that too.

Just out of curiosity I also wrote a `C Loop` solver. Basically a rewrite of the `Array Loop` solver into C. It can serve as a performance baseline. Source code is in [this gist](https://gist.github.com/vojtamolda/bd85033cf62877e6f8ada68b8bbb32a0#file-cloopsolution-c) because I don't think it belongs in this repo.

Here are the results I got on macOS (MacBook Pro 16", 2019, 2.6Ghz Intel i7 CPU) with `DEVELOPMENT-2020-09-16-a` toolchain:
```
name                                        time         std        iterations warmup      
-------------------------------------------------------------------------------------------
Shallow Water PDE Solver.C Loop              355.480 ms ±   4.35 %         10       nan ms
Shallow Water PDE Solver.Array Loop          665.574 ms ±   1.52 %         10  1298.180 ms
Shallow Water PDE Solver.Tensor Slice        536.672 ms ±   4.60 %         10  1201.721 ms
Shallow Water PDE Solver.Tensor Slice (XLA)  640.667 ms ±   1.91 %         10  1470.422 ms
Shallow Water PDE Solver.Tensor Conv        7502.167 ms ±   0.21 %         10 14969.662 ms
Shallow Water PDE Solver.Tensor Conv (XLA)   562.674 ms ±   0.25 %         10  1333.303 ms
```

I think the numbers are very encouraging. S4TF toolchain is about 2x slower than C. Not shabby at all!

BradLarson (approved):
There's still one or two things left to get CMake fully working with swift-benchmark, so I'm not not going to hold this up to wait for that. We can add in CMake support across the project for things that need it once we get that resolved.

Similarly, I'd like to simplify the image writing routines here so that people could run the model in one command from the root of the project (we've tried to make all the other targets work this way), but that can come later. Once I get PR #669 in, that would be a great time to rework the image writing to use the animated GIF writer directly and to write into a shared output directory at the root level. That isn't worth waiting on.

Thanks for assembling and evaluating this simulation, it's a fantastic illustration of the applications of differentiable Swift outside of purely machine learning. There's a lot to explore in this direction.

PR #667, opened 2020-09-15
Revert non-active try-apply differentiation workarounds

efremale:
Since TF-433 was fixed, we can remove all the workarkounds for (non-active) try-apply.
For example, calls to `Array.map(_:)` and `Bool.&&`.


efremale:
cc @dan-zheng

dan-zheng (approved):
Nice!

(Since this PR depends on compiler changes that landed after the 0.11 release, let's consider holding off on merging until the next release is cut to avoid breaking `tensorflow/swift-models` users.)

Issue #663, opened 2020-09-02
Add logging callbacks for training statistics 


BradLarson:
The training loop abstraction has a visual progress bar for tracking statistics during training. In many cases, it's desirable to log or store training statistics for later analysis, or to send those statistics to different services for visualization.

Further discussion is present in PR #662, but we'd like to have callbacks such as the following:

- [ ]  CSV logging (example: [tf.keras.callbacks.CSVLogger](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/CSVLogger))
- [ ]  Custom closure- or function-defined logging (for logging to text files, transmission, etc.)
- [ ]  TensorBoard scalar logging (example: [tf.keras.callbacks.TensorBoard](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/TensorBoard))

as well as a general mechanism for writing more.

A TrainingStatistics callback exists now as a means of collecting statistics, but it is only directly called by TrainingProgress. The idea is to have a single statistics gathering function, so as to avoid the overhead of materializing statistics-related tensors multiple times, and to have that report those statistics to all interested parties. For this, we'd need to have one TrainingStatistics instance that acts as a kind of notification center to other callbacks.

We'd want this implementation detail to be hidden from end-users, so that they just need to specify which logging or visualization callbacks they need, and not have to worry about managing the TrainingStatistics instance itself.

vkuznet:
+1

PR #668, opened 2020-09-15
TrainingLoop: refactor progress printer and add CSVLogger

xihui-wu:
This is the first step addressing of https://github.com/tensorflow/swift-models/issues/663. Highlights of the change:

1. Add StatisticsRecorder and ProgressPrinter as pre-registered callbacks to TrainingLoop. (as [fastai](https://github.com/fastai/fastai/blob/master/fastai/callback/progress.py#L55) does)
2. Add CSVLogger callback.


TrainingLoop/TrainingLoop.swift
```diff
@@ -146,87 +187,130 @@ where
...
-    self.callbacks = callbacks
+    self.metrics = metrics
+    let statisticsRecorder = StatisticsRecorder(liveStatistics: true, metrics: [.loss] + metrics)
+    let progressPrinter = ProgressPrinter(liveStatistics: true)
+    self.statisticsRecorder = statisticsRecorder
+    self.progressPrinter = progressPrinter
+    self.callbacks = [
+      statisticsRecorder.record,
+      progressPrinter.print,
+    ] + callbacks
```
BradLarson, line 290: I can imagine cases where we don't want to record per-batch or per-epoch statistics. By building statistics gathering directly into the training loop, rather than having it be a user-provided component, can we account for those cases where we want to strip down the training loop as much as possible?

xihui-wu: Yes please see latest change where I make both whether to include default callbacks and when to record stats configurable.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+import ModelSupport
+
+/// A callback-based handler for logging the statistics to CSV file.
+public class CSVLogger {
+  public var path: String
+
+  let foundationFS: FoundationFileSystem
+  let foundationFile: FoundationFile
+
+  /// Create an instance that log statistics during the training loop.
```
dabrahams, line 11: “during the training loop” is not true about this initializer, only about how you happen to use it later.  The doc comment should say something about how the parameters are used.

```suggestion
  /// Creates an instance that logs to a file with the given path.
```

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+  public init(withPath path: String = "run/log.csv") {
```
dabrahams, line 12: Almost never use "with" to introduce parameters.  It adds no information for the reader beyond what you'd get from `path`.  Of course you're calling this intializer “with” that path!

```suggestion
  public init(path: String = "run/log.csv") {
```

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+    self.path = path
```
dabrahams, line 13: It seems very unlikely to me that we actually need to store `path` in addition to `foundationFile`.  Consider whether it can/should be dropped because you can get it from `foundationFile`.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+    self.foundationFS = FoundationFileSystem()
```
dabrahams, line 14: Storing a copy of the filesystem seems suspect, just logically speaking.  Are these instances not really all the same thing (singleton)?  Should we just construct the instance on the fly when we need it?

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
+import Foundation
+import ModelSupport
+
+/// A callback-based handler for logging the statistics to CSV file.
+public class CSVLogger {
+  public var path: String
+
+  let foundationFS: FoundationFileSystem
+  let foundationFile: FoundationFile
```
dabrahams, line 9: Doc comment please. Don't name things after their types.

If you rename the thing `CSVLogFile` or something, then:

```suggestion
  let underlyingFile: FoundationFile
```

or

```suggestion
  let rawFile: FoundationFile
```

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+
+  /// Create an instance that log statistics during the training loop.
+  public init(withPath path: String = "run/log.csv") {
+    self.path = path
+    self.foundationFS = FoundationFileSystem()
+    self.foundationFile = FoundationFile(path: path)
+  }
+
+  /// The callback used to hook into the TrainingLoop for logging statistics.
```
dabrahams, line 18: This is a function, so you say what it *does*, per 

> Describe what a function or method does and what it returns, omitting null effects and Void returns

Please read https://swift.org/documentation/api-design-guidelines/#fundamentals

```suggestion
  /// Logs the end of an iteration of `loop`; ignoring other events.
```

or something.

Also, please describe the conditions under which it might throw.  It is probably enough to say that filesystem errors are reported by throwing.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+  ///
+  /// - Parameters:
```
dabrahams, line 20: You almost certainly don't need a parameters block because you can describe the parameters in the summary.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+  ///   - loop: The TrainingLoop where an event has occurred. 
+  ///   - event: The training or validation event that this callback is responding to.
+  public func log<L: TrainingLoopProtocol>(_ loop: inout L, event: TrainingLoopEvent) throws {
```
dabrahams, line 23: Both parameters are named after their types.  You might not be able to find a better name for `event` (but you might! “`occurrence`” probably isn't an improvement).  “`loop`” surely could be `session` or something… though maybe we should be using `Session` as part of the type and protocol names.  Anyway, it's worth doing some work on this.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+    switch event {
+    case .batchEnd:
+      guard let epochIndex = loop.epochIndex, let epochCount = loop.epochCount,
+        let batchIndex = loop.batchIndex, let batchCount = loop.batchCount
+      else {
+        return
```
dabrahams, line 29: Why would any of these things be `nil`, and why are we bailing out when they're `nil`?  That should be explained in a comment.  Ditto for `stats` below.  Also, why are these two separate guard statements?

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+      }
+
+      guard let stats = loop.lastStatsLog else {
+        return
+      }
+
+      if !FileManager.default.fileExists(atPath: path) {
```
dabrahams, line 36: I would do this upon creation of the logger, holding the file open for write, rather than checking the filesystem each time anything is logged.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+        try foundationFS.createDirectoryIfMissing(at: String(path[..<path.lastIndex(of: "/")!]))
```
dabrahams, line 37: It took me a while to figure out what was happening here, but IIUC this will do the wrong thing if the path you were passed happens to end with "/".

dabrahams, line 37: The fact that you append to a log file that already exists rather than overwriting it needs to be in a doc comment somewhere.

dabrahams, line 37: The fact that you try to create the containing directory if it is missing needs to be in a doc comment somewhere.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+        try writeHeader(stats: stats)
+      }
+      try writeDataRow(
```
dabrahams, line 40: Could probably be 1 line?

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+      return
+    }
+  }
+
+  func writeHeader(stats: [(String, Float)]) throws {
+    let head: String = (["epoch", "batch"] + stats.map { $0.0 }).joined(separator: ", ")
+    do {
+      try head.write(toFile: path, atomically: true, encoding: .utf8)
+    } catch {
+      print("Unexpected error in writing header line: \(error).")
```
dabrahams, line 54: This is the worst of both worlds: while forcing clients to deal with a thrown error, it gives them no choice about whether the error is reported.  Either:

1. consider the error thrown by `write` to be informative enough and simply skip the `do… catch`, or
2. Create your own error (`CSVLogger.HeaderWriteError`) that wraps the thrown error, and throw that instead from the Catch block.

my instinct says #1 is the answer.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+        epoch: "\(epochIndex + 1)/\(epochCount)",
+        batch: "\(batchIndex + 1)/\(batchCount)",
+        stats: stats)
+    default:
+      return
+    }
+  }
+
+  func writeHeader(stats: [(String, Float)]) throws {
+    let head: String = (["epoch", "batch"] + stats.map { $0.0 }).joined(separator: ", ")
```
dabrahams, line 50: `.0` is indexing a tuple, but I assume it's a statistic name.  It would be better if the fields of the tuple were labeled and you could use the label (`$0.name`).  If you have control over that, consider making the change.

dabrahams, line 50:
```suggestion
    let head: String = (["epoch", "batch"] + stats.lazy.map { $0.0 }).joined(separator: ", ")
```

dabrahams, line 50: This is another example where the let binding "`head`" is providing dubious value.  The name is not great if you're going to use a let binding; it isn't “a head.” Otherwise,
```suggestion
    (["epoch", "batch"] + stats.lazy.map { $0.0 }).joined(separator: ", ")
      .write(toFile: path, atomically: true, encoding: .utf8)
```

TrainingLoop/Callbacks/CSVLogger.swift
dabrahams: Same remarks apply to `writeDataRow`.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+    do {
+      try head.write(toFile: path, atomically: true, encoding: .utf8)
+    } catch {
+      print("Unexpected error in writing header line: \(error).")
+      throw error
+    }
+  }
+
+  func writeDataRow(epoch: String, batch: String, stats: [(String, Float)]) throws {
+    let dataRow: Data = (
```
dabrahams, line 60: Why are we creating `Data` from a `String` here?  Surely `foundationFile` can append a string?

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+      "\n" + ([epoch, batch] + stats.map { String($0.1) }).joined(separator: ", ")
```
dabrahams, line 61: It's traditional good form that files end with `"\n"`, so I would write this at the end of the header and at the end of each line you write.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+    ).data(using: .utf8)!
+    do {
+      try foundationFile.append(dataRow)
```
dabrahams, line 64: Isn't it important to do this atomically if there are multiple writers to the same log file?

dabrahams (changes requested):
Reviewed the first file.  Please revise (and apply anything you learned from doing so to the other files) and I can look again.  Thanks!

Examples/LeNet-MNIST/main.swift
```diff
@@ -31,24 +31,33 @@ let dataset = MNIST(batchSize: batchSize, on: device)
...
   training: dataset.training,
   validation: dataset.validation,
   optimizer: optimizer,
   lossFunction: softmaxCrossEntropy,
-  callbacks: [trainingProgress.update])
+  metrics: [.accuracy],
+  callbacks: [CSVLogger().log])
+
+// Compute statistics only when last batch ends.
+(trainingLoop.statisticsRecorder!).shouldCompute = {
```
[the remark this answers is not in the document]
xihui-wu: Done.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
+import Foundation
+import ModelSupport
+
+/// A callback-based handler for logging the statistics to CSV file.
```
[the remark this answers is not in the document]
xihui-wu: Updated the doc comment. Any reason why this is LogFile? (Logger makes sense to me since its a logger not a file)

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+public class CSVLogger {
+  public var path: String
```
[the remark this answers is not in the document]
xihui-wu: Done

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,70 @@
...
+
+  let foundationFS: FoundationFileSystem
```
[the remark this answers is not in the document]
xihui-wu: Removed the FoundationFileSystem instance and create on the fly.

Support/FoundationFileSystem.swift
```diff
@@ -58,4 +58,11 @@ public struct FoundationFile: File {
     // TODO: Incorporate file offset.
     try value.write(to: location)
   }
+
+  public func append(_ value: Data) throws {
```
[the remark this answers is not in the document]
xihui-wu: Done.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Done.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Same. Removed foundationFile and kept the path.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Removed foundationFile.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Good to know. Done.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Merged stats into the same guard statements.
These are designed as optionals in existing TrainingLoop. I think that's because it's not the must-have values to complete a training process. I added an inline comment.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Oh yes I omitted this case.

xihui-wu: done.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: got it.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Good suggestion. Done.

xihui-wu: Does lazy have much impact in the performance in this case ?

xihui-wu: How about header ?

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Not it can't. It appears to me that we have designed the FoundationFile to handle operations on Data only.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Got it.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Done.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Yeah good to create on the fly

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Done.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: I assume "trainingLoopEvent" is named after their types but "event" is not. Is there a strict rule for this? 
To me "event" is much more straightforward than "occurrence" tho.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Got it.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: I'd rather to overwrite it actually. Added comment inside the new init.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: This is outdated. Moved createDirectoryIfMissing to init.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Does FileHandle support atomically write ? (Also, we won't expect multiple files written at same time I think)

TrainingLoop/Callbacks/CSVLogger.swift
dabrahams: There's no strict rule.

Consider:

```
let arrayInt = students.map(\.grade) // definitely named after the type, I'm sure you agree
let array = students.map(\.grade)    // still named after type, just less specific
let grades = students.map(\.grade)   // named after role/meaning
```

`event` is like `array` above.  I agree "event" is better than "occurrence," (not more than a long synonym for “event”) which is why I said “occurrence” is probably not an improvement.   The type and name of a parameter name both make a statement about that parameter. If you just use words from the type in the parameter name one often gives up an opportunity to make a more meaningful statement with the parameter name.  I'm not saying that's the case here with `event`; sometimes there is no way to do better than repeating type information.  I'm saying, **notice** that the name contains only type information and  **take a few moments** to see if you can do better.  That's all.  If you don't come up with anything, that's fine; just move on.

TrainingLoop/Callbacks/CSVLogger.swift
dabrahams: OK, first, I would rather see something like 

```
// These properties will be `nil` unless stats logging was requested
```
because it describes the situation at a semantic level rather than at the level of what some code did.

(also writing "No-Op" adds nothing to what is already very obvious from the code)

But that said, it seems very unlikely that the comment I'd like to see is true of any but the last property.  All the others refer to values that have nothing to do with logging.  So I want to know what causes `epochIndex` to be `nil`, for example.

It's a big design flaw in trainingLoop that it has so many optionals, and that makes this task more difficult, but I believe that's not the code you're working on(?)

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,68 @@
+import Foundation
+import ModelSupport
+
+/// A handler for logging training and validation statistics to a CSV file.
+public class CSVLogger {
+  /// The path of the file that statistics are logged to.
+  public var path: String
+
+  // The boolean variable indicating if header of the CSV file has been written or not.
```
dabrahams, line 9:
```suggestion
  // True iff the header of the CSV file has been written.
```

Everyone can see it's a `Bool`.  The phrasing “indicating if … or not” is not clear about the polarity of the `Bool`.

TrainingLoop/Callbacks/CSVLogger.swift
dabrahams: Hmm, I'm not sure you want to open and close the file for every line logged, though.  (I am presuming that  the `foundationFile` object keeps the file open)

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,68 @@
...
+
+  /// Creates an instance that logs to a file with the given path.
+  ///
+  /// Throws: File system errors.
+  public init(path: String = "run/log.csv") throws {
+    self.path = path
+
+    // Create the containing directory if it is missing.
+    let containingDir = String(path[..<path.lastIndex(of: "/")!])
+    if containingDir != "" {
```
dabrahams, line 20: why would this string ever be empty?  I don't think it would.  Why add a check for it?

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,68 @@
...
+    case .batchEnd:
+      guard let epochIndex = loop.epochIndex, let epochCount = loop.epochCount,
+        let batchIndex = loop.batchIndex, let batchCount = loop.batchCount,
+        let stats = loop.lastStatsLog
+      else {
+        // No-Op if trainingLoop doesn't set the required values for stats logging.
+        return
+      }
+
+      if !headerWritten {
```
dabrahams, line 44: Why track this value rather than simply writing the header when the `Logger` is initialized?

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,68 @@
...
+        epoch: "\(epochIndex + 1)/\(epochCount)",
+        batch: "\(batchIndex + 1)/\(batchCount)",
+        stats: stats)
+    default:
+      return
+    }
+  }
+
+  func writeHeader(stats: [(name: String, value: Float)]) throws {
+    let header: String = (["epoch", "batch"] + stats.map { $0.name }).joined(separator: ", ") + "\n"
```
dabrahams, line 59:
```suggestion
    let header: String = (["epoch", "batch"] + stats.lazy.map { $0.name }).joined(separator: ", ") + "\n"
```

Didn't I already make this suggestion?

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: If the path is sth like "/log.csv" then containingDir is empty.
Thought twice and I just changed to use URL to handle the parsing.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: I need to figure out what header to write from the stats names.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Got it!

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Got it. The TrainingLoopCallback was designed with these parameters https://github.com/tensorflow/swift-models/blob/master/TrainingLoop/TrainingLoop.swift#L129 and I took what already there. 

I could understand from your example that when the type is very generic, naming after role/meaning is important, but when the type is relatively specific then it's harder to pick a better meaningful name. Yeah if you can't think of a better name, let me keep "event".

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: These variables were originally designed to be optionals to store temporary data in protocol: https://github.com/tensorflow/swift-models/blob/master/TrainingLoop/TrainingLoop.swift#L76

The generic TrainingLoop that implements the protocol does set all these optionals. My guess on why it was designed so is that it allows other TrainingLoops not setting them.

So the point on the comment is NOT "if stats logging doesn't request it then these properties will be nil", but "if these properties are nil No-op on the CSVLogger".

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Got it.

TrainingLoop/Metrics.swift
```diff
@@ -0,0 +1,100 @@
...
+    switch self {
+    case .loss:
+      return LossMeasurer(self.name)
+    case .accuracy:
+      return AccuracyMeasurer(self.name)
+    }
+  }
+}
+
+/// A protocal defining functionalities of a metrics measurer.
```
dabrahams, line 27: That doc comment doesn't tell me anything that isn't obvious from the following line of code.  Never start a protocol's doc summary with "A protocol".  I have to admit that I'm still searching for an non-awkward way to describe protocols; recently I have settled on “Types that…” because a protocol represents a category of types, e.g., "Types that accumulate statistics from a series of `Tensor<Float>`-valued samples,”  (or something) but you could also just say ”An accumulator of statistics.”

Is this, conceptually, like what's described here?  https://www.boost.org/doc/libs/1_74_0/doc/html/accumulators.html
If so, I suggest borrowing naming and documentation language choices from that library (I was one of the library's designers).

TrainingLoop/Metrics.swift
```diff
@@ -0,0 +1,100 @@
...
+public protocol MetricsMeasurer {
+  var name: String { get set }
```
dabrahams, line 29: doc comments are required on all methods and properties.

TrainingLoop/Metrics.swift
```diff
@@ -0,0 +1,100 @@
...
+  mutating func reset()
+  mutating func accumulate<Output, Target>(
+    loss: Tensor<Float>?, predictions: Output?, labels: Target?
+  )
+  func measure() -> Float
+}
+
+/// A measurer for measuring loss.
+public struct LossMeasurer: MetricsMeasurer {
+  public var name: String
```
dabrahams, line 39: Doc comments are required on all methods and properties of this type.

TrainingLoop/Metrics.swift
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  public func measure() -> Float {
+    return totalBatchLoss / Float(batchCount)
+  }
+}
+
+/// A measurer for measuring accuracy
+public struct AccuracyMeasurer: MetricsMeasurer {
+  public var name: String
```
dabrahams, line 69: Doc comments are required on all methods and properties of this type.

TrainingLoop/Metrics.swift
```diff
@@ -0,0 +1,100 @@
...
+
+  public mutating func reset() {
+    correctGuessCount = 0
+    totalGuessCount = 0
+  }
+
+  public mutating func accumulate<Output, Target>(
+    loss: Tensor<Float>?, predictions: Output?, labels: Target?
+  ) {
+    guard let predictions = predictions as? Tensor<Float>, let labels = labels as? Tensor<Int32>
```
dabrahams, line 86: This is an unfortunately late point at which to detect the type mismatch.  Ideally it would be done at compile-time.

TrainingLoop/TrainingLoop.swift
```diff
@@ -18,7 +18,7 @@ import TensorFlow
 // Workaround https://bugs.swift.org/browse/TF-1122 that prevents us from registering a
 // loss function inside our TrainingLoop struct
 public final class LossFunctionWrapper<Output: Differentiable, Target> {
-  public typealias F = @differentiable (Output, @noDerivative Target) -> Tensor<Float>
+  public typealias F = @differentiable(Output, @noDerivative Target) -> Tensor<Float>
```
dabrahams, line 21:
```suggestion
  public typealias F = @differentiable (Output, @noDerivative Target) -> Tensor<Float>
```

I don't _think_ we'd normally drop that space.  Regardless, it's a good idea to avoid making needless whitespace changes on lines you don't otherwise have to edit.  If you want to do a whitespace cleanup commit, it should usually be a separate PR.

TrainingLoop/TrainingLoop.swift
```diff
@@ -34,64 +34,94 @@ public protocol TrainingLoopProtocol {
...
   var validation: Validation { get }
 
   // Optimizer and loss function
   /// The optimizer.
   var optimizer: Opt { get set }
+
   /// The loss function.
   var lossFunction: LossFunction { get set }
 
+  /// The metrics
```
dabrahams, line 81: Do better, e.g.,

```suggestion
  /// The statistics being tracked.
```

TrainingLoop/TrainingLoop.swift
```diff
@@ -34,64 +34,94 @@ public protocol TrainingLoopProtocol {
...
   /// The callbacks used to customize the training loop.
   var callbacks: [TrainingLoopCallback<Self>] { get set }
 
   // Temporary data
+
+  // MARK: - Step-level data
+
   /// The last input fed to the model.
-  var lastInput: Input? { get set }
+  var lastStepInput: Input? { get set }
```
dabrahams, line 93: When you see repetition like this, the instinct should be to replace it with something like:

```swift
var lastStep: Step?
```

TrainingLoop/TrainingLoop.swift
```diff
@@ -34,64 +34,94 @@ public protocol TrainingLoopProtocol {
...
   /// The number of batches in the current collection of batches.
   var batchCount: Int? { get set }
+
   /// The index of the current batch.
   var batchIndex: Int? { get set }
+
+  // MARK: - Epoch-level data
+
+  /// The number of epochs we are currently fitting for.
+  var epochCount: Int? { get set }
```
dabrahams, line 116: Why are these optional?  Usually, 2-phase initialization indicates a design error.

TrainingLoop/TrainingLoop.swift
```diff
@@ -34,64 +34,94 @@ public protocol TrainingLoopProtocol {
...
+
+  /// The index of the current epoch.
+  var epochIndex: Int? { get set }
+
+  // MARK: - Others
+
+  /// The log for last statistics
```
dabrahams, line 123: last what?  We have last epoch, last step, …

TrainingLoop/TrainingLoop.swift
```diff
@@ -146,87 +187,139 @@ where
...
   /// The optimizer.
   public var optimizer: Opt
+
   /// The loss function
   public var lossFunction: LossFunction
 
-  // Callbacks
-  /// The callbacks used to customize the training loop.
-  public var callbacks: [TrainingLoopCallback<Self>] = []
+  /// The metrics
```
dabrahams, line 218: Do better please

TrainingLoop/TrainingLoop.swift
```diff
@@ -146,87 +187,139 @@ where
...
+  public var metrics: [TrainingMetrics]
+
+  /// Callbacks
+
+  // MARK: - The callbacks used to customize the training loop.
+
+  public var callbacks: [TrainingLoopCallback<Self>]
```
dabrahams, line 225: Doc comments missing.

dabrahams (changes requested):

TrainingLoop/Callbacks/CSVLogger.swift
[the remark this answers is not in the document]
dabrahams: This thing can only be constructed to log to a path in the filesystem, and when we invoke its one public method, `log`, it appends to that file.  Properly used, if you have multiple instances, each instance logs to a different file.  The word `Logger` doesn't imply any of those things.  A more general `Logger` might be an interesting abstraction, but it would have a different API.

If I see this code, I know exactly what's happening.  
```swift
// No argument label needed, arguably, because when you construct a thing called “File” with
// a string, the string is obviously a path.
LogFile log(s)

log.append(blah, blah, blah)
```
With `Logger` and `log`, it's less clear.

Support/FoundationFileSystem.swift
```diff
@@ -58,4 +58,14 @@ public struct FoundationFile: File {
     // TODO: Incorporate file offset.
     try value.write(to: location)
   }
+
+  /// Append data to the file.
```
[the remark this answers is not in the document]
xihui-wu: Got it.

Support/FoundationFileSystem.swift
```diff
@@ -58,4 +58,14 @@ public struct FoundationFile: File {
...
+  ///
+  /// Parameter value: data to be appended at the end.
```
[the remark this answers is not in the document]
xihui-wu: Got it

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import ModelSupport
+
+public enum CSVLoggerError: Error {
+  case InvalidPath
+}
+
+/// A handler for logging training and validation statistics to a CSV file.
+public class CSVLogger {
+  /// The path of the file that statistics are logged to.
```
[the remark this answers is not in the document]
xihui-wu: Done

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Thanks for the clarification.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: Chatted offline and resolved

TrainingLoop/Callbacks/CSVLogger.swift
dabrahams: Bleah!

dabrahams (approved):
Approving to unblock progress but there are lots of things to fix, including many that are simply mechanical (missing doc comments).

xihui-wu:
> Approving to unblock progress but there are lots of things to fix, including many that are simply mechanical (missing doc comments).

Thanks! Will come with new PRs to address:
1. @dabrahams's open comments
2. An easier interface for specifying when to compute stats (as per @BradLarson )
3. Potential changes in metrics / loss as I have new findings when converting GPT2 to TrainingLoop. (this is also what I'm blocked on and wanted to make iteration on top of this PR change)

PR #669, opened 2020-09-16
Growing Neural Cellular Automata

BradLarson:
This adds an implementation of the model from ["Growing Neural Cellular Automata"](https://distill.pub/2020/growing-ca/), as well as an accompanying notebook. Experiments 1, 2, and 3 from that publication have been implemented in the command-line example. Experiment 4 will be added later, once PR #631 has been merged to add necessary checkpointing and restoration functionality.

Both eager-mode and X10 backends are supported, and most hyperparameters can be set via command-line options.

In addition to the neural cellular automata model, this makes some more significant modifications to the image loading and saving APIs within swift-models (as needed by this model):

- Support for loading and saving PNGs with an alpha channel has been added.
- The loading and saving interfaces have been changed to reflect the support for both PNG and JPG images.
- A new `AnimatedImage` type has been added, which allows for direct writing of animated GIFs from arrays of Tensors. This uses a native-Swift GIF writer in the `GIF` type, and should work across platforms and in Colab. The animated GIF writer uses a default quantized colorspace, and better results could be obtained in the future by using quantization that adapts to the image frames.


Examples/GrowingNeuralCellularAutomata/CellRule.swift
```diff
@@ -0,0 +1,131 @@
...
+@inlinable
+@derivative(of: clipBackwardsTrace)
+func _vjpClipBackwardsTrace(
+  _ input: Tensor<Float>
+) -> (value: Tensor<Float>, pullback: (Tensor<Float>) -> Tensor<Float>) {
+  return (
+    input,
+    {
+      LazyTensorBarrier()
+      return $0
```
brettkoonce, line 128: 🕳

marcrasi (changes requested):
Useful additions to the image functionality! I had some comments about the APIs.

BradLarson:
@marcrasi - I believe that I've addressed all of your (really helpful) comments about the image APIs. I now determine colorspace based on the size of the number of color channels present, and I've extracted all the special-case handling we had for various image cases into composable methods. For image saving, it seemed cleanest to do these as extensions on Tensor, because that lead to simple code at the use site.

I've added documentation to both Image and AnimatedImage to try to write down all the assumptions that were implicit before, and added preconditions where appropriate.

In the process of testing all these changes across the models, I found existing problems with three of them, so it was nice to uncover those and fix them as part of this.

8bitmp3:
Can't wait @BradLarson That was an awesome demo 🖼

marcrasi (approved):

8bitmp3:
Awesome job @BradLarson and thanks team @saeta @ematejska et al. Just found out via an appendix in a research [paper](https://arxiv.org/pdf/1912.01588.pdf) that cellular automata are used in the Procgen Benchmark environments, like [CaveFlyer](https://github.com/openai/procgen), to generate game level layouts (OpenAI, 2019). E.g. "Procedural generation controls the level layout via cellular automata..."

image

PR #670, opened 2020-09-24
Address doc comments in TrainingLoop Callbacks

xihui-wu:
This change addressed the doc comments per requested in https://github.com/tensorflow/swift-models/pull/668.

Two questions from there remains for further discussion:
1.  Use CSVLogger or CSVLog.
2.  ProgressPrinter does not need to exist as a type but: 
     extension TrainingLoopProtocol {
         public mutating func printProgress(event: TrainingLoopEvent) throws { ... }
     }


Support/FoundationFileSystem.swift
```diff
@@ -59,9 +59,7 @@ public struct FoundationFile: File {
     try value.write(to: location)
   }
 
-  /// Append data to the file.
-  ///
-  /// Parameter value: data to be appended at the end.
+  /// Appends the bytes in 'value' to the file.
```
dabrahams, line 62: backtix

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -53,21 +52,27 @@ public class CSVLogger {
       }
 
       try writeDataRow(
-        epoch: "\(epochIndex + 1)/\(epochCount)",
-        batch: "\(batchIndex + 1)/\(batchCount)",
+        epochData: "\(epochIndex + 1)/\(epochCount)",
```
dabrahams, line 55: Why should we consider this an improvement?  _Everything_ is "data" in some sense.

Support/FoundationFileSystem.swift
xihui-wu: oh my neglect

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: I changed It for better doc comment for this func. But yes it doesn't make it better. I changed it back.

Support/FoundationFileSystem.swift
```diff
@@ -59,9 +59,7 @@ public struct FoundationFile: File {
     try value.write(to: location)
   }
 
-  /// Append data to the file.
-  ///
-  /// Parameter value: data to be appended at the end.
+  /// Appends the bytes in `value` to the file.
   public func append(_ value: Data) throws {
```
dabrahams, line 63:
```suggestion
  /// Appends the bytes in `suffix` to the file.
  public func append(_ suffix: Data) throws {
```

This came from me thinking, “there's always a better name than `value`, because that word applies to every parameter.”

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -7,13 +7,13 @@ public enum CSVLoggerError: Error {
 
 /// A handler for logging training and validation statistics to a CSV file.
```
dabrahams, line 8: I still don't know what a handler is.  Still don't think this is a strong abstraction.

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -32,7 +32,7 @@ public class CSVLogger {
     self.headerWritten = false
   }
 
-  /// Logs the statistics for the 'loop' when 'batchEnd' event happens; 
+  /// Logs the statistics for the `loop` when `batchEnd` event happens;
```
dabrahams, line 35:
```suggestion
  /// Logs the statistics for `loop` when a `batchEnd` event happens; 
```

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -61,11 +60,17 @@ public class CSVLogger {
     }
   }
 
+  /// Writes column names to file header.
```
dabrahams, line 63:
```suggestion
  /// Writes a row of column names to the file.
```

TrainingLoop/Callbacks/CSVLogger.swift
```diff
@@ -61,11 +60,17 @@ public class CSVLogger {
...
+  /// 
+  /// Column names include "epoch", "batch" and `stats` `name`s.
```
dabrahams, line 65:
```suggestion
  /// Column names are "epoch", "batch" and the `name` of each element of `stats`, in that order.
```

dabrahams: Actually the ordering of dictionary elements, for a given dictionary value (its contents) is not deterministic and if this component isn't used in exactly the way you've tested it, it will be printing the wrong stats under a given column, or printing a column that contains statistic A in one row but statistic B in a second row.

I would accept an array of statistic names in the `init()` of this component, which determines the column order.

dabrahams (commented):
Some feedback; GTG now but can give more later.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: The order will preserve. I made `stats` to be an array of tuples instead of dictionary.

TrainingLoop/Callbacks/CSVLogger.swift
xihui-wu: It's an observer callback where TrainingLoop is designed upon. https://docs.google.com/document/d/1CtVFhV8OcQ4E7CmNyfFeZu0IgnUPx86tfQGXiHJUXz0/edit?ts=5ebef977#heading=h.b2so9ayrnyyp

You proposed to make it a function in TrainingLoop. Here are some points I'm more in favor of making callbacks wrapped in a separate classes:
1. Decouple from TrainingLoop
2. Use stored properties to share callback settings
3. Follow this pattern for all callbacks

Let's discuss more offline or in seminar meeting !?!

dabrahams (approved):
LGTM!

PR #671, opened 2020-09-25
Convert GPT2-WikiText2 to TrainingLoop

xihui-wu:
Also make the training on XLA by default.

### DIFF
--- a/Examples/GPT2-WikiText2/main.swift
+++ b/Examples/GPT2-WikiText2/main.swift
@@ -17,6 +17,9 @@ import TensorFlow
 import TextModels
 import TrainingLoop
 
+// Avoid the eager mode runtime from taking all memory 
+// and leaving none to X10 when run on the GPU.
+_ = _ExecutionContext.global
 // Until https://github.com/tensorflow/swift-apis/issues/993 is fixed, default to the eager-mode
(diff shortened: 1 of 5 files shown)


Examples/GPT2-WikiText2/main.swift
```diff
@@ -15,73 +15,46 @@
...
 import TensorFlow
 import TextModels
+import TrainingLoop
+
+// Until https://github.com/tensorflow/swift-apis/issues/993 is fixed, default to the eager-mode
+// device on macOS instead of X10.
+#if os(macOS)
+  let device = Device.defaultTFEager
+#else
+  let device = Device.defaultXLA
```
BradLarson, line 25: For this model, you may need to add the line `_ = _ExecutionContext.global` before this line in order to avoid the eager mode runtime from taking all memory when run on the GPU. Without that line, this model runs out of memory on the GPUs I tried. Nominally, we'd have this taken care of for the user within the training loop, but as long as we have to specify the device outside of the loop, this line has to come before the initialization of the device.

xihui-wu: Got it.

BradLarson (approved):

PR #672, opened 2020-09-25
Add an interface for toggling live statistics in TrainingLoop

xihui-wu:


TrainingLoop/Callbacks/StatisticsRecorder.swift
```diff
@@ -121,3 +121,27 @@ public class StatisticsRecorder {
...
+      shouldCompute = {
+          (
+            _ batchIndex: Int, _ batchCount: Int, _ epochIndex: Int, _ epochCount: Int,
+            _ event: TrainingLoopEvent
+          ) -> Bool in
+          return event == .batchEnd && batchIndex + 1 == batchCount
+        }
+    }
+  }
+}
```
dabrahams, line 147:
```suggestion
extension StatisticsRecorder {
  /// The events on which statistics will be reported for training and validation phases.
  public enum ReportTrigger {
    /// Report statistics at end of training and validation, once per epoch.
    endOfEpoch,
    /// Report statistics at end of training and validation, once per batch.
    endOfBatch
  }
  
  func setReportTrigger(_ trigger: ReportTrigger) {
    if trigger == .endOfBatch {
      shouldCompute = {
          (
            _ batchIndex: Int, _ batchCount: Int, _ epochIndex: Int, _ epochCount: Int,
            _ event: TrainingLoopEvent
          ) -> Bool in
          return event == .batchEnd
        }
    } else {
      shouldCompute = {
          (
            _ batchIndex: Int, _ batchCount: Int, _ epochIndex: Int, _ epochCount: Int,
            _ event: TrainingLoopEvent
          ) -> Bool in
          return event == .batchEnd && batchIndex + 1 == batchCount
        }
    }
  }
}
```

TrainingLoop/Callbacks/StatisticsRecorder.swift
```diff
@@ -121,3 +121,34 @@ public class StatisticsRecorder {
...
+extension StatisticsRecorder {
+  /// The events on which statistics will be reported for training and validation phases.
+  public enum ReportTrigger {
+    /// Report statistics at end of training and validation, once per epoch.
+    case endOfEpoch
+    /// Report statistics at end of training and validation, once per batch.
+    case endOfBatch
+  }
+
+  /// Set the StatisticsRecorder to report statistics when `trigger` is triggered.
```
dabrahams, line 134: Grammatically, it would be “Sets…”, not “Set…”. Do you understand the difference?  It's called the “indicative mood.”

Triggers don't get triggered; they trigger something else (e.g. a gun firing).

```suggestion
  /// Updates `self` to report statistics when `trigger` is encountered.
```

dabrahams (approved):

TrainingLoop/Callbacks/StatisticsRecorder.swift
xihui-wu: Learnt it. :)

PR #674, opened 2020-09-29
Fix out of bound points crashing line drawing function

joaqo:
When drawing lines on images, if one of the points in the line is outside the image bounds, it causes a crash in the line drawing function. This PR fixes that.


BradLarson (approved):
Looks good to me. Thanks for the fix!

PR #677, opened 2020-10-09
Use TrainingLoop in Autoencoder1D

xihui-wu:


Autoencoder/Autoencoder1D/main.swift
```diff
@@ -16,68 +16,65 @@ import Datasets
...
-                directory: outputFolder, name: "epoch-\(epoch)-input")
-            try saveImage(
-                testImages[0..<1], shape: (imageWidth, imageHeight), format: .grayscale,
-                directory: outputFolder, name: "epoch-\(epoch)-output")
-        } catch {
-            print("Could not save image with error: \(error)")
-        }
+var trainingLoop = TrainingLoop(
+  training: dataset.training.map { $0.map { LabeledData(data: $0.data, label: $0.data) } },
+  validation: dataset.validation.map { LabeledData(data: $0.data, label: $0.data) },
```
BradLarson, line 75: I don't think this is operating in the same way as the original model. For an autoencoder like this, we just want to train through the images once, then output a single input / output image pair at the end as the validation stage. We don't want to iterate through the validation set.

Also, in the original model the input / output images were captured before training, to show the random state at the very beginning. The test images themselves were different at each epoch, and they appear to be constant here.

xihui-wu: Thanks for pointing out. Yes we'd better select different images every epoch to save for nicer manual validation.
Regarding iterating through the validation set, I'd in favor of keeping it for its is a standard way of computing the stats.

BradLarson (approved):
The conversion makes sense, and is closer to the original implementation that what we have in the models right now, so this looks great to me.

Thanks for working this through.

PR #678, opened 2020-10-12
Adding Catalina-compatible FileHandle operations

BradLarson:
PR #668 added an `append()` method to the FileSystem protocol and an implementation in FoundationFileSystem. This implementation used FileHandle and several methods which have had their signatures change between macOS 10.15 (Catalina) and 10.16 (Big Sur). The current implementation uses the new interface, which will build with the latest Linux toolchain but will not build on the current macOS 10.15 (Catalina).

This provides the older interface for use when building on macOS, and the newer one for other OS versions. Once Big Sur is the dominant build environment for macOS, we can remove this fallback.


xihui-wu (approved):
Thanks for the fix!

PR #679, opened 2020-10-14
Convert to TrainingLoop for BERTCoLA

xihui-wu:
Some highlights of the changes:
1. Add a LearningRateScheduler callback for TrainingLoop and move Models/Text/ScheduledParameters.swift code to there.
2. Add mcc measurer for TrainingLoop and move Models/Text/Evaluation.swift code to there.


BradLarson (approved):
As we'd discussed earlier, we'll have a follow-on review of the learning rate schedule design.

I tested out GPU training for the BERT-CoLA model, and it seems to work as it did before. The code looks to function, so if the migration of this to the new training loop is blocking other work, we can merge this in now and follow on with refinements of the scheduler and other components.

PR #681, opened 2020-10-21
Clean up BERT model

xihui-wu:
1. We can now remove Models/Text/WeightDecayedAdam.swift after https://github.com/tensorflow/swift-apis/pull/1107 is merged.
2. Move files to BERT folder.


texasmichelle (approved):
Thank you for this cleanup, well done!!

PR #682, opened 2020-10-21
Removed extraneous comma

mlsawyer:
The extraneous comma prevented the notebook from loading in colab.


BradLarson (approved):
Ha, good catch. Wonder how that crept in there.

Thanks for the fix.

PR #683, opened 2020-10-22
Fixed links to Colab and GitHub

mlsawyer:
The links did't point the master branch of tensorflow/swift-model.


BradLarson (approved):
Thanks again for spotting this. I'd changed the main import links, but totally forgot to update the image links. I appreciate the fix.

PR #684, opened 2020-10-22
Update Shallow Water PDE example to use GIF writer

vojtamolda:
Hello S4TF team,

Here's a small update to PR #666. At the time the GIF writer wasn't yet merged so the example relied on ImageMagick's `convert` utility to create animations from a sequence of JPEG images. Now the GIFs are written directly without the conversion step.

Because I only need to store grayscale animations, I added support to save single channel tensor of shape `[width, height, 1]` as GIFs to `AnimatedImage`

Cheers,

  Vojta


BradLarson:
Are you getting a high enough quality with the grayscale quantization you've implemented? I'm asking, because better quantization was on my to-do list in part to improve the results from your model. If your grayscale-oriented quantization works well enough, I can de-prioritize that.

Also, now that the intermediate images are not required, would it be possible to rework the image input and image output paths to support running this at the root of the swift-models directory? All of our other models work that way, and I've been standardizing on using the `output/` directory for output products. That would remove a step from the use of this model and bring it in line with our others.

Thanks for the follow-on, this will save me some work.

vojtamolda:
Yeah. I think the quality is okay. For my purposes I just map fixed -1 to +1  range of solution values onto a palette of 64 grays spanning from white to black. I tried both larger and smaller palettes and 64 seems like the best compromise between quality and file size. For the time being I don't have a problem with de-prioritizing improvements in this area. I tried using the existing default color palette but the grays there were just too sparse.

I agree that running from the root is a much better solution. I can add a CLI argument for the input image and save the GIFs in the `output/` folder by default. I'll look around the other examples to see how things are done there and update the PR.

vojtamolda:
Here's how the updated CLI looks like:
```
Solve shallow water PDE on a unit square.

Animations of the solution are saved in the 'output' directory.

USAGE: Shallow-Water-PDE [--splash] [--optimization] [--benchmark] [--resolution <N>]
[--duration <T>] [--target <image>] [--iterations <I>] [--learning-rate <α>]

OPTIONS:
  --splash/--optimization/--benchmark
                          Task to run. 
  --resolution <N>        Number of simulated values along X/Y directions. (default: 256)
  --duration <T>          Number of simulated time-steps. (default: 512)
  --target <image>        Image to use as an optimization target. (default:
                          Examples/Shallow-Water-PDE/Images/Target.jpg)
  --iterations <I>        Number of optimization iterations. (default: 200)
  --learning-rate <α>     Optimization learning rate. (default: 500.0)
  -h, --help              Show help information.
```

BradLarson (approved):
I ran through everything, and it looks great. It's a ton easier to just run this and see the results immediately.

Also, I think I have a good use for the grayscale quantization in another application, so that will save me some time in prototyping. Thanks for the nice update!

vojtamolda:
Perfect! Thanks for the review and comments, Brad.

PR #685, opened 2020-10-22
Refactor LearningRateScheduler

xihui-wu:
This [line](https://github.com/tensorflow/swift-models/blob/master/TrainingLoop/Callbacks/LearningRateScheduler.swift#L55) of code and below were taken into that file from another earlier existing file. It implemented the fundamental learning rate scheduling algorithms and two of them were used by BERT model. 

It was a good starting point to build our TrainingLoop LearningRateScheduler Callback so I took it as it is in https://github.com/tensorflow/swift-models/pull/679. However this is not the right interface for user to use, so this followup PR comes out to implement a higher level interface.

### DIFF
--- a/TrainingLoop/Callbacks/LearningRateScheduler.swift
+++ b/TrainingLoop/Callbacks/LearningRateScheduler.swift
@@ -11,20 +11,17 @@ public enum ScheduleShape {
 
 public struct WarmupSchedule {
   public var shape: ScheduleShape
-  public var endLearningRate: Float = 0
+  public var endLearningRate: Float
   public var endStep: Float
 
-  public func toScheduledParameter<WSP: ScheduledParameter>() -> WSP? {
-    let endParameter = FixedParameter<Float>(endLearningRate)
-    switch shape {
-    case .linear:
-      return LinearlyWarmedUpParameter(
-        baseParameter: endParameter,
-        warmUpStepCount: UInt64(endStep),
-        warmUpOffset: 0) as? WSP
-    default:
-      return nil
-    }
+  public init(shape: ScheduleShape, endLearningRate: Float, endStep: Float) {
+    self.shape = shape
+    self.endLearningRate = endLearningRate
+    self.endStep = endStep
+  }
+
+  public func toScheduledParameter() -> WarmupScheduledParameter {
+    return WarmupScheduledParameter(warmupSchedule: self)
   }
 }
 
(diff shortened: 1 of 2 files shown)


TrainingLoop/Callbacks/LearningRateScheduler.swift
```diff
@@ -1,54 +1,192 @@
...
+  public var shape: ScheduleShape
+
+  /// The learning rate with which warmup ends.
+  public var endLearningRate: Float
+
+  /// The step at which warmup ends.
+  public var endStep: Float
+
+  /// Create an instance of WarmupSchedule with the curve in `shape` shape and
+  /// warmup ends with `endLearningRate` rate at `endStep` step.
```
dabrahams, line 26:
```suggestion
  /// Creates an instance with the curve in `shape`,
  /// ending with `endLearningRate`  at `endStep`.
```

TrainingLoop/Callbacks/LearningRateScheduler.swift
```diff
@@ -1,54 +1,192 @@
...
+/// warmup steps during training.
+public struct WarmupSchedule {
+  /// The shape of learning rate curve.
+  public var shape: ScheduleShape
+
+  /// The learning rate with which warmup ends.
+  public var endLearningRate: Float
+
+  /// The step at which warmup ends.
+  public var endStep: Float
```
dabrahams, line 23:
```suggestion
  public var endStep: Int
```

TrainingLoop/Callbacks/LearningRateScheduler.swift
```diff
@@ -1,54 +1,192 @@
...
+
+  /// Create an instance of WarmupSchedule with the curve in `shape` shape and
+  /// warmup ends with `endLearningRate` rate at `endStep` step.
+  public init(shape: ScheduleShape, endLearningRate: Float, endStep: Float) {
```
dabrahams, line 27:
```suggestion
  public init(shape: ScheduleShape, end rate: Float, _ step: Int) {
```

```suggestion
  public init(shape: ScheduleShape, finalRate: Float, finalStep: Int) {
```

```suggestion
  public init(shape: ScheduleShape, endLearningRate: Float, steps: Int) {
```

```suggestion
  public init(shape: ScheduleShape, endWith rate: Float, endAt step: Int) {
```

```suggestion
  public init(shape: ScheduleShape, endRate: Float, endStep: Int) {
```

xihui-wu:
Have a new one here https://github.com/tensorflow/swift-models/pull/711

PR #686, opened 2020-10-23
Update FileHandler value -> suffix to match the changes to the Linux side

BradLarson:
In PR #670, the parameter to FoundationFileSystem's append() method was changed from `value` to `suffix` on the Linux side, but not for macOS. This causes model compilation to fail on macOS.

This quick fix updates the macOS side to re-enable building models there.


xihui-wu (approved):
Thanks for the fix!

PR #687, opened 2020-10-24
Install and import Checkpoints

mlsawyer:
Notebook needed to be updated to install and import Checkpoints


BradLarson (approved):
This looks great. Thanks for finding issues and keeping these up to date.

PR #688, opened 2020-10-25
add imagenet dataset + training demo

brettkoonce:
would like feedback on best way to transfer file and best way to strip alpha channel from incoming (out of spec) images.


brettkoonce:
ps happy to upload file to a bucket if that helps

brettkoonce:
uploaded imagnet to a bucket, happy to share!

BradLarson:
Sorry for the delay in testing this, working with the dataset is a little unwieldy. One quick comment that I have is that I believe they capitalize the N in ImageNet in [the official branding](http://www.image-net.org/).

brettkoonce:
@BradLarson updated!

Datasets/Imagenette/ImageNet.swift
```diff
@@ -0,0 +1,230 @@
...
+  //   filename: "\(base)\(size.suffix)", fileExtension: "tgz",
+  //   remoteRoot: location.deletingLastPathComponent(), localStorageDirectory: directory)
+  // END ORIGINAL CODE
+
+  print("Downloading file to '/tmp/imagenet.tgz'.")
+  let downloadProcess = Process()
+  downloadProcess.environment = ProcessInfo.processInfo.environment
+  downloadProcess.executableURL = URL(fileURLWithPath: "/bin/bash")
+  // doesn't work, needs to wait for download
+  downloadProcess.arguments = ["wget -O /tmp/imagenet.tgz https://REMOTE-SERVER/imagenet/imagenet.tgz"]
```
compnerd, line 148: Can we use `FoundationNetworking` or some of the helpers that we have in the support module?  This definitely won't work on all platforms (e.g. Windows).

Datasets/Imagenette/ImageNet.swift
```diff
@@ -0,0 +1,230 @@
...
+
+  // this approach tries to work in memory --> ~150GB in-memory download --> hits swap  -> stream to file instead?
+  // let location = URL(
+  //   string: "https://REMOTE-SERVER/imagenet/imagenet.tgz")!
+  // let _ = DatasetUtilities.downloadResource(
+  //   filename: "\(base)\(size.suffix)", fileExtension: "tgz",
+  //   remoteRoot: location.deletingLastPathComponent(), localStorageDirectory: directory)
+  // END ORIGINAL CODE
+
+  print("Downloading file to '/tmp/imagenet.tgz'.")
```
compnerd, line 143: Debugging left over?

Datasets/Imagenette/ImageNet.swift
```diff
@@ -0,0 +1,230 @@
...
+  downloadProcess.environment = ProcessInfo.processInfo.environment
+  downloadProcess.executableURL = URL(fileURLWithPath: "/bin/bash")
+  // doesn't work, needs to wait for download
+  downloadProcess.arguments = ["wget -O /tmp/imagenet.tgz https://REMOTE-SERVER/imagenet/imagenet.tgz"]
+  try! downloadProcess.run()
+  downloadProcess.waitUntilExit()
+
+  let wouldLikeToTestLocalExtract = false
+  if (wouldLikeToTestLocalExtract) {
+      print("Extracting file at '/tmp/imagenet.tgz'.")
```
compnerd, line 154: Debugging left overs?

Datasets/Imagenette/ImageNet.swift
```diff
@@ -0,0 +1,230 @@
...
+      extractArchive(at: URL(string:"/tmp/imagenet.tgz")!, to: URL(string: downloadPath)!,
+                   fileExtension: "tgz", deleteArchiveWhenDone: false)
+      print("Done extracting'.")
```
compnerd, line 157: Same.

Datasets/Imagenette/ImageNet.swift
```diff
@@ -0,0 +1,230 @@
...
+//    https://raw.githubusercontent.com/soumith/imagenetloader.torch/master/valprep.sh
+//    remove blacklisted validation images:
+//    https://raw.githubusercontent.com/fastai/imagenet-fast/master/imagenet_nv/blacklist.sh
+// 2) create imagenet.tgz: tar -czvf imagenet.tgz train val
+
+import Foundation
+import ModelSupport
+import TensorFlow
+
+public struct ImageNet<Entropy: RandomNumberGenerator> {
```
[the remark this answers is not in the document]
brettkoonce: this comes from Imagenette/Imagewoof.swift, for what it's worth.

Datasets/Imagenette/ImageNet.swift
brettkoonce: The existing approach (eg https://github.com/tensorflow/swift-models/blob/7e93db5c4f6e454fc60c1fb7d79c41b53aa1ec37/Datasets/DatasetUtilities.swift#L70) fails because it is loading (150GB of) data into memory.  If there is a better way to download/untar said file happy to update to match that.

Datasets/Imagenette/ImageNet.swift
brettkoonce: See above!

Datasets/Imagenette/ImageNet.swift
brettkoonce: The extractArchive call works nicely if the file is present locally.  The test harness can be removed once there's a proper way to download the file.

Datasets/Imagenette/ImageNet.swift
brettkoonce: takes a while to extract this file, so a some sort of feedback to the user is helpful.

brettkoonce:
@BradLarson [some of the files](https://github.com/tensorflow/datasets/blob/7c4c9f823a7409163fc02af63f8fbb39a793080a/tensorflow_datasets/image_classification/imagenet.py#L108) are CMYK, will STBImage handle them properly?

BradLarson:
@brettkoonce - I believe [CMYK is supported](https://github.com/tensorflow/swift-models/blob/main/Support/STBImage/include/stb_image.h#L3809), but the only way to be sure would be to pull out one of those images and try loading and displaying or saving it to make sure. I unfortunately still haven't had a chance to set up a good environment for testing out the remaining ImageNet dataset here.

We also may need to check some of [the experimental JPEGTurbo loading](https://github.com/tensorflow/swift-models/tree/main/ImageLoader) if and when we get the build environment set up to integrate that. I'm guessing compatibility may be different between the two.

brettkoonce:
@BradLarson cymk works:

n01739381_1309
resaved-image

rebased + added tensorboard logging, works here (thanks @xihui-wu!)

happy to upload the .tgz to a bucket/setup a vm somewhere to simplify getting bootstrapped!

brettkoonce:
@BradLarson also, what's the best way to extend resnet to support checkpointing?  There be a fair amount of magic in the GPT2 stuff.

BradLarson:
From the CMake build failure, it looks like you might need to add ImageNet to the CMakeLists.txt around [here](https://github.com/tensorflow/swift-models/blob/main/Datasets/CMakeLists.txt#L10).

Regarding checkpoint loading / saving, I need to fix the tests and convert [this](https://github.com/tensorflow/swift-models/pull/631) to a full pull request, because that should lay the groundwork for more generalized checkpoint loading. In particular, the programmatic mapping of tensor names from on-disk checkpoints to internal model naming will simplify the loading and saving process. I'm very behind on that, but I'll try to move it forward.

brettkoonce:
@BradLarson updated the cmake file, thanks!  The only remaining thing we need to train models from scratch is checkpoints.  Would be super-cool to have it working!

BradLarson (approved):
Thanks for pulling this together. We can resolve the issues around hosting such a large dataset later, so let's bring this in now as at least something to build from.

Having the full ImageNet 2012 dataset as a training source is a significant step forward for image classification models.

PR #692, opened 2020-10-27
WordSeg: repair build on Windows

compnerd:
Importing `ucrt` is insufficient since we rely on the type-generic
operation to convert the `Float` to what would be operations on
`Double` otherwise.


compnerd:
Note that the complexity here is due to the fact that the TF branch is significantly behind main at this point.  MSVCRT has been renamed CRT and this just will make the transition easier for swift-models.

texasmichelle (approved):

BradLarson (approved):

PR #693, opened 2020-10-27
Shallow Water PDE: make build on Windows

compnerd:
We need to import `MSVCRT` (or rather the new name of `CRT`) to get the
type-generic operations as the C library functions which are referenced
here expect double values but the values being used are single precision
(`Float`) which fails to type-check.


compnerd:
Note the complexity here is due to TF being behind main which has renamed `MSVCRT` to `CRT`.  In theory, only `CRT` should be needed.

texasmichelle (approved):

BradLarson (approved):

PR #694, opened 2020-10-28
Take SwiftProtobuf dependency out from ModelSupport

xihui-wu:
This will avoid redundant heavy SwiftProtobuf import when ModelSupport is imported in other non-transformer cases.


BradLarson:
Rather than moving all of Tokenization into its own module, would it be possible to just move [this one extension on Vocabulary](https://github.com/tensorflow/swift-models/blob/master/Support/Text/Tokenization.swift#L91) into the BERT model? I believe [this](https://github.com/tensorflow/swift-models/blob/97f386958a43bd363479d29ad590cdcba0f97b85/Models/Text/BERT/BERT.swift#L722) is the only place that extension is used.

Having all of Tokenization in its own module would still cause a dependency on SwiftProtobuf for all Datasets, when none of them need it. I believe that relocating that one extension into BERT would let you remove the SwiftProtobuf dependency from ModelSupport and everywhere else, and then only require you to add that dependency to TextModels and move the protos to be underneath that module.

BradLarson (approved):
That looks great, and is a really simple solution. This will be a significant help with build times.

Thanks for taking the extra time to simplify this further.

PR #695, opened 2020-10-28
docs: update build docs for Windows

compnerd:
Windows snapshots as of 10/28 have SwiftPM in the snapshots and works
to build the models.  Update the docs for build instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ swift build -Xswiftc -use-ld=lld -c release
 #### Test
 
 ```cmd
-swift test -Xswiftc -use-ld=lld -c release
+swift test -Xswiftc -use-ld=lld -c debug
 ```
 
 ### Using CMake for Development
--- a/README.md
+++ b/README.md
@@ -29,13 +29,25 @@ swift test
 
 ### Windows
 
+#### Build
+
+```cmd
+swift build -Xswiftc -use-ld=lld -c release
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,13 +29,25 @@ swift test
 
 ### Windows
```
texasmichelle, line 31: How about adding a note that this only works with the 0.12 RC?

README.md
```diff
@@ -29,13 +29,25 @@ swift test
...
+#### Build
+
+```cmd
+swift build -Xswiftc -use-ld=lld -c release
+```
+
+#### Test
+
+```cmd
+swift test -Xswiftc -use-ld=lld -c release
```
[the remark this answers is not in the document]
texasmichelle: This command doesn't actually succeed (I get duplicate symbol errors), so it would be good to add fixes or a note here.

texasmichelle (approved):
Combining the copy commands in https://github.com/tensorflow/swift-models/tree/master/Examples/GPT2-Inference#run would round out these updates 👍

README.md
compnerd: It works with the currently linked snapshot as well, so I think that this is accurate as is.

compnerd:
Yes, I will try to update the other instructions in a follow up change.

PR #698, opened 2020-11-12
Update Package.swift

compnerd:
Jump to the latest swift-argument-parser as it is needed for Windows.


texasmichelle (approved):
LGTM as long as kokoro passes, e.g. it doesn't break other platforms.

PR #699, opened 2020-11-12
build: correct missing dependency for `Models`

compnerd:
Repair the CMake based build which was under-specifying dependencies.


texasmichelle (approved):
Tested on Windows and fixes the cmake build 🎉

PR #702, opened 2020-11-23
rename Benchmarks to SwiftModelsBenchmarks

marcrasi:
Swift does not like it when you depend on packages that have targets with the same name (https://forums.swift.org/t/multiple-target-issue-with-spm/16696).

"Benchmarks" conflicts with "Benchmarks" in Penguin, which makes it impossible to depend on swift-models and Penguin at the same time.

Prefixing the target name with the package name seems like a reasonable way to prevent these name conflicts.

We have already changed Penguin's "Benchmarks" to "PenguinBenchmarks" so this swift-models PR isn't actually necessary to unbreak anything right now, but it seems like a good idea anyways to prevent more name conflicts in the future.


BradLarson (approved):
It's a bit of a bummer that we can't keep the simplicity of `swift run Benchmarks`, but I understand the need. We built this before swift-benchmark existed, and didn't anticipate the conflict. This should be just fine, and it's great that things will start working reliably again.

marcrasi:
Good point about `swift run Benchmarks`. We can actually rename the target, but still have `swift run Benchmarks`, by explicitly specifying the executable product. I uploaded a commit that does that.

BradLarson:
Great, that's even better. Looks good to me.

PR #707, opened 2020-11-28
Add Xception network

diwakar-vsingh:
I have created the Xception model which is the backbone in the deepLabV3+ model.


Models/ImageClassification/Xception.swift
```diff
@@ -0,0 +1,274 @@
...
+// limitations under the License.
+
+import TensorFlow
+
+// Original Paper:
+// "Xception: Deep Learning with Depthwise Separable Convolutions"
+// François Chollet
+// https://arxiv.org/abs/1610.02357
+
+public struct ConvBlock: Layer {
```
BradLarson, line 22: It looks like the definition of ConvBlock here is conflicting with the existing ConvBlock in MobileNetV1.swift (leading to a failure in compilation via `swift build`). If these blocks aren't being re-used, we could consider possibly making them fileprivate or placing each within the namespace of the model itself.

diwakar-vsingh: I think the best way to move forward is to place each ConvBlock within the namespace of the model itself as no one will be using these blocks outside the scope of the model itself.

diwakar-vsingh (commented):
@BradLarson Also could you direct me to resources so that I can learn to run swift build on my side itself.

BradLarson:
@diwakar-vsingh - Sorry for the slow reply. In regards to building this on your own, that depends on which platform you wanted to work from.

If you're on a Mac, it's pretty straightforward to [download a precompiled toolchain](https://github.com/tensorflow/swift/blob/main/Installation.md#releases) and follow [these instructions](https://github.com/tensorflow/swift/blob/main/Installation.md#macos) to install it within Xcode. You can then open the directory of your local clone of swift-models directly into Xcode and it will be parsed into a project. Select the ImageClassificationModels build target and you'll be able to build this model and the code around it by itself.

For Ubuntu, you can [download a precompiled toolchain](https://github.com/tensorflow/swift/blob/main/Installation.md#releases) and follow [these instructions](https://github.com/tensorflow/swift/blob/main/Installation.md#linux) to install the appropriate dependencies and to place the Swift toolchain in an accessible location. You can then go to your fork of swift-models and run `swift build` to build the entire project. This is the closest to what's run during our presubmit tests.

diwakar-vsingh:
@BradLarson Thanks for the help. I have just set up everything locally on my Mac. I followed the instructions given by you above and everything builds fine locally. I have changed the ConvBlock layer name to ConvBlockModule so that it doesn't conflict with the one define in MobileNetV1.swift.

diwakar-vsingh:
I have also added an inference test for the Xception model in inference.swift.

diwakar-vsingh:
> If you're on a Mac, it's pretty straightforward to [download a precompiled toolchain](https://github.com/tensorflow/swift/blob/main/Installation.md#releases) and follow [these instructions](https://github.com/tensorflow/swift/blob/main/Installation.md#macos) to install it within Xcode

One thing I couldn't set up is to use the Swift toolchain with the command-line tool. Whenever I run the command `xcrun --toolchain swift`, it always throws an error stating "xcrun: error: no utility name specified." Do you know why am I not able to set this step? My Xcode version is 12.3.

BradLarson (approved):
Thank you for the model implementation, and for the quick response to my comments.

Sorry for the delay in review, but I had to fix something on our CI side to get the presubmit to build this correctly. That's now fixed, and it all builds correctly, so this looks good to go.

BradLarson:
@diwakar-vsingh - Honestly, when building at the command line I tend to just use `swift build` or `swift run [target]` rather than going through xcrun. If you're missing xcrun, you may need to install the supplemental developer tools via `xcode-select --install`. swift-models is set up as a Swift Package Manager package which doesn't need Xcode to build, so it's often simpler to just use Swift Package Manager directly.

diwakar-vsingh:
> Thank you for the model implementation, and for the quick response to my comments.
> 
> Sorry for the delay in review, but I had to fix something on our CI side to get the presubmit to build this correctly. That's now fixed, and it all builds correctly, so this looks good to go.

Thanks for merging the branch. I'm looking forward to committing more.

diwakar-vsingh:
> Honestly, when building at the command line I tend to just use `swift build` or `swift run [target]` rather than going through xcrun.

Whenever I try to build it directly from the command line using `swift build`, I get the error that it can't import Tensorflow. That's why I was trying to follow the instructions mentioned [here](https://github.com/tensorflow/swift/blob/main/Installation.md#macos).

BradLarson:
@diwakar-vsingh - If it can't find TensorFlow, odds are that the Swift compiler being found at the command line is the default toolchain within Xcode. Typically, what I do in that case is either set the toolchain that's installed in Xcode to the Swift for TensorFlow toolchain (done by going to the Xcode | Toolchains menu option, then choosing the right one within that menu) and then use `export PATH=/Library/Developer/Toolchains/swift-latest/usr/bin:"${PATH}"` at the command line to set the path to that, or use an `export PATH=` statement that points to the explicit path of the Swift for TensorFlow toolchain. Sorry that dealing with our custom toolchain is a little bit of a pain on macOS.

diwakar-vsingh:
@BradLarson  Fixed it. My path was not correct. I had to resort to using `export PATH=` statement that points to the explicit path of the Swift for TensorFlow toolchain. Thanks for the help.

PR #715, opened 2020-12-10
Add perplexity metric for GPT2-WikiText2

xihui-wu:


TrainingLoop/Metrics.swift
```diff
@@ -14,6 +15,8 @@ public enum TrainingMetrics {
       return "accuracy"
     case .matthewsCorrelationCoefficient:
       return "mcc"
+    case .perplexity:
+      return "ppl"
```
BradLarson, line 19: Should we use the full word of "perplexity" here to make this clear? "ppl" isn't a standard abbreviation that I'm familiar with, and I'd have a hard time associating it with perplexity.

TrainingLoop/Metrics.swift
```diff
@@ -190,3 +195,41 @@ public struct MCCMeasurer: MetricsMeasurer {
...
+  public var name: String
+
+  /// Sum of losses cumulated from batches.
+  private var totalBatchLoss: Float = 0
+
+  /// Count of batchs cumulated so far.
+  private var batchCount: Int32 = 0
+
+  /// Creates an instance with the PerplexityMeasurer named `name`.
+  public init(_ name: String = "ppl") {
```
BradLarson, line 211: My comments about "ppl" above might apply here, as well.

TrainingLoop/Metrics.swift
```diff
@@ -190,3 +195,41 @@ public struct MCCMeasurer: MetricsMeasurer {
...
+  ) {
+    if let newBatchLoss = loss {
+      totalBatchLoss += newBatchLoss.scalarized()
+      batchCount += 1
+    }
+  }
+
+  /// Computes averaged perplexity as e^(averaged loss).
+  public func measure() -> Float {
+    return exp(totalBatchLoss / Float(batchCount))
```
BradLarson, line 233: Is there ever a case where this will be called with `batchCount` equal to zero? Do we need to guard against that?

In trying to match this to [HuggingFace's explanation](https://huggingface.co/transformers/perplexity.html), is loss assumed to be equal to log likelihood and batch count equal to the `end_loc` they calculate there? Do we need to state that in the metric description or name the properties to match?

xihui-wu: The measurer is used in reset -> accumulate -> measure sequence so it won't cause batchCount be zero, but I can guard it for safety.

I think it is aligned. The perplexity is computed as  e^(per token cross entropy loss). Let's add the comment that we expect the loss to be per token cross entropy loss ?

BradLarson: A comment explaining the values makes sense, and would answer future questions like mine. Thanks for the explanation.

TrainingLoop/Metrics.swift
xihui-wu: done.

TrainingLoop/Metrics.swift
xihui-wu: done

BradLarson (approved):

PR #716, opened 2020-12-10
Repairing two text-based tests on macOS

BradLarson:
The `testCreateWikiText103WithoutBpe` test will always fail [a precondition](https://github.com/tensorflow/swift-models/blob/main/Datasets/TextUnsupervised/TextUnsupervised.swift#L212) because it sets bpe to nil, while `encodedFileName` is also nil for that dataset. This test case is not active on Linux, but will run on macOS, where it crashes tests. As a result, I've removed this test case.

Additionally, the WordSegDataset used a hardcoded path to /dev/null when no filename is specified for two components. Direct access to /dev/null is blocked on macOS, and is problematic for Windows. I've replaced this with creation of an empty Data instance if the file name is missing.

With these changes, all tests now pass on macOS.

### DIFF
--- a/Datasets/WordSeg/WordSegDataset.swift
+++ b/Datasets/WordSeg/WordSegDataset.swift
@@ -147,13 +147,19 @@ public struct WordSegDataset {
       contentsOf: URL(fileURLWithPath: trainingFile),
       options: .alwaysMapped)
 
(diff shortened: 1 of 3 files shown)


xihui-wu (approved):
Thanks for the fix!

PR #717, opened 2020-12-11
Work around SR-13945 segfault.

pschuh:
See https://bugs.swift.org/browse/SR-13945 for more info.

### DIFF
--- a/Tests/TextTests/WordSegmentationTests/ProbeLayers.swift
+++ b/Tests/TextTests/WordSegmentationTests/ProbeLayers.swift
@@ -184,6 +184,8 @@ class WordSegProbeLayerTests: XCTestCase {
     let lattice = model.buildLattice(abab, maxLen: 5, device: device)
     XCTAssert(lattice.isAlmostEqual(to: Example1.lattice, tolerance: 1e-5))
 
+/*
+// TODO: renable once SR-13945 is fixed.
     func f(_ x: SNLM) -> Float {
       x.buildLattice(abab, maxLen: 5, device: device)[4].semiringScore.logr
     }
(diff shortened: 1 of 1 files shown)


marcrasi:
looks like the test flaked while downloading something

marcrasi (approved):
Wow this looks difficult.

I hope someone fixes SR-13945 soon so that we can undo this.

Looks good to me as long as the tests pass.

pschuh:
> Wow this looks difficult.
> 
> I hope someone fixes [SR-13945](https://bugs.swift.org/browse/SR-13945) soon so that we can undo this.
> 
> Looks good to me as long as the tests pass.

Should we consider disabling the test instead?

marcrasi:
> Should we consider disabling the test instead?

Yeah, that sounds like it would be easier. Wordseg is going to be completely broken until we fix SR-13954, but I think no one is using wordseg these days anyways.

texasmichelle (approved):
I also tried a bunch of things to work around this and they were all ugly. Disabling until the upstream regression is fixed SGTM

Tests/TextTests/WordSegmentationTests/ProbeLayers.swift
```diff
@@ -184,6 +184,8 @@ class WordSegProbeLayerTests: XCTestCase {
     let lattice = model.buildLattice(abab, maxLen: 5, device: device)
     XCTAssert(lattice.isAlmostEqual(to: Example1.lattice, tolerance: 1e-5))
 
+/*
+// TODO: renable once SR-13945 is fixed.
```
texasmichelle, line 188:
```suggestion
// TODO: re-enable once SR-13945 is fixed.
```

PR #723, opened 2020-12-17
add presubmit_build.sh that passes a flag to swiftpm if it is a stock toolchain

marcrasi:
I copied this from the internal script in google3, and modified it to pass a flag when building with a stock toolchain.

I'll also make a CL that deletes the script from google3 and calls this one instead.

### DIFF
--- a/docker/presubmit_build.sh
+++ b/docker/presubmit_build.sh
@@ -31,8 +31,8 @@ sudo -E docker build -t built-img -f docker/Dockerfile --build-arg swift_tf_url 
 
 # SwiftPM-based build.
 docker run built-img /bin/bash -c "
-swift build ${STOCK_TOOLCHAIN:+"-D -Xswiftc -D TENSORFLOW_USE_STANDARD_TOOLCHAIN"} ;
-swift test ${STOCK_TOOLCHAIN:+"-D -Xswiftc -D TENSORFLOW_USE_STANDARD_TOOLCHAIN"} ;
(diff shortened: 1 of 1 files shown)


docker/presubmit_build.sh
```diff
@@ -0,0 +1,43 @@
...
+# Help debug the job's disk space.
+df -h
+
+cd github/swift-models
+sudo -E docker build -t built-img -f docker/Dockerfile --build-arg swift_tf_url .
+
+# SwiftPM-based build.
+docker run built-img /bin/bash -c "
+swift build ${STOCK_TOOLCHAIN:+"-D -Xswiftc -D TENSORFLOW_USE_STANDARD_TOOLCHAIN"} ;
+swift test ${STOCK_TOOLCHAIN:+"-D -Xswiftc -D TENSORFLOW_USE_STANDARD_TOOLCHAIN"} ;
```
BradLarson, line 35: I think there might be an extra -D in both of these. Should it just be `"-Xswiftc -D TENSORFLOW_USE_STANDARD_TOOLCHAIN"`?

marcrasi: Good catch. And there should also be another `-Xswiftc`, right?

BradLarson: I think just the one is needed if you smoosh the last two together into `"-Xswiftc -DTENSORFLOW_USE_STANDARD_TOOLCHAIN"`. That works for me at the command line.

BradLarson: But I tried what you have now with the two of them, and that also works well.

BradLarson (approved):

PR #724, opened 2020-12-22
Stock toolchain cleanup

BradLarson:
AnyLayer can now build using both the stock (+swift-apis) and current Swift for TensorFlow toolchains, so the define-based conditionals have been removed. This also removes the need for explicit Numerics imports, due to Numerics now being exported from swift-apis.

Due to slight differences in ElementaryFunctions, I've temporarily disabled a handful of exponential and logarithm variants within AnyLayer (which weren't being used for anything currently, anyways).

With these changes, swift-models can now build without any extra compiler flags for both the current Swift for TensorFlow toolchains and the new stock (+swift-apis) toolchains.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,7 +27,6 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.10.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", .branch("main")),
         .package(url: "https://github.com/google/swift-benchmark", from: "0.1.0"),
-        .package(url: "https://github.com/apple/swift-numerics", .branch("main")),
     ],
     targets: [
         .target(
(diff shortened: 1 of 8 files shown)


Support/AnyLayer.swift
```diff
@@ -41,27 +40,27 @@ fileprivate func mustOverride(function: StaticString = #function, file: StaticSt
 ///   - Input: the input type of the underlying layar
 ///   - Output: the output type of the underlying layer
 ///   - Scalar: the scalar type of the underlying tangent vector
-internal class AnyLayerBox<Input: Differentiable, Output: Differentiable, Scalar: FloatingPoint & ElementaryFunctions> {
+internal class AnyLayerBox<Input: Differentiable, Output: Differentiable> {
```
dan-zheng, line 43: It's cool that removing the `Scalar` generic parameter from `AnyLayer` is possible!

I wonder how this was possible: is it because `AnyLayer` is now specialized to `Scalar == Float`? If so, that should work for `Tensor`, but not for other types in general.

Package.swift
```diff
@@ -27,7 +27,6 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-protobuf.git", from: "1.10.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", .branch("main")),
         .package(url: "https://github.com/google/swift-benchmark", from: "0.1.0"),
-        .package(url: "https://github.com/apple/swift-numerics", .branch("main")),
```
dan-zheng, line 30: I wonder how it is that `swift-numerics` can be removed as a dependency? What about usages of `ElementaryFunctions`?

Is it because `swift-numerics` is now a dependency for `tensorflow/swift-apis`?

dan-zheng (commented):
Some quick questions.

Package.swift
BradLarson: Yes, at least for the stock toolchains. It's also now [exported from swift-apis](https://github.com/tensorflow/swift-apis/pull/1159), so we can have `ElementaryFunctions` without the conditional imports in builds using the stock toolchains.

dan-zheng: Nice!

Support/AnyLayer.swift
BradLarson: Specializing `Scalar == Float` was the only way I saw to align with the new VectorProtocol in the stock toolchains, which [makes `VectorSpaceScalar` equal to Float](https://github.com/tensorflow/swift-apis/blob/13bc8bb49e4ea4558e1d177772da1d626910a0f1/Sources/TensorFlow/Core/VectorProtocol.swift#L26). This may restrict things too far for general use, but I was just trying to find a way for this to build under both toolchains at present.

dan-zheng: I see, thanks for the context! That sounds good.

dan-zheng (approved):

PR #718, opened 2020-12-11
add top 5 accuracy metric

brettkoonce:
ping @xihui-wu

PR #725, opened 2020-12-22
Fix for device placement when calculating statistics

BradLarson:
Within the statistics callback introduced in PR #718 is a temporary tensor that by default is being placed on the eager device. This is causing a crash at runtime in examples using that callback on X10 devices, so this introduces a quick fix to place that on an appropriate device.

With this, the previously failing examples now run correctly with X10.


xihui-wu (approved):
Thanks for the fix!

PR #729, opened 2021-01-12
Big Transfer

acoadmarmon:
**Resubmitting with proper commit history for google-cla bot.**

Hey S4TF Team!

I've re-implemented the Big Transfer (BiT): General Visual Representation Learning (https://arxiv.org/abs/1912.11370) paper in Swift for Tensorflow. Through this process, I've also added support for CIFAR-100 by slightly modifying the existing CIFAR-10 dataset, as well as implemented Mixup, a StandardizedConv2D layer, and more.

Let me know your thoughts on the code! I've done a lot of cleanup and worked to replicate the existing project structure / testing patterns.

Andrew Marmon

### DIFF
--- a/Datasets/CIFAR100/CIFAR100.swift
+++ b/Datasets/CIFAR100/CIFAR100.swift
@@ -1,0 +1,159 @@
+// Original source:
+// "The CIFAR-100 dataset"
+// Alex Krizhevsky, Vinod Nair, and Geoffrey Hinton.
+// https://www.cs.toronto.edu/~kriz/cifar.html
+
+import Foundation
+import ModelSupport
+import TensorFlow
+
+public struct CIFAR100<Entropy: RandomNumberGenerator> {
+  /// Type of the collection of non-collated batches.
+  public typealias Batches = Slices<Sampling<[(data: [UInt8], label: Int32)], ArraySlice<Int>>>
+  /// The type of the training data, represented as a sequence of epochs, which
+  /// are collection of batches.
+  public typealias Training = LazyMapSequence<
+    TrainingEpochs<[(data: [UInt8], label: Int32)], Entropy>,
+    LazyMapSequence<Batches, LabeledImage>
+  >
+  /// The type of the validation data, represented as a collection of batches.
+  public typealias Validation = LazyMapSequence<Slices<[(data: [UInt8], label: Int32)]>, LabeledImage>
+  /// The training epochs.
+  public let training: Training
+  /// The validation batches.
+  public let validation: Validation
+
+  /// Creates an instance with `batchSize`.
+  ///
+  /// - Parameter entropy: a source of randomness used to shuffle sample 
+  ///   ordering.  It  will be stored in `self`, so if it is only pseudorandom 
+  ///   and has value semantics, the sequence of epochs is deterministic and not 
+  ///   dependent on other operations.
+  public init(batchSize: Int, entropy: Entropy) {
+    self.init(
+      batchSize: batchSize,
+      entropy: entropy,
+      device: Device.default,
+      remoteBinaryArchiveLocation: URL(
+        string: "https://www.cs.toronto.edu/~kriz/cifar-100-binary.tar.gz")!, 
+      normalizing: true)
+  }
+  
+  /// Creates an instance with `batchSize` on `device` using `remoteBinaryArchiveLocation`.
+  ///
+  /// - Parameters:
+  ///   - entropy: a source of randomness used to shuffle sample ordering.  It  
+  ///     will be stored in `self`, so if it is only pseudorandom and has value 
+  ///     semantics, the sequence of epochs is deterministic and not dependent 
+  ///     on other operations.
+  ///   - normalizing: normalizes the batches with the mean and standard deviation
+  ///     of the dataset iff `true`. Default value is `true`.
+  public init(
+    batchSize: Int,
+    entropy: Entropy,
+    device: Device,
+    remoteBinaryArchiveLocation: URL, 
+    localStorageDirectory: URL = DatasetUtilities.defaultDirectory
+      .appendingPathComponent("CIFAR100", isDirectory: true), 
+    normalizing: Bool
+  ){
+    downloadCIFAR100IfNotPresent(from: remoteBinaryArchiveLocation, to: localStorageDirectory)
+    
+    // Training data
+    let trainingSamples = loadCIFAR100TrainingFiles(in: localStorageDirectory)
+    training = TrainingEpochs(samples: trainingSamples, batchSize: batchSize, entropy: entropy)
+      .lazy.map { (batches: Batches) -> LazyMapSequence<Batches, LabeledImage> in
+        return batches.lazy.map{ makeBatch(samples: $0, normalizing: normalizing, device: device) }
+      }
+      
+    // Validation data
+    let validationSamples = loadCIFAR100TestFiles(in: localStorageDirectory)
+    validation = validationSamples.inBatches(of: batchSize).lazy.map {
+      makeBatch(samples: $0, normalizing: normalizing, device: device)
+    }
+  }
+}
+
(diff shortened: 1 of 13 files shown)


Examples/BigTransfer-CIFAR100/main.swift
```diff
@@ -0,0 +1,213 @@
...
+    return [500, 3000, 6000, 9000, 10000]
+  }
+  else {
+    return [500, 6000, 12000, 18000, 20000]
+  }
+}
+
+
+func getLearningRate(step: Int, datasetSize: Int, baseLearningRate: Float = 0.003) -> Float? {
+  /* Returns learning-rate for `step` or nil at the end. */
```
dan-zheng, line 82: Doc comments in Swift are written above the declaration using `///` single line document comment markers. Could you please write the comment like that?

```swift
/// Returns learning-rate for `step` or nil at the end.
///
/// - Parameters:
///   - step: ...
///   - datasetSize: ...
///   - baseLearningRate: ...
```

Definitely check out the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines), especially [the detailed section on writing documentation comments](https://swift.org/documentation/api-design-guidelines#write-doc-comment). I use this resource like a bible - though it's not universally applicable, especially in domain-specific contexts, hence why they are merely guidelines, not rules or laws.

Models/ImageClassification/BigTransfer.swift
```diff
@@ -0,0 +1,328 @@
...
+import TensorFlow
+import PythonKit
+
+let subprocess = Python.import("subprocess")
+let np  = Python.import("numpy")
+
+struct Weights {
+    let name: String
+    let layer: Tensor<Float>
+}
```
dan-zheng, line 16: Could you please add documentation comments to this struct declaration?

```
/// Weights for ... → Note: this needs to be rewritten. I suspect `Weights` is not a good name - maybe `BigTransferParameters` would be better.
struct Weight {
    /// The name of the ... → Note: needs to be rewritten.
    let name: String
    // Note: `layer` is not a good name for a `Tensor<Float>` value, which is not a `Layer` value.
    let layer: Tensor<Float>
}
```

Same for other declarations in this PR, please. Here's the third fundamental principle in the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#write-doc-comment):

> <strong>Write a documentation comment</strong> for every declaration. Insights gained by writing documentation can have a profound impact on your design, so don’t put it off.
> 
> If you are having trouble describing your API’s functionality in simple terms, you may have designed the wrong API.

Thank you!

acoadmarmon: Sounds good! I'll refactor these public declarations as well as add api docs for the functions / structs in BigTransfer.swift and BigTransfer-CIFAR100.swift. 

Let me know if there's any other suggested improvements!

dan-zheng: Thanks Andrew! That would be a big improvement and first step :)

dan-zheng: I personally enjoy precise writing (which in turn requires precise thinking, hence the blurb about "you may have designed the wrong API"). An example is that I love literate programming, which requires performing the exercise of programming twice: once in the formal (programming language), once in the informal (human language).

I hope you find this exercises enjoyable too! It's a big reason why I like writing Swift.

Datasets/CIFAR100/CIFAR100.swift
```diff
@@ -0,0 +1,159 @@
+// Original source:
```
BradLarson, line 1: For this and other .swift files contributed here, we'd desire them to have the standard copyright header used throughout this project. You can copy-and-paste that from any of the other examples in this repository, such lines 1-14 here: https://github.com/tensorflow/swift-models/blob/main/Checkpoints/CheckpointReader.swift#L1

Examples/BigTransfer-CIFAR100/main.swift
```diff
@@ -0,0 +1,213 @@
...
+      }
+      else {
+        continue
+      }
+
+      var (eagerImages, eagerLabels) = (batch.data, batch.label)
+      let npImages = eagerImages.makeNumpyArray()
+      let resized = tf.image.resize(npImages, [resizeSize.0, resizeSize.0])
+      let cropped = tf.image.random_crop(resized, [batchSize, resizeSize.1, resizeSize.1, 3])
+      let flipped = tf.image.random_flip_left_right(cropped)
```
BradLarson, line 163: Rather than requiring access to Python TensorFlow (which I know from experience can be problematic when simultaneously using TensorFlow via our Swift APIs), I'd recommend resizing images using `resize()` as is done here: https://github.com/tensorflow/swift-models/blob/main/Support/Image.swift#L143 . We don't yet have wrappers around this, but cropping can be called via `_Raw.randomCrop`. I think random flipping can be done with the aid of a random number generator and a flipping operation.

By handling these natively, instead of calling out to Python TensorFlow, you'll avoid the roundtrip through Numpy arrays, which should be much faster and use less memory. It'll also allow for better acceleration on processing the dataset,

dan-zheng: <details>
<summary>nothing of review value</summary>

Funny if Python TensorFlow interop (which was broken for years) now works.

</details>

BradLarson:
I don't mean to be a bother, but we're working towards finalizing the last pull requests before archiving the repository and I'd really like to have this in there. Do you believe it will be possible to complete this model by Friday? If so, that should give us one last chance to make a quick pass over it and pull it in.

acoadmarmon:
@BradLarson @dan-zheng I have a lot of time today and tomorrow to finish this up. When is this repo scheduled to be archived? Will I still be able to get it in tomorrow?

Andrew

dan-zheng:
Hi Andrew,

I believe Brad's currently OOO. I'm not sure about an exact archival time, but if you have time to complete the model by this Friday, that would be appreciated. I can update here if an archival time becomes known.

acoadmarmon:
Sounds good. Thanks Dan!

acoadmarmon:
Hey Dan,

With these new changes I've attempted to resolve your concerns around poorly named structs and lack of documentation. I also started work on removing the Python Tensorflow calls, but I've run into some issues. Specifically, when I try to use the _Raw.randomCrop function I get the error: 

`/swift-base/swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift:301: Fatal error: Op RandomCrop is not available in GraphDef version 561. It has been removed in version 8. Random crop is now pure Python.
Illegal instruction`

While debugging I found this implementation of RandomCrop per image which I would need to modify to include batches of data: https://github.com/tensorflow/swift-models/blob/5e5d7b4db27ada4c840fc6ba564125b62449706c/Datasets/Imagenette/ImageNet.swift#L208 . Is this the recommended way to randomly crop an image since it seems _Raw.randomCrop isn't available?

BradLarson:
@acoadmarmon - I'd say that the above-linked implementation for cropping would probably be the easiest to use. Worst case, if you can't get random cropping working, feel free to comment it out and we'll bring in the rest of your functional code before the repository gets archived. Someone in the future might be able to extend it in a fork to complete the rest.

BradLarson:
Also, I was able to fix the test failures that we've been seeing, so those should be out of the way whenever you feel comfortable with bringing this in.

acoadmarmon:
Ok got it! I think I will comment it out for now and in the future me or someone else can finish this off so we can get it through before the repo is archived.

acoadmarmon:
Ok I've pushed the latest changes that remove the use of Python TensorFlow from the training loop! Ready for review.

BradLarson (approved):
Looks good to me to pull in. Maybe at some point in the future, someone will carry this on and build out the remaining components, but this is a good base to work from.

Thanks for taking the time to build this and then refine it per our requests.

acoadmarmon:
Thank you @BradLarson @dan-zheng! Appreciate the time you all took to review this and help clean it up 😄 .

PR #730, opened 2021-01-14
Documenting models and components

BradLarson:
There are a lot of examples, models, and components contained within this repository. To highlight everything that resides here, this reworks the front-page README to act as an index for the examples and modules of interest within swift-models.


dan-zheng (approved):
Skimmed. LGTM, thank you!

